PR #1, opened 2023-01-19
Include EDPM Ceph client files to the list of roles

fmount:
This patch includes the `edpm_ceph_client_files` role to the list of the `edpm` roles.
It will be used to copy on the `external compute nodes` both `ceph.conf` and the
related `keyring`.

Signed-off-by: Francesco Pantano <fpantano@redhat.com>

### DIFF
--- a/edpm_ansible/roles/edpm_ceph_client_files/defaults/main.yml
+++ b/edpm_ansible/roles/edpm_ceph_client_files/defaults/main.yml
@@ -1,0 +1,24 @@
+---
+# Copyright 2022 Red Hat, Inc.
+# All Rights Reserved.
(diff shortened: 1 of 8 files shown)


rebtoor (approved):

fultonj:
/lgtm
/approve

I was going to ask why this is necessary because it already merged in tripleo-ansible [1] so I thought it would be inherited but now I see that edpm-ansible only has a subset of roles from tripleo-ansible. In that case, yes, let's please merge this.

[1] https://review.opendev.org/c/openstack/tripleo-ansible/+/865197

fultonj (approved):

PR #2, opened 2023-01-19
add requirements.txt

rebtoor:

Added missing requirements.txt file


fmount (approved):
+1

requirements.txt
```diff
@@ -0,0 +1 @@
+pbr>=1.6
```
kajinamit, line 1: wondering if we should attempt to add ansible back. In the past we had to remove it to avoid issue with upper constraint globally maintained but we don't have that in the independent org now.

https://review.opendev.org/c/openstack/tripleo-ansible/+/668465

kajinamit (approved):

requirements.txt
kajinamit: This can be discussed later. We definitely need the requrements.txt first to make installation pass.

PR #8, opened 2023-01-24
edpm_kernel: Remove HEAT_TEMPLATE from variable name

kajinamit:
Because we no longer use heat, the current variable name no longer makes sense.


kajinamit:
Note we later have to find out the appropriate update path as we need to remove the previous setting but that is not really new as we already renamed these variables as part of replacement of TRIPLEO.

rebtoor (approved):

rebtoor:
@kajinamit sign your commit with `git commit -s` please. :)

kajinamit:
> @kajinamit sign your commit with `git commit -s` please. :)

Done. I'm not sure whether this is the documented requirement but if so then we probably need a check to enforce it.

PR #10, opened 2023-01-24
Remove unused edpm_container_config_scripts module

kajinamit:
This module (originally tripleo_container_config_scripts module) is used to run puppet in a container, but we should NOT use puppet in this new architecture.


kajinamit:
This module is implemented but is not really used. We should remove it before somebody tries to use it.
```
[tkajinam@mylaptop edpm-ansible]$ grep -r edpm_container_config_scripts . --exclude-dir .git
./edpm_ansible/ansible_plugins/modules/edpm_container_config_scripts.py:module: edpm_container_config_scripts
./edpm_ansible/ansible_plugins/modules/edpm_container_config_scripts.py:  edpm_container_config_scripts:
[tkajinam@mylaptop edpm-ansible]$ 
```

rebtoor (approved):

PR #11, opened 2023-01-24
Remove unused filters

kajinamit:
Some filters are not used in any playbooks in this repository. We should avoid maintaining unused items. Especially some of these are quite specific to TripleO and do not make sense in EDPM.


jpodivin (approved):
Some of these are quite generic, but if truly necessary, and if they are not available from other sources, we could always reimplement them. 

All things considered I don't think we should carry them forward unless there is a strong reason to.

rebtoor (approved):

PR #12, opened 2023-01-24
Adding minimal tox setup to keep style parity with tripleo-ansible

jpodivin:
I've removed bulk of the tox configuration that wasn't directly involved in the style checking, while porting over files we need for linting ansible and yaml files.

I didn't include the documentation and release notes environments because it may be better to start from scratch there.

Openstack constraints file isn't compatible with required version of tenacity, so it's not being considered for now. 

Signed-off-by: Jiri Podivin <jpodivin@redhat.com>

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -14,16 +14,5 @@ mock_roles:
     - ceph-facts
 mock_modules:
     - baremetal_nodes_validate
-    - ceph_crush_rule
-    - ceph_dashboard_user
-    - ceph_key
-    - ceph_fs
-    - ceph_pool
-    - ceph_mkspec
-    - ceph_spec_bootstrap
     - config_template
     - container_startup_config
-    - lvm2_physical_devices_facts
-    - metalsmith_instances
-    - os_baremetal_clean_node
-    - os_baremetal_provide_node
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -9,9 +9,7 @@ verbosity: 1
 # Mocking modules is not recommended as it prevents testing of invalid
 # arguments or lack of their presence at runtime. It is preffered to
 # make use of requirements.yml to declare them.
-mock_roles:
-    - ceph-defaults
-    - ceph-facts
+# mock_roles:
 mock_modules:
     - baremetal_nodes_validate
     - config_template
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -11,6 +11,5 @@ verbosity: 1
 # make use of requirements.yml to declare them.
 # mock_roles:
 mock_modules:
-    - baremetal_nodes_validate
     - config_template
     - container_startup_config
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,0 +1,29 @@
+exclude_paths:
+    - releasenotes/
+    - ../
+parseable: true
+quiet: false
+rulesdir:
+    - .ansible-lint_rules/
+verbosity: 1
+# Mocking modules is not recommended as it prevents testing of invalid
+# arguments or lack of their presence at runtime. It is preffered to
+# make use of requirements.yml to declare them.
+mock_roles:
+    - ceph-defaults
+    - ceph-facts
+mock_modules:
+    - baremetal_nodes_validate
+    - ceph_crush_rule
+    - ceph_dashboard_user
+    - ceph_key
+    - ceph_fs
+    - ceph_pool
+    - ceph_mkspec
+    - ceph_spec_bootstrap
+    - config_template
+    - container_startup_config
+    - lvm2_physical_devices_facts
+    - metalsmith_instances
+    - os_baremetal_clean_node
+    - os_baremetal_provide_node
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,57 @@
+[tox]
+minversion = 4.0.0
+envlist = linters
+ignore_base_python_conflict = True
+
+[testenv]
+basepython = python3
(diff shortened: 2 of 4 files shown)


rebtoor:
@jpodivin hey! Thanks for the PR. I added tenacity in the requirements.txt file because the module was used but never installed. I didn't know which version was required so I added the last stable but if you need a particular version feel free to propose it so we can test it.

jpodivin:
The openstack requirements, which is what we usually use to sync dependencies across projects , allows for [6.3.0](https://github.com/openstack/requirements/blob/master/upper-constraints.txt#L493) as upper bound. 

We could go for same but I'm not certain if two major versions downgrade won't have an impact.

jpodivin:
@rebtoor I've updated the PR with new version of tenacity. I've also added .yamllint config and constraints.

jpodivin:
One more thing. I forgot about pre commit config.

jpodivin:
> One more thing. I forgot about pre commit config.

Turns out it's fine as it is.

rebtoor (changes requested):

.ansible-lint
```diff
@@ -0,0 +1,29 @@
...
+    - ceph-facts
+mock_modules:
+    - baremetal_nodes_validate
+    - ceph_crush_rule
+    - ceph_dashboard_user
+    - ceph_key
+    - ceph_fs
+    - ceph_pool
+    - ceph_mkspec
+    - ceph_spec_bootstrap
```
rebtoor, line 23: i think that ceph modules won't be used anymore
cc @fmount

.ansible-lint
```diff
@@ -0,0 +1,29 @@
...
+quiet: false
+rulesdir:
+    - .ansible-lint_rules/
+verbosity: 1
+# Mocking modules is not recommended as it prevents testing of invalid
+# arguments or lack of their presence at runtime. It is preffered to
+# make use of requirements.yml to declare them.
+mock_roles:
+    - ceph-defaults
+    - ceph-facts
```
rebtoor, line 14: i think that ceph modules won't be used anymore
cc @fmount

fmount: +1

fmount (changes requested):
Ceph mock_{roles, modules} can be deleted.

rebtoor:
> @rebtoor I've updated the PR with new version of tenacity. I've also added .yamllint config and constraints.

It seems that everything is working correctly with this version of tenacity.

.ansible-lint
```diff
@@ -0,0 +1,29 @@
...
+    - ceph_crush_rule
+    - ceph_dashboard_user
+    - ceph_key
+    - ceph_fs
+    - ceph_pool
+    - ceph_mkspec
+    - ceph_spec_bootstrap
+    - config_template
+    - container_startup_config
+    - lvm2_physical_devices_facts
```
kajinamit, line 26: In addition to ceph modules, there are some more modules (lvm2_physical_devices_facts, metalsmith_instances, os_baremetal_* ) which are not present in edpm repo.

tox.ini
```diff
@@ -0,0 +1,57 @@
...
+   # paramiko CryptographyDeprecationWarning: https://github.com/ansible/ansible/issues/52598
+   PYTHONWARNINGS=ignore:DEPRECATION::pip._internal.cli.base_command,ignore::UserWarning
+   PIP_DISABLE_PIP_VERSION_CHECK=1
+   TRIPLEO_ANSIBLE_COMPUTE_NODE_MOLECULE_CACHE={homedir}/.cache/edpm-ansible/containers
+   TRIPLEO_ANSIBLE_COMPUTE_NODE_MOLECULE_VOLUMES=['{homedir}/.cache/edpm-ansible/containers:/var/lib/containers:rw','/sys/fs/cgroup:/sys/fs/cgroup:rw']
+sitepackages = True
+deps =
+   -r {toxinidir}/requirements.txt
+   -r {toxinidir}/test-requirements.txt
+   -c {env:TOX_CONSTRAINTS_FILE:https://releases.openstack.org/constraints/upper/master}
```
kajinamit, line 35: we may want to use zed u-c for the time being ?

kajinamit (changes requested):

tox.ini
jpodivin: ack.

.ansible-lint
jpodivin: If there are no plans to import them I'm going to remove the lines.

.ansible-lint
jpodivin: ditto

.ansible-lint
```diff
@@ -0,0 +1,16 @@
...
+quiet: false
+rulesdir:
+    - .ansible-lint_rules/
+verbosity: 1
+# Mocking modules is not recommended as it prevents testing of invalid
+# arguments or lack of their presence at runtime. It is preffered to
+# make use of requirements.yml to declare them.
+# mock_roles:
+mock_modules:
+    - baremetal_nodes_validate
```
kajinamit, line 14: you can also remove this because I don't see any usage of the baremetal_nodes_validate module in the playbooks in this repo atm.

kajinamit (changes requested):

.ansible-lint
rebtoor: +1

jpodivin: Item removed.

rebtoor (approved):

PR #14, opened 2023-01-25
Get rid of reference to /var/lib/config/puppet-generated

kajinamit:
The path was used in TripleO but it no longer makes sense because we'll not use puppet to generate config files.


rebtoor:
@kajinamit you should rebase this patch to main. It's not mergiable as is.

rebtoor:
Tested with `quay.io/ralfieri/openstack-ansibleee-runner:fix-conf-path`

rebtoor (approved):

PR #15, opened 2023-01-25
Raising the bar on code compliance

jpodivin:
Most of the previously ignored style violations have been resolved. This includes unused imports and variables, as well as less intrusive style deficiencies.

Since we are starting essentially from scratch here I believe we should avoid perpetuating some of the issues that plagued our codebase before. Appropriate baseline is essential for this, as it prevents reintroduction of previously resolved problems. 

Any further code ported from tripleo-ansible will possibly require some changes to comply. But since it isn't going to be done in one go it should be manageable, while also offering an opportunity to look over some other cases inviting possible refactor. 

Signed-off-by: Jiri Podivin <jpodivin@redhat.com>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,10 +19,12 @@ repos:
     hooks:
       - id: flake8
         additional_dependencies: [flake8-typing-imports==1.6.0]
-        entry: flake8 --ignore=E24,E121,E122,E123,E124,E126,E226,E265,E305,E402,F401,F405,E501,E704,F403,F841,W503
-        # TODO(cloudnull): These codes were added to pass the lint check.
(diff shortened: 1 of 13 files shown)


jpodivin:
helpers.py were changed in the previous commit, I'll rebase and push again. At least this proves that the CI works.

rebtoor:
100% agree on this @jpodivin !

jpodivin:
It appears that the pr-check needs to be expanded to other states of PR. It didn't trigger when I force pushed a new patch.

jpodivin:
This should make sure that action is triggered on PR when the hash changes. 
#16

rebtoor (approved):

PR #19, opened 2023-01-27
Linting fixes for roles

rebtoor:
I tried to fix linting errors in every imported role, using FQCN for tasks, fixing indentation, comments, task names and so on.

This first patch has been successfully tested with this image: quay.io/ralfieri/openstack-ansibleee-runner:linting-fixes


edpm_ansible/roles/test_deps/vars/centos-8.yml
```diff
@@ -21,6 +21,6 @@ test_deps_yaml_packages:
 test_deps_selinux_packages:
   - python3-libselinux
 test_deps_edpm_packages:
-  - python3-edpmclient
+  - python3-tripleoclient
```
kajinamit, line 24: We no longer need this client file.

edpm_ansible/roles/test_deps/vars/centos-8.yml
```diff
@@ -21,6 +21,6 @@ test_deps_yaml_packages:
...
   - python3-openstacksdk
-test_deps_edpm_repos: current-edpm
+test_deps_tripleo_repos: current-tripleo
```
kajinamit, line 26: this var is not used so can be removed, I think

kajinamit (changes requested):
I generally agree with the idea but if we fix lint errors then we have to enforce the rules in the job. Please consider updating the pre-commit file to make sure we no longer introduce the regression.

Also, as a side note, I really hope we describe the intention of each commit in commit message. Just 'fixed this role' makes it quite hard to understand why the change was required

edpm_ansible/roles/test_deps/vars/centos-8.yml
rebtoor: ack, removed for every var file.

edpm_ansible/roles/test_deps/vars/centos-8.yml
rebtoor: same as above.

rebtoor:
> I generally agree with the idea but if we fix lint errors then we have to enforce the rules in the job. Please consider updating the pre-commit file to make sure we no longer introduce the regression.

Indeed, I'll open another PR for that, since we need to reach an agreement on which linting test should be disable. (e.g. commands instead of modules, widely used in these roles)

> Also, as a side note, I really hope we describe the intention of each commit in commit message. Just 'fixed this role' makes it quite hard to understand why the change was required

My fault, it started with a mere linting fixing job but i changed some things in few repos.

raukadah:
Based on above image: https://review.rdoproject.org/r/c/testproject/+/46569/95/roles/external_compute_play/templates/edpm-play.yaml.j2#64 and The job is green https://review.rdoproject.org/zuul/build/7d3cae93fda246a9a0fda0b5983a28c6 and here is the logs from the pod
 https://logserver.rdoproject.org/69/46569/95/check/ansibleee-operator-github-rdo-integration-centos-8-crc-singlenode-centos-9-external-compute/7d3cae9/logs/controller/pod/deploy-external-dataplane-compute-wwcj2-logs.txt

PR #20, opened 2023-01-30
Stop creating /var/lib/heat-config

kajinamit:
This directory was used in TripleO but should no longer be used in EDPM which does not use heat.


rebtoor:
@kajinamit please rebase this patch to main.

rebtoor:
Tested with quay.io/ralfieri/openstack-ansibleee-runner:heat-config

rebtoor (approved):

PR #21, opened 2023-01-30
Remove remaining reference to puppet package

kajinamit:
This removes the leftover of the puppet package, which was missed in e5f38797e18ff5639692027e62549870921773f4 .


kajinamit:
This one does not affect actual playbook or modules but only molecule tests which are not yet imported.

rebtoor (approved):

PR #26, opened 2023-02-01
remove edpm_container_image_prepare module

rebtoor:

This python module isn't needed by any roles in this repo and it has "tripleo-common" as its dependency.

If needed will be added in the future.


kajinamit (approved):
We have to decide how we manage containers for external nodes and we might restore this later.
However until we decide the plan we should not keep this incomplete implementation

PR #28, opened 2023-02-02
Adding Makefile which will content - Build openstack-ansibleee-runner…

bshewale:
… image from openstack-ansibleee-operator/blob/main/Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,184 @@
+##@ Build openstack-ansibleee-runner image
+.PHONY: docker-build-ee
+docker-build-ee:
+	if [ "${EDPM_LOCAL_REPO}" != "" ] ; then \
+		cd ansibleee; podman build -t ${EEIMG} . --volume ${EDPM_LOCAL_REPO}:/var/tmp/edpm-ansible:z --build-arg EDPM_LOCAL_REPO=${EDPM_LOCAL_REPO}; \
+	else \
+		cd ansibleee; podman build -t ${EEIMG} .; \
+	fi
+
+## Push openstack-ansible-runner image
+.PHONY: docker-push-ee
+docker-push-ee:
+	cd ansibleee; podman push --tls-verify=${VERIFY_TLS} ${EEIMG}
+
+##@ Deployment
+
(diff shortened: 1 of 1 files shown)


rebtoor (changes requested):
This Makefile, as is, is almost useless here (this isn't an operator repo, this repo doesn't follow the operators lifecycle, there's no go code and so on...). We only need the runner related targets `docker-build-ee` and `docker-push-ee`.  Could you kindly fix? Thanks!

Makefile
```diff
@@ -0,0 +1,184 @@
...
+.PHONY: gowork
+gowork: ## Generate go.work file to support our multi module repository
+	test -f go.work || go work init
+	go work use .
+	go work use ./api
+
+.PHONY: operator-lint
+operator-lint: gowork ## Runs operator-lint
+	GOBIN=$(LOCALBIN) go install github.com/gibizer/operator-lint@latest
+	go vet -vettool=$(LOCALBIN)/operator-lint ./... ./api/...
```
rebtoor, line 184: you can remove all these lines.

Makefile
```diff
@@ -0,0 +1,184 @@
+##@ Build openstack-ansibleee-runner image
```
rebtoor, line 1: Before the first target, add some useful variables that we lost in the last edit:
```

# openstack-ansibleee-runner image
EEIMG ?= quay.io/openstack-k8s-operators/openstack-ansibleee-runner

# edpm-ansible local repo
EDPM_LOCAL_REPO ?= ""

VERIFY_TLS ?= true


```

rebtoor (changes requested):

jpodivin:
@rebtoor I'd prefer not to do the makefile at all tbh. At this point it would be just two, exceedingly short, shell snippets and we can handle that in other ways, like with shell.

rebtoor:
> @rebtoor I'd prefer not to do the makefile at all tbh. At this point it would be just two, exceedingly short, shell snippets and we can handle that in other ways, like with shell.

Yeah, probably you're right.

PR #30, opened 2023-02-06
Trigger the container restart when configuration change

rebtoor:
- added `_create_checksum` function in `container_config_hash` ansible module, to generate hashes for given folders of configuration files. This function was performed in a bash script called `container-puppet.sh`
- removed debug and step parameters from `container_config_hash` ansible module
- changed `edpm_container_manage_update_config_hash` boolean variable to true in edpm_container_manage role in order to enable hashes updates.

### DIFF
--- a/edpm_ansible/ansible_plugins/modules/container_config_hash.py
+++ b/edpm_ansible/ansible_plugins/modules/container_config_hash.py
@@ -169,8 +169,7 @@ class ContainerConfigHashManager:
             if os.path.isfile(file) and file_relpath not in exclusions:
                 total_hash.update(self._read_file(file))
 
-        h = total_hash.hexdigest()
-        return h
+        return total_hash.hexdigest()
 
     def _set_config_hash(self, config_volume, exclusions=[]):
         """Create a config hash for a config_volume.
@@ -178,8 +177,7 @@ class ContainerConfigHashManager:
         :param exclusions: list
         :returns: string
         """
-        hash = self._create_checksum(config_volume, exclusions)
-        return hash
+        return self._create_checksum(config_volume, exclusions)
 
     def _get_config_base(self, prefix, volume):
         """Returns a config base path for a specific volume.
--- a/edpm_ansible/ansible_plugins/modules/container_config_hash.py
+++ b/edpm_ansible/ansible_plugins/modules/container_config_hash.py
@@ -241,7 +241,7 @@ class ContainerConfigHashManager:
                     exclude_from_hash = startup_config_json['environment'].get(
                         'EDPM_EXCLUDE_FROM_HASH', '').split(':')
                 new_hashes = [
-                    self._set_config_hash(vol_path,exclude_from_hash) for vol_path in config_volumes
+                    self._set_config_hash(vol_path, exclude_from_hash) for vol_path in config_volumes
                 ]
                 new_hash = '-'.join(new_hashes)
                 if new_hash != old_config_hash:
--- a/edpm_ansible/ansible_plugins/modules/container_config_hash.py
+++ b/edpm_ansible/ansible_plugins/modules/container_config_hash.py
@@ -22,6 +22,9 @@ import fnmatch
 import json
 import os
 import yaml
+import tarfile
+import tempfile
+import hashlib
 
 
 ANSIBLE_METADATA = {
@@ -35,6 +38,7 @@ DOCUMENTATION = """
 module: container_config_hash
 author:
   - "Emilien Macchi (@EmilienM)"
+  - "Roberto Alfieri (@rebtoor)"
 version_added: '2.9'
 short_description: Generate config hashes for container startup configs
 notes: []
@@ -53,16 +57,6 @@ options:
       - Config volume prefix
     type: str
     default: '/var/lib/config-data'
-  debug:
-    description:
-      - Enable debug
-    type: bool
-    default: False
-  step:
-    description:
-      - Step number
-    default: 6
-    type: int
 """
 
 EXAMPLES = """
@@ -145,6 +139,42 @@ class ContainerConfigHashManager:
         os.chmod(path, 0o600)
         self.results['changed'] = True
 
+    def _tar_cleanup(self, tarinfo):
+        tarinfo.mtime = 0
+        return tarinfo
+
+    def _create_checksum(self, config_volume):
+        """Create an hash from a tar-ed version of the given
+        folder
+
+        :param config_volume: string
+        :returns: string
+        """
+
+        total_hash = hashlib.new('md5')
+        tmp = tempfile.TemporaryFile()
+        tar = tarfile.open(fileobj=tmp, mode="w:")
+        tar.add(config_volume, recursive=True, filter=self._tar_cleanup)
+        tar.close
+        tmp.flush()
+        tmp.seek(0)
+        tar = tarfile.open(fileobj=tmp, mode='r:')
+        chunk_size = 100 * 1024
+
+        for member in tar:
+            if not member.isfile():
+                continue
+            f = tar.extractfile(member)
+            data = f.read(chunk_size)
+            while data:
+                total_hash.update(data)
+                data = f.read(chunk_size)
+        tar.close()
+
+        tmp.close()
+        h = total_hash.hexdigest()
+        return h
+
     def _get_config_hash(self, config_volume):
         """Returns a config hash from a config_volume.
 
@@ -154,6 +184,19 @@ class ContainerConfigHashManager:
         hashfile = "%s.md5sum" % config_volume
         if os.path.exists(hashfile):
             return self._slurp(hashfile).strip('\n')
+
+    def _set_config_hash(self, config_volume):
+        """Create a config hash for a config_volume.
+        :param config_volume: string
+        :returns: string
+        """
+        hash = self._create_checksum(config_volume)
+        hashfile = "%s.md5sum" % config_volume
+        with open(hashfile, 'w') as f:
+            f.write(hash)
+        os.chmod(hashfile, 0o600)
+
+        return hash
 
     def _get_config_base(self, prefix, volume):
         """Returns a config base path for a specific volume.
@@ -209,29 +252,28 @@ class ContainerConfigHashManager:
                 continue
             startup_config_json = json.loads(self._slurp(config))
             config_volumes = self._match_config_volumes(startup_config_json)
-            config_hashes = [
-                self._get_config_hash(vol_path) for vol_path in config_volumes
-            ]
-            config_hashes = filter(None, config_hashes)
-            if 'environment' in startup_config_json:
-                old_config_hash = startup_config_json['environment'].get(
-                    'EDPM_CONFIG_HASH', '')
-            if config_hashes is not None and config_hashes:
-                config_hash = '-'.join(config_hashes)
-                if config_hash == old_config_hash:
-                    # config doesn't need an update
-                    continue
-                self.module.debug(
-                    'Config change detected for {}, new hash: {}'.format(
-                        cname,
-                        config_hash
-                    )
-                )
-                if 'environment' not in startup_config_json:
-                    startup_config_json['environment'] = {}
-                startup_config_json['environment']['EDPM_CONFIG_HASH'] = (
-                    config_hash)
-                self._update_container_config(config, startup_config_json)
+            if config_volumes:
+                config_hashes = [
+                    self._get_config_hash(vol_path) for vol_path in config_volumes
+                ]
+                if 'environment' in startup_config_json:
+                    old_config_hash = startup_config_json['environment'].get(
+                        'EDPM_CONFIG_HASH', '')
+                if config_hashes is not None and config_hashes:
+                    new_hashes = [
+                        self._set_config_hash(vol_path) for vol_path in config_volumes
+                    ]
+                    new_hash = '-'.join(new_hashes)
+                    if new_hash != old_config_hash:
+                        if 'environment' not in startup_config_json:
+                            startup_config_json['environment'] = {}
+                        startup_config_json['environment']['EDPM_CONFIG_HASH'] = (
+                            new_hash)
+                        self._update_container_config(
+                            config, startup_config_json)
+                    else:
+                        # config doesn't need an update
+                        continue
 
 
 def main():
--- a/edpm_ansible/ansible_plugins/modules/container_config_hash.py
+++ b/edpm_ansible/ansible_plugins/modules/container_config_hash.py
@@ -22,8 +22,7 @@ import fnmatch
 import json
 import os
 import yaml
-import tarfile
-import tempfile
+import glob
 import hashlib
 
 
@@ -65,7 +64,7 @@ EXAMPLES = """
 """
 
 CONTAINER_STARTUP_CONFIG = '/var/lib/edpm-config/container-startup-config'
-
+BUF_SIZE = 65536
 
 class ContainerConfigHashManager:
     """Notes about this module.
@@ -139,39 +138,35 @@ class ContainerConfigHashManager:
         os.chmod(path, 0o600)
         self.results['changed'] = True
 
-    def _tar_cleanup(self, tarinfo):
-        tarinfo.mtime = 0
-        return tarinfo
+    def _read_file(self, file):
+        """Read a given file and return its content
+        :param file: string
+        :returns: bytes
+        """
 
-    def _create_checksum(self, config_volume):
+        r = b''
+        with open(file, 'rb') as f:
+            while True:
+                data = f.read(BUF_SIZE)
+                r += data
+                if not data:
+                    break
+        return r
+
+    def _create_checksum(self, config_volume, exclusions=[]):
         """Create an hash from a tar-ed version of the given
         folder
 
         :param config_volume: string
+        :param exclusions: list
         :returns: string
         """
 
         total_hash = hashlib.new('md5')
-        tmp = tempfile.TemporaryFile()
-        tar = tarfile.open(fileobj=tmp, mode="w:")
-        tar.add(config_volume, recursive=True, filter=self._tar_cleanup)
-        tar.close
-        tmp.flush()
-        tmp.seek(0)
-        tar = tarfile.open(fileobj=tmp, mode='r:')
-        chunk_size = 100 * 1024
+        for file in glob.glob(config_volume + '/**/*', recursive=True):
+            if os.path.isfile(file) and file not in exclusions:
+                total_hash.update(self._read_file(file))
 
-        for member in tar:
-            if not member.isfile():
-                continue
-            f = tar.extractfile(member)
-            data = f.read(chunk_size)
-            while data:
-                total_hash.update(data)
-                data = f.read(chunk_size)
-        tar.close()
-
-        tmp.close()
         h = total_hash.hexdigest()
         return h
 
@@ -259,21 +254,20 @@ class ContainerConfigHashManager:
                 if 'environment' in startup_config_json:
                     old_config_hash = startup_config_json['environment'].get(
                         'EDPM_CONFIG_HASH', '')
-                if config_hashes is not None and config_hashes:
-                    new_hashes = [
-                        self._set_config_hash(vol_path) for vol_path in config_volumes
-                    ]
-                    new_hash = '-'.join(new_hashes)
-                    if new_hash != old_config_hash:
-                        if 'environment' not in startup_config_json:
-                            startup_config_json['environment'] = {}
-                        startup_config_json['environment']['EDPM_CONFIG_HASH'] = (
-                            new_hash)
-                        self._update_container_config(
-                            config, startup_config_json)
-                    else:
-                        # config doesn't need an update
-                        continue
+                new_hashes = [
+                    self._set_config_hash(vol_path) for vol_path in config_volumes
+                ]
+                new_hash = '-'.join(new_hashes)
+                if new_hash != old_config_hash:
+                    if 'environment' not in startup_config_json:
+                        startup_config_json['environment'] = {}
+                    startup_config_json['environment']['EDPM_CONFIG_HASH'] = (
+                        new_hash)
+                    self._update_container_config(
+                        config, startup_config_json)
+                else:
+                    # config doesn't need an update
+                    continue
 
 
 def main():
--- a/edpm_ansible/ansible_plugins/modules/container_config_hash.py
+++ b/edpm_ansible/ansible_plugins/modules/container_config_hash.py
@@ -172,12 +172,13 @@ class ContainerConfigHashManager:
         h = total_hash.hexdigest()
         return h
 
-    def _set_config_hash(self, config_volume, exclude_from_hash):
+    def _set_config_hash(self, config_volume, exclusions=[]):
         """Create a config hash for a config_volume.
         :param config_volume: string
+        :param exclusions: list
         :returns: string
         """
-        hash = self._create_checksum(config_volume, exclude_from_hash)
+        hash = self._create_checksum(config_volume, exclusions)
         return hash
 
     def _get_config_base(self, prefix, volume):
--- a/edpm_ansible/ansible_plugins/modules/container_config_hash.py
+++ b/edpm_ansible/ansible_plugins/modules/container_config_hash.py
@@ -229,7 +229,7 @@ class ContainerConfigHashManager:
                 old_config_hash = startup_config_json['environment'].get(
                     'EDPM_CONFIG_HASH', '')
                 exclude_from_hash = startup_config_json.get(
-                    'edpm_exclude_from_hash', [])
+                    'edpm_exclude_files_from_hash', [])
                 new_hashes = [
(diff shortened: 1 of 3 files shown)


rebtoor:
I created this PR as a draft because i'm unsure about a pair of things:

1. I replicated the generation of the md5sum following the [script|https://opendev.org/openstack/tripleo-heat-templates/src/branch/master/common/container-puppet.sh] command. In that script a tar container with a folder and a list of exclusions was created, resetting its mtime files, then the tar file was 'extracted' to stdout and passed to a sed command (needed to strip some puppet comments AFAIK) and finally hashed by md5sum. The first two tar commands basically create a concatenated text file from a list of files that is evaluated via sed and then hashed. Do we still need this? Why we can't just concatenate the files and generate the md5sum of all of them without using tar? We don't even need to reset their mtime (md5sum of the same file created in different time is the same).
2. The list of exclusion indicated in the original shell script, is still valid for edpm-ansible? And, if some file shouldn't be parsed, the "list" should be maintained in the  module itself? Shouldn't be the container role (e.g. edpm_iscsid) that calls the module the owner of that list? 

Let me know what do you think about it, i think there's room for optimizations. @slagle @kajinamit @bshephar @rabi

rebtoor:
BTW, i created a runner (`quay.io/ralfieri/openstack-ansibleee-runner:hash`) with this branch' code, and i ran the playbook two times: when the edpm-compute-0 node was deployed and after I changed the configuration in the playbook (the hostname of rabbitmq endpoints).

The variable `EDPM_CONFIG_HASH` for `nova_compute` container has changed after the first and the second executions. The hash for nova_compute is a list of hashes, separated by `-` character because it mounts volumes from 2 different impacted folders: `/var/lib/config-data/ansible-generated/iscsid` (unchanged) and `/var/lib/config-data/ansible-generated/nova_libvirt` (changed)

```
[root@edpm-compute-0 ~]# podman inspect -f json nova_compute | jq .[].Config.Env 
  "LANG=en_US.UTF-8",
  "EDPM_CONFIG_HASH=d41d8cd98f00b204e9800998ecf8427e-78cab918a5af3acf96d285935f6490d9",
  "KOLLA_CONFIG_STRATEGY=COPY_ALWAYS",
  "LIBGUESTFS_BACKEND=direct",
  "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
  "TERM=xterm",
  "container=oci",
  "HOME=/var/lib/nova",
  "HOSTNAME=edpm-compute-0"
]


[
  "LIBGUESTFS_BACKEND=direct",
  "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
  "TERM=xterm",
  "container=oci",
  "LANG=en_US.UTF-8",
  "EDPM_CONFIG_HASH=d41d8cd98f00b204e9800998ecf8427e-668689aeb2d7d9c4e708b75281993888",
  "KOLLA_CONFIG_STRATEGY=COPY_ALWAYS",
  "HOME=/var/lib/nova",
  "HOSTNAME=edpm-compute-0"
]
```

slagle (commented):
This looks good. I would not worry about the excluded files or sed logic right now. The excluded file we might need to worry about it would be /etc/libvirt/passwd.db. Maybe we could add a way to list files to exclude from the config hashing in the container defs json. The module could then exclude files based on the container instead of hardcoded like they are in the old shell script.

For the sed, I think that would only apply if ansible adds comments to files. I know that the blockinfile module does that, but we are not using that module for any container configuration, so I think we're good there.

rebtoor:
Hey @slagle , thanks for the review!

> This looks good. I would not worry about the excluded files or sed logic right now. The excluded file we might need to worry about it would be /etc/libvirt/passwd.db. Maybe we could add a way to list files to exclude from the config hashing in the container defs json. The module could then exclude files based on the container instead of hardcoded like they are in the old shell script.

Makes sense, I was thinking something like that. Maintaining the hash generating function like this, we only need to change the tarfile filter function (_tarfile_cleanup) to avoid files with certain names.

> For the sed, I think that would only apply if ansible adds comments to files. I know that the blockinfile module does that, but we are not using that module for any container configuration, so I think we're good there.

Ansible generated comments are "static" (IIRC the template module put `#Managed by Ansible` while puppet put something with timestamp) so I'm not worried since it will be always the same.

Any thought on dropping tar file generation entirely?

edpm_ansible/ansible_plugins/modules/container_config_hash.py
```diff
@@ -145,6 +139,42 @@ def _update_container_config(self, path, config):
...
+    def _create_checksum(self, config_volume):
+        """Create an hash from a tar-ed version of the given
+        folder
+
+        :param config_volume: string
+        :returns: string
+        """
+
+        total_hash = hashlib.new('md5')
+        tmp = tempfile.TemporaryFile()
```
kajinamit, line 155: can we use context manager so that we ensure close this file?

edpm_ansible/ansible_plugins/modules/container_config_hash.py
```diff
@@ -145,6 +139,42 @@ def _update_container_config(self, path, config):
...
+        tar = tarfile.open(fileobj=tmp, mode="w:")
```
kajinamit, line 156: ditto. we should use context manager

edpm_ansible/ansible_plugins/modules/container_config_hash.py
```diff
@@ -145,6 +139,42 @@ def _update_container_config(self, path, config):
...
+        tar.add(config_volume, recursive=True, filter=self._tar_cleanup)
+        tar.close
+        tmp.flush()
+        tmp.seek(0)
+        tar = tarfile.open(fileobj=tmp, mode='r:')
+        chunk_size = 100 * 1024
```
kajinamit, line 162: I'd define this as global constant instead of variable within the method.

edpm_ansible/ansible_plugins/modules/container_config_hash.py
```diff
@@ -209,29 +252,28 @@ def _update_hashes(self):
...
-                    config_hash)
-                self._update_container_config(config, startup_config_json)
+            if config_volumes:
+                config_hashes = [
+                    self._get_config_hash(vol_path) for vol_path in config_volumes
+                ]
+                if 'environment' in startup_config_json:
+                    old_config_hash = startup_config_json['environment'].get(
+                        'EDPM_CONFIG_HASH', '')
+                if config_hashes is not None and config_hashes:
```
kajinamit, line 262: config_hashes can't be none and is always a list.

edpm_ansible/ansible_plugins/modules/container_config_hash.py
```diff
@@ -145,6 +139,42 @@ def _update_container_config(self, path, config):
         os.chmod(path, 0o600)
         self.results['changed'] = True
 
+    def _tar_cleanup(self, tarinfo):
+        tarinfo.mtime = 0
+        return tarinfo
+
+    def _create_checksum(self, config_volume):
```
kajinamit, line 146: We have to pull the logic to add/exclude some files, I guess.

https://github.com/openstack/tripleo-heat-templates/blob/master/common/container-puppet.sh#L164
https://github.com/openstack/tripleo-heat-templates/blob/master/common/container-puppet.sh#L170

Probably we want these override-able so that each role can define own requirements.

edpm_ansible/ansible_plugins/modules/container_config_hash.py
kajinamit, line 155: I think we no longer need tar file as it was used to just join file contents, which can be implemented differently here.

kajinamit (changes requested):
Thanks for starting this work. Please find my comments inline.

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: i changed completely the approach, see new commits :)

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: removed completely config_hashes since it's useless with the new approach.

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: 100% agree, see new commits.

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: For what I saw, there's no roles, at least for those that we have imported here, who write their config outside `/var/lib/config-data/ansible-generated` Even `logrotate-crond` and `iscsid` have their folder there. What should we add then?

About exclusions, I implemented the logic inside the `_create_checksum` function, defining an `exclusion` list, now defaulted to `[]`, but it's not referenced in its call yet. 

This module is called by the edpm_container_standalone role without any parameter because it was designed to run on *all* container startup config definition found in `CONTAINER_STARTUP_CONFIG` (which is a constant that points to `/var/lib/edpm-config/container-startup-config`) Handling an exclusion list for each service/role means that:

1. we change the logic behind this entire module, calling it for each service with their own parameters.
2. we add a variable in the container definition json file that enumerate a list of files that need to be excluded from hashing.

Are we sure that we still need to exclude files? Even if those files contains passwords or they binaries (like the ceph keyring linked), should they have the same hash if aren't changed?

I'm waiting for your feedback. (cc @slagle for another opinion)

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: no needed anymore

edpm_ansible/ansible_plugins/modules/container_config_hash.py
slagle: I think option 2 is what I was referring to earlier. I think we could manage the exclusion list in the container json (startup_config_json), and then pass the list into _set_config_hash (and then _create_checksum). So, it would still work per container, without any change to how the module is called.

We likely still need file exclusion for /etc/libvirt/passwd.db at least.

rebtoor: @slagle done. I'm searching in the `container_config_json` for an environment var called `EDPM_EXCLUDE_FROM_HASH`. The variable itself is a list of files separated by colon, just like the ${PATH} variable style. e.g. `EDPM_EXCLUDE_FROM_HASH=/etc/nova/nova.conf:/etc/libvirt/passwd.db`

rebtoor:
@kajinamit @slagle can you take another look at this so we can merge it? :) Thanks!

slagle (approved):

edpm_ansible/ansible_plugins/modules/container_config_hash.py
```diff
@@ -145,15 +139,47 @@ def _update_container_config(self, path, config):
...
+        """Create a config hash for a config_volume.
         :param config_volume: string
+        :param exclusions: list
         :returns: string
         """
-        hashfile = "%s.md5sum" % config_volume
-        if os.path.exists(hashfile):
-            return self._slurp(hashfile).strip('\n')
+        hash = self._create_checksum(config_volume, exclusions)
+        return hash
```
bshephar, line 182: The variable assignment is unnecessary here, we can just return `self._create_checksum(config_volume, exclusions)`

edpm_ansible/ansible_plugins/modules/container_config_hash.py
```diff
@@ -145,15 +139,47 @@ def _update_container_config(self, path, config):
...
+        """
+
+        total_hash = hashlib.new('md5')
+        for file in glob.glob(config_volume + '/**/*', recursive=True):
+            file_relpath = '/' + os.path.relpath(file, config_volume)
+            if os.path.isfile(file) and file_relpath not in exclusions:
+                total_hash.update(self._read_file(file))
+
+        h = total_hash.hexdigest()
+        return h
```
bshephar, line 173: Variable isn't necessary here, can just return `total_hash.hexdigest()`.

bshephar (approved):
Few comments that can be addressed in a follow up PR. Happy to merge this one as is.

edpm_ansible/ansible_plugins/modules/container_config_hash.py
```diff
@@ -209,29 +235,26 @@ def _update_hashes(self):
...
-            ]
-            config_hashes = filter(None, config_hashes)
-            if 'environment' in startup_config_json:
-                old_config_hash = startup_config_json['environment'].get(
-                    'EDPM_CONFIG_HASH', '')
-            if config_hashes is not None and config_hashes:
-                config_hash = '-'.join(config_hashes)
-                if config_hash == old_config_hash:
+            if config_volumes:
+                if 'environment' in startup_config_json:
```
kajinamit, line 239: Current logic does not initialize old_config_hash and exclude_from_hash in case environment key does not exist, which causes reference without assignment.

edpm_ansible/ansible_plugins/modules/container_config_hash.py
```diff
@@ -209,29 +235,26 @@ def _update_hashes(self):
...
+                    old_config_hash = startup_config_json['environment'].get(
+                        'EDPM_CONFIG_HASH', '')
+                    exclude_from_hash = startup_config_json['environment'].get(
+                        'EDPM_EXCLUDE_FROM_HASH', '').split(':')
```
kajinamit, line 243: I think we better put this out of environment (I mean, add the key to the same level as environment) because this is not really used by container itself.

kajinamit (changes requested):

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: ack

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: ack

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: uh, thanks!

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: It was my first thought but I kept variables in the container environment as they was. Let me do some tests.

raukadah:
https://logserver.rdoproject.org/30/30/6f35f21fa9fccf2eefb02b449650edae43e197ae/github-check/edpm-ansible-github-rdo-integration-centos-8-crc-singlenode-centos-9-external-compute/f321a44/logs/controller/pod/deploy-external-dataplane-compute-f65pm-logs.txt
```
TASK [edpm_container_manage : Update container configs with new config hashes] ***
[0;31mAn exception occurred during task execution. To see the full traceback, use -vvv. The error was: KeyError: 'edpm_config_hash'[0m
[0;31mfatal: [38.102.83.101]: FAILED! => {"changed": false, "module_stderr": "Shared connection to 38.102.83.101 closed.\r\n", "module_stdout": "Traceback (most recent call last):\r\n  File \"/root/.ansible/tmp/ansible-tmp-1676629918.98837-1355-260497113266536/AnsiballZ_container_config_hash.py\", line 107, in <module>\r\n    _ansiballz_main()\r\n  File \"/root/.ansible/tmp/ansible-tmp-1676629918.98837-1355-260497113266536/AnsiballZ_container_config_hash.py\", line 99, in _ansiballz_main\r\n    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)\r\n  File \"/root/.ansible/tmp/ansible-tmp-1676629918.98837-1355-260497113266536/AnsiballZ_container_config_hash.py\", line 47, in invoke_module\r\n    runpy.run_module(mod_name='ansible.modules.container_config_hash', init_globals=dict(_module_fqn='ansible.modules.container_config_hash', _modlib_path=modlib_path),\r\n  File \"/usr/lib64/python3.9/runpy.py\", line 225, in run_module\r\n    return _run_module_code(code, init_globals, run_name, mod_spec)\r\n  File \"/usr/lib64/python3.9/runpy.py\", line 97, in _run_module_code\r\n    _run_code(code, mod_globals, init_globals,\r\n  File \"/usr/lib64/python3.9/runpy.py\", line 87, in _run_code\r\n    exec(code, run_globals)\r\n  File \"/tmp/ansible_container_config_hash_payload_nl2gi8aq/ansible_container_config_hash_payload.zip/ansible/modules/container_config_hash.py\", line 264, in <module>\r\n  File \"/tmp/ansible_container_config_hash_payload_nl2gi8aq/ansible_container_config_hash_payload.zip/ansible/modules/container_config_hash.py\", line 260, in main\r\n  File \"/tmp/ansible_container_config_hash_payload_nl2gi8aq/ansible_container_config_hash_payload.zip/ansible/modules/container_config_hash.py\", line 90, in __init__\r\n  File \"/tmp/ansible_container_config_hash_payload_nl2gi8aq/ansible_container_config_hash_payload.zip/ansible/modules/container_config_hash.py\", line 237, in _update_hashes\r\nKeyError: 'edpm_config_hash'\r\n", "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error", "rc": 1}[0m

```

rebtoor:
> https://logserver.rdoproject.org/30/30/6f35f21fa9fccf2eefb02b449650edae43e197ae/github-check/edpm-ansible-github-rdo-integration-centos-8-crc-singlenode-centos-9-external-compute/f321a44/logs/controller/pod/deploy-external-dataplane-compute-f65pm-logs.txt
> 
> ```
> TASK [edpm_container_manage : Update container configs with new config hashes] ***
> �[0;31mAn exception occurred during task execution. To see the full traceback, use -vvv. The error was: KeyError: 'edpm_config_hash'�[0m
> �[0;31mfatal: [38.102.83.101]: FAILED! => {"changed": false, "module_stderr": "Shared connection to 38.102.83.101 closed.\r\n", "module_stdout": "Traceback (most recent call last):\r\n  File \"/root/.ansible/tmp/ansible-tmp-1676629918.98837-1355-260497113266536/AnsiballZ_container_config_hash.py\", line 107, in <module>\r\n    _ansiballz_main()\r\n  File \"/root/.ansible/tmp/ansible-tmp-1676629918.98837-1355-260497113266536/AnsiballZ_container_config_hash.py\", line 99, in _ansiballz_main\r\n    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)\r\n  File \"/root/.ansible/tmp/ansible-tmp-1676629918.98837-1355-260497113266536/AnsiballZ_container_config_hash.py\", line 47, in invoke_module\r\n    runpy.run_module(mod_name='ansible.modules.container_config_hash', init_globals=dict(_module_fqn='ansible.modules.container_config_hash', _modlib_path=modlib_path),\r\n  File \"/usr/lib64/python3.9/runpy.py\", line 225, in run_module\r\n    return _run_module_code(code, init_globals, run_name, mod_spec)\r\n  File \"/usr/lib64/python3.9/runpy.py\", line 97, in _run_module_code\r\n    _run_code(code, mod_globals, init_globals,\r\n  File \"/usr/lib64/python3.9/runpy.py\", line 87, in _run_code\r\n    exec(code, run_globals)\r\n  File \"/tmp/ansible_container_config_hash_payload_nl2gi8aq/ansible_container_config_hash_payload.zip/ansible/modules/container_config_hash.py\", line 264, in <module>\r\n  File \"/tmp/ansible_container_config_hash_payload_nl2gi8aq/ansible_container_config_hash_payload.zip/ansible/modules/container_config_hash.py\", line 260, in main\r\n  File \"/tmp/ansible_container_config_hash_payload_nl2gi8aq/ansible_container_config_hash_payload.zip/ansible/modules/container_config_hash.py\", line 90, in __init__\r\n  File \"/tmp/ansible_container_config_hash_payload_nl2gi8aq/ansible_container_config_hash_payload.zip/ansible/modules/container_config_hash.py\", line 237, in _update_hashes\r\nKeyError: 'edpm_config_hash'\r\n", "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error", "rc": 1}�[0m
> ```
> 
> Here is the container built by the job: quay.rdoproject.org/openstack-k8s-operators/openstack-ansibleee-runner:86bf25981bbfebfc9168601878af072e115781b9

yeah, i need to do some tests locally first. Let me put this PR on draft.

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: ok, now `edpm_config_hash` and `edpm_exclude_from_hash` are top level fields just like `environment`.

rebtoor:
@kajinamit @slagle @bshephar I should have solved your reviews' points, let me know if it might work now.

edpm_ansible/ansible_plugins/modules/container_config_hash.py
```diff
@@ -22,6 +22,8 @@
 import json
 import os
 import yaml
+import glob
+import hashlib
```
kajinamit, line 26: Please keep dictionary order. (eg. glob should be placed before json after fnmatch.) This is not allowed by lint which is not enabled yet.

edpm_ansible/ansible_plugins/modules/container_config_hash.py
```diff
@@ -145,15 +139,45 @@ def _update_container_config(self, path, config):
...
+
+        total_hash = hashlib.new('md5')
+        for file in glob.glob(config_volume + '/**/*', recursive=True):
+            file_relpath = '/' + os.path.relpath(file, config_volume)
+            if os.path.isfile(file) and file_relpath not in exclusions:
+                total_hash.update(self._read_file(file))
+
+        return total_hash.hexdigest()
 
+    def _set_config_hash(self, config_volume, exclusions=[]):
```
kajinamit, line 174: this function looks redundant. Also, this does not 'set' config hash and get would be more appropriate naming

edpm_ansible/ansible_plugins/modules/container_config_hash.py
```diff
@@ -145,15 +139,45 @@ def _update_container_config(self, path, config):
...
+        r = b''
+        with open(file, 'rb') as f:
+            while True:
+                data = f.read(BUF_SIZE)
+                r += data
+                if not data:
+                    break
+        return r
+
+    def _create_checksum(self, config_volume, exclusions=[]):
```
kajinamit, line 157: IMO create indicates this creates a resource. Probably get or calculate would be better.

edpm_ansible/ansible_plugins/modules/container_config_hash.py
```diff
@@ -209,29 +233,20 @@ def _update_hashes(self):
...
-                if config_hash == old_config_hash:
+            if config_volumes:
+                old_config_hash = startup_config_json.get('edpm_config_hash', '')
+                exclude_from_hash = startup_config_json.get('edpm_exclude_from_hash', [])
+                new_hashes = [
+                    self._set_config_hash(vol_path, exclude_from_hash) for vol_path in config_volumes
+                ]
+                new_hash = '-'.join(new_hashes)
+                if new_hash != old_config_hash:
+                    startup_config_json['edpm_config_hash'] = (new_hash)
```
kajinamit, line 244: the calculated hash should be part of environment so that the underlying edpm_container_manage and podman module cna detect the different hash.

https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/edpm_ansible/ansible_plugins/modules/edpm_container_manage.py#L299-L300

You have to update tripleo_container_manage if you get the result out of the environment hash.

kajinamit (changes requested):
This version looks better now, but left me leave some comments as I found a few things we still have to fix.

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: ack

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: ack

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: ack

edpm_ansible/ansible_plugins/modules/container_config_hash.py
rebtoor: we talked about it in chat, i misunderstood your original comment. I restore `EDPM_CONFIG_HASH` as `environment` variable.

kajinamit (approved):
LGTM. It might be useful to squash all commits into a single one before we merge this PR to keep git history clean.

rebtoor:
> LGTM. It might be useful to squash all commits into a single one before we merge this PR to keep git history clean.

Indeed, let me wait for CI test to succeed and I'll merge it squashing the commits.

Thanks a lot!

rebtoor:
recheck

PR #36, opened 2023-02-13
Consistent naming of bond_interface_ovs_options variable

bshephar:
There is currently no default defined for bond_interface_ovs_options. This change adds a default for the option to avoid Ansible issues if the variable isn't defined. This default was determined using the defaults were using in TripleO.

Additionally, this change is also bringing the name of the variable in-line with the standard variable naming convention used in the module by pre-pending edpm_ to the start of the name.


kajinamit (approved):
lgtm

raukadah:
recheck

PR #40, opened 2023-02-20
Added requirements for edpm_ovn_mac_address module

rebtoor:
- Added `validate_playbook_dir` and `tags_to_dict` functions in `network_data_v2.py` module
- Replaced `tripleo_` occurrency with `edpm_` (i'm a bit unsure about this...)
- Added `openstack.cloud` collection requirement


kajinamit:
We probably should rename the module because networkv2 is the terminology in TripleO.

Also updating the copyright does not look correct. Although the one in the code looks wired ( because the code was not written by anyone in the foundation side) we probably should keep the original one . I'd like to hear some other voices about this.

rebtoor:
> We probably should rename the module because networkv2 is the terminology in TripleO.

Ack. How about a more generic `edpm_network_helpers` ?

> Also updating the copyright does not look correct. Although the one in the code looks wired ( because the code was not written by anyone in the foundation side) we probably should keep the original one . I'd like to hear some other voices about this.

:+1: @slagle @bshephar @rabi WDYT?

rabi:
I would probably not change copyright, though it's not consistent across TripleO/OpenStack.

bshephar:
> We probably should rename the module because networkv2 is the terminology in TripleO.
> 
> Also updating the copyright does not look correct. Although the one in the code looks wired ( because the code was not written by anyone in the foundation side) we probably should keep the original one . I'd like to hear some other voices about this.

My preference would be that we use a licence file in the root directory of the repo and not worry about putting it at the top of each and every file? I'm not sure that is even necessary right? Can't we just do it this way?

https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository

edpm_ansible/ansible_plugins/module_utils/network_data_v2.py
```diff
@@ -0,0 +1,44 @@
...
+            playbook_dir_path))
+
+    if not os.path.isdir(playbook_dir_path):
+        raise Exception(
+            'ERROR: Playbook directory {} is not a directory'.format(
+                playbook_dir_path))
+
+
+def tags_to_dict(resource_tags):
+    tag_dict = dict()
```
bshephar, line 33: I'm not sure if this is still the case, but there was a cool blog post a few years ago about `dict()` vs `{}`. 
https://doughellmann.com/posts/the-performance-impact-of-using-dict-instead-of-in-cpython-2-7-2/

Be interesting to see if that is still the case in Python3.11, but I guess for this reason I have always preferenced using `variable = {}`, rather than `variable = dict()`.

rebtoor:
> > We probably should rename the module because networkv2 is the terminology in TripleO.
> > Also updating the copyright does not look correct. Although the one in the code looks wired ( because the code was not written by anyone in the foundation side) we probably should keep the original one . I'd like to hear some other voices about this.
> 
> My preference would be that we use a licence file in the root directory of the repo and not worry about putting it at the top of each and every file? I'm not sure that is even necessary right? Can't we just do it this way?
> 
> https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository
> 
> I feel like it's easier to keep the licence up-to-date if it's just in a single place. Rather than iterating the date every few years when we are touching a random file.

+1 for this approach!

edpm_ansible/ansible_plugins/module_utils/network_data_v2.py
rebtoor: TBH, i just imported code as-is from its original module, but I think you're right about this. Let me fix it.

bshephar (approved):
lgtm, we might be able to make it a little more efficient, but happy to follow up. I understand we're just bringing this over from tripleo-ansible.

We can probably follow up on the Licence thing too in a seperate PR.

requirements.yml
```diff
@@ -18,6 +18,7 @@ collections:
   - ansible.posix
   - community.general
   - containers.podman
+  - openstack.cloud
```
slagle, line 21: what's the reason for this? we don't want to use the openstack collection anywhere in edpm-ansible. we need to move away from that.

slagle (changes requested):

requirements.yml
rebtoor: It's [still used](https://github.com/openstack-k8s-operators/edpm-ansible/blob/67995fb067b069751d55f358c5541e9d96222bbb/edpm_ansible/ansible_plugins/modules/edpm_ovn_mac_addresses.py#L27-L29) in `edpm_ovn_mac_addresses` module, which is used by `edpm_ovn` role. We could think about importing only the required functions if they are needed.

rebtoor: The functions are:

- [openstack_full_ansible_spec](https://github.com/openstack/ansible-collections-openstack/blob/edd4e1b2e944370bc22dcfd75f8e41d886811fec/plugins/module_utils/openstack.py#L120-L146)
- [openstack_module_kwargs](https://github.com/openstack/ansible-collections-openstack/blob/edd4e1b2e944370bc22dcfd75f8e41d886811fec/plugins/module_utils/openstack.py#L149-L157)
- [openstack_cloud_from_modules](https://github.com/openstack/ansible-collections-openstack/blob/edd4e1b2e944370bc22dcfd75f8e41d886811fec/plugins/module_utils/openstack.py#L161-L207)

if we choose to import them we also add `openstacksdk` as dependency.

slagle: It's used in edpm_ansible/roles/edpm_ovn/tasks/externaldeploy.yml which came from external_deploy_tasks in deployment/ovn/ovn-controller-container-puppet.yaml from tripleo-heat-templates. That's not something we're going to use in edpm-ansible as it actually talks to neutron to allocate ports in the undercloud neutron. So, that doesn't fit the edpm architecture.

We may need something to still generate MAC's and populate the vars file that the module was doing (ovn_bridge_mac_address_vars.yaml). But, that won't need to talk to Neutron, so we can do without importing the openstack collection or openstacksdk in this module. @hjensas any thoughts about that?

hjensas: Totally agree we should not use neutron to allocate ports for OVN bridges. But AFIK we need something to generate MAC addresses and keep track that the MAC for each host:bridge is uniq. May need to create a helper service for this, some simple python rest service keep track of used MAC's and which host[bridge] mac belongs to? Or we might be able to create a kubernetes operator/controller? that would use the kubernetes etcd as the persistent store?

stuggi: in ospdo we created this controller [1] to create unique mac addresses. We might we want to add something similar to the infra-operator. 

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/controllers/openstackmacaddress_controller.go

PR #42, opened 2023-02-20
Adapt files to run tests

adrianfusco:
I've adapted the way to execute the tests for adding the job in Prow: https://github.com/openshift/release/pull/36549
Thanks @cjeanner I took part of your makefile for executing it https://github.com/openstack-k8s-operators/ci-framework/blob/master/Makefile


cjeanner:
I've pushed some more changes in a new PR, https://github.com/openstack-k8s-operators/ci-framework/pull/7 - it hopefully will provide the right env for pre-commit + molecule, with a custom container image, and some more Makefile target + the ability to NOT use the venv (especially in the CI - afaik that was my current issue)

rebtoor (approved):

PR #45, opened 2023-02-21
Move OS install tasks into edpm_bootstrap

fao89:
following: https://github.com/openstack-k8s-operators/dataplane-operator/pull/36#discussion_r1113357435


fao89:
recheck

fultonj (commented):
I'm +2 on the code. I just think a helpful commit message would be worthwhile. Could you update this with the following or write an alternative?

"Move OS install tasks from `deploy-edpm-os-install.yaml` playbook into `edpm_bootstrap` role."

rebtoor (approved):

PR #47, opened 2023-02-22
added edpm_tuned role

rebtoor:
* Imported `tuned` role as `edpm_tuned`
* Fixed variables names according to new name
* Added documentation
* Added install and configure steps to playbooks.

PR #49, opened 2023-02-22
adapted tuned role to be consistent to other edpm roles

rebtoor:

This is a follow-up patch for #47 , trying to fix all the comments from the @slagle review.

* updated copyright year to 2023
* modified tasks file name to `install.yml` and  `configure.yml`
* moved systemd service enable and restart to `run.yml` 
* enabled stat task for the `/etc/tuned/active_profile` file so the `slurp` task can be executed without errors only if that file exists
* modified the when clause for the `Enabled tuned profile` command. Now it will be executed if the `/etc/tuned/active_profile` doesn't exists or if that file exists but its content is different from the `edpm_tuned_profile` variable.


slagle (approved):

raukadah:
@rabi @bshephar Can we merge this PR ? last patch https://github.com/openstack-k8s-operators/edpm-ansible/pull/46 broke the zuul ci and it will fix it.

PR #53, opened 2023-02-27
Replace allovercloud with all

fao89:
dataplane-operator fails when trying to ssh into allovercloud

### DIFF
--- a/edpm_ansible/roles/edpm_ssh_known_hosts/tasks/main.yml
+++ b/edpm_ansible/roles/edpm_ssh_known_hosts/tasks/main.yml
@@ -50,7 +50,7 @@
       run_once: true
       ansible.builtin.set_fact:
         ssh_known_hosts_lines: |-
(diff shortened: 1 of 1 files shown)


rebtoor (approved):

edpm_ansible/roles/edpm_ssh_known_hosts/tasks/main.yml
```diff
@@ -50,7 +50,7 @@
       run_once: true
       ansible.builtin.set_fact:
         ssh_known_hosts_lines: |-
-          {% for host in groups['allovercloud'] | intersect(play_hosts) %}
+          {% for host in groups['all'] | intersect(play_hosts) %}
```
fao89, line 53: @slagle I'm wondering if it would be better if we have the dataplane-operator to set some fact and we could adapt things here according to that fact.

```
run_by: dataplane-operator

- task 
   when: run_by == "dataplane-operator"

- task 
   when: run_by != "dataplane-operator"
 ```

PR #57, opened 2023-03-01
Add OWNERS file to edpm-ansible

adrianfusco:
Propose OWNERS file for `edpm-ansible`.

PR in `release` project: https://github.com/openshift/release/pull/36549

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,10 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+approvers:
+  - edpm-approvers
+  - ci-approvers
+  - openstack-approvers
+
+reviewers:
+  - edpm-approvers
+  - ci-approvers
+  - openstack-approvers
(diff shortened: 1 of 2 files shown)


OWNERS
```diff
@@ -0,0 +1,29 @@
+approvers:
```
Sandeepyadav93, line 1: I think we should start using OWNERS_ALIASES along with OWNERS file, 

We already did this in dataplane-operator:
[1]  https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/OWNERS 
[2] https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/OWNERS_ALIASES

Could you please follow dataplane-operator pattern.

OWNERS
```diff
@@ -0,0 +1,29 @@
...
+  - raukadah
+  - sandeepyadav93
+  - viroel
+  - slagle
+  - rebtoor
+  - kajinamit
+  - bshephar
+  - raukadah
+  - rabi
+  - adrianfusco
```
rebtoor, line 29: Please, add an EOF newline :)

rebtoor (changes requested):

OWNERS
Sandeepyadav93: thanks for the fix.

lgtm now, I will let edpm-approver merge this patch - Incase they want to add someone else in the list.

rebtoor (approved):

PR #58, opened 2023-03-01
Add support for /runner/env/cmdline

slagle:
The contents of the $RUNNER_CMDLINE env var will be written to
/runner/env/cmdline, which will be passed to the ansible execution. This
will allow further configuration of the execution to set things such as
--tags.


raukadah:
recheck

fao89 (approved):

PR #62, opened 2023-03-06
Reuse container images variable in EDPM role

raukadah:
It modifies following roles to use existing container image variables:
- In edpm_nova_libvirt role
  - use edpm_nova_libvirt_container_image as edpm_nova_libvirt_container_config_image
- In edpm_ovn
  - use edpm_ovn_metadata_agent_image instead of hardcoding it.


rebtoor (approved):
ok for me

PR #65, opened 2023-03-10
Refactoring helpers module to be somewhat more pythonic

jpodivin:
cmp function was removed, as it essentially just wraps normal condition check.

Explicit inheritance from object was removed.

Retrieval of default value from dictionary was simplified.

I've also slightly expanded the docstrings so they would be more accurate,
 and reorganized conditional on line 108 to short circuit the evaluation.
Comparing two dictionaries is relatively expansive compared to plain boolean value.

### DIFF
--- a/edpm_ansible/ansible_plugins/filter/helpers.py
+++ b/edpm_ansible/ansible_plugins/filter/helpers.py
@@ -19,12 +19,7 @@ import json
 import re
 
 
-# cmp() doesn't exist on python3
-def cmp(a, b):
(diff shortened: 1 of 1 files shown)


jpodivin:
Eh, license appears to have an EOF issue?

rebtoor (approved):

jpodivin:
This needs rebase.

jpodivin:
done

rebtoor:
recheck

rebtoor:
LGTM

PR #67, opened 2023-03-13
Do not dump container definitions

kajinamit:
Some containers might have credentials in their environments, and dumping the whole definition can reveal the secrets. This ensures the definition is not dumped in logs. Note that users can still review the generated files to review the data.


kajinamit:
See https://bugs.launchpad.net/tripleo/+bug/1998181 to understand its context. We might have no contianers with such credential environments in computes but IMO we better avoid dumping redundant information to reduce possibility of secret leaking.

raukadah:
recheck

rebtoor (approved):

Issue #68, opened 2023-03-14
edpm-ansible container urls are out of sync with operators


SeanMooney:
The operators are currently using stable/zed images until at least stable/2023.1 is released.

the roles currently default to master images.

this means if we deploy using the roles as It stands today we will have zed on openshift
and master on the edpm hosts.

these should be the same by default.

PR #69, opened 2023-03-14
use stable/zed images by default

SeanMooney:
This change updates the contianer images in the role defaults
to use stable/zed to align with the current images used in the
operators.

Closes: #68


fultonj (commented):
+2

slagle (approved):

PR #70, opened 2023-03-15
Removed containers-common downgrade workaround

rebtoor:
I had already commented those tasks in the past, but they can be safely removed since [this](https://review.opendev.org/c/openstack/tripleo-quickstart/+/877386) has been merged.


cjeanner (approved):
That's what we call "cleaning old ***" :).

raukadah (approved):

PR #71, opened 2023-03-16
Convert edpm-ansible to a collection

rebtoor:
This is an attempt to convert this project to an ansible collection, so it could be installed by ansible-galaxy without using python setuptools. It could also packaged more easily (due to spec ansible-macros).


rebtoor:
recheck

rebtoor:
recheck

rebtoor:
recheck

kajinamit (approved):

PR #75, opened 2023-03-20
Drop incomplete implementation for virtlogd wrapper

kajinamit:
... because we no longer used the wrapper service. The current implementation does not work completely mainly because it still depends on puppet.


rebtoor:
@kajinamit can you fix the patch according to new layout? Thanks!

bogdando (approved):

rebtoor (approved):

PR #77, opened 2023-03-22
Correct how molecule is launched from within a container

cjeanner:
This should allow to run Molecule wihtout any issue. Note that the related commit[1] is also needed in order to make it fully usable.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/71


rebtoor (approved):

jpodivin:
I keep getting the error from before. But it may be just me.

PR #79, opened 2023-03-23
updated pre-commit tests

rebtoor:
This PR aim to upgrade the `pre-commit` hooks since they were imported from the parent project and they were rather old.

* pre-commit from version 3.4.0 to 4.4.0
* flake8 from version 3.9.0 to version 5.0.4 (version 6.0.0 is failing with an error, see comments)
* ansible-lint from 5.3.2 to 6.14.2 (old version requires ansible < 2.12 and we're using ansible 2.14)
* bashate from version 2.0.0 to version 2.1.1


jpodivin (approved):
Successfully tested on local machine.

PR #80, opened 2023-03-23
Remove plan and edpm_stack_name 

fao89:
```
TASK [osp.edpm.edpm_hosts_entries : Prepare new /etc/hosts] ********************
fatal: [edpm-compute-0]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'plan' is undefined. 'plan' is undefined\n\nThe error appears to be in '/usr/share/ansible/collections/ansible_collections/osp/edpm/roles/edpm_hosts_entries/tasks/main.yml': line 50, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Prepare new /etc/hosts\n  ^ here\n"}
```
https://github.com/openstack-k8s-operators/dataplane-operator/pull/116#discussion_r1145298940

### DIFF
--- a/roles/edpm_hosts_entries/tasks/main.yml
+++ b/roles/edpm_hosts_entries/tasks/main.yml
@@ -57,8 +57,8 @@
     insertbefore: BOF
     block: "{{ edpm_hosts_entries_block }}"
     marker: "# {mark}"
-    marker_begin: "START_HOST_ENTRIES_FOR_STACK: {{ edpm_stack_name | default(plan) }}"
-    marker_end: "END_HOST_ENTRIES_FOR_STACK: {{ edpm_stack_name | default(plan) }}"
+    marker_begin: START_HOST_ENTRIES_FOR_STACK
+    marker_end: END_HOST_ENTRIES_FOR_STACK
   register: edpm_hosts_entries_new_entries
   tags:
     - edpm_hosts_entries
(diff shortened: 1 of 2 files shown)


roles/edpm_hosts_entries/tasks/main.yml
```diff
@@ -57,8 +57,8 @@
     insertbefore: BOF
     block: "{{ edpm_hosts_entries_block }}"
     marker: "# {mark}"
-    marker_begin: "START_HOST_ENTRIES_FOR_STACK: {{ edpm_stack_name | default(plan) }}"
-    marker_end: "END_HOST_ENTRIES_FOR_STACK: {{ edpm_stack_name | default(plan) }}"
+    marker_begin: "START_HOST_ENTRIES_FOR_STACK: {{ edpm_stack_name | default(edpm_hosts_entries_plan) }}"
+    marker_end: "END_HOST_ENTRIES_FOR_STACK: {{ edpm_stack_name | default(edpm_hosts_entries_plan) }}"
```
slagle, line 61: I know you didn't cause this, but instead of defaulting the plan var, we need to just drop it. Plan isn't used with EDPM, and a default value of overcloud doesn't make sense and is a TripleO leftover.
I would then just change the marker message to be START_HOST_ENTRIES_FOR_EDPM, etc. It does not need to be dynamic based on any ansible var.

slagle (changes requested):

fao89:
recheck

rebtoor (approved):

PR #92, opened 2023-03-31
Remove logic for docker

kajinamit:
We have been using podman instead of docker since RHEL8, so we no longer need to maintain implementations to support docker now.


rebtoor (approved):

rebtoor:
recheck

rebtoor:
recheck

rebtoor:
recheck

PR #93, opened 2023-03-31
Remove roles that will not be used going forward.

SeanMooney:
the nova and libvirt roles have been replaced with playbooks in
the nova operator. while we may at somepoint revisit this
decision if we do we will not be using the currnet edpm_nova_*
roles in this repo. as such the roles are removed to avoid confusion
and ongoing maintance of unused roles

The nvdimm role is also removed as we removed support for
nvdimm in osp 17 following the dicontinuation of nvdimm support
and production of intel pmem. given We do not support this feature
in osp 17 and do not plan to support it going forward we do not
need this role.


SeanMooney:
i forgot to creat a new branch so I'm going to recreate this pr

PR #95, opened 2023-04-03
Added missing documentation for edpm_growvols role

rebtoor:
Added missing source documentation for `edpm_growvols` role


jpodivin (approved):
Patch is fine, although I have noticed that growvols vars have empty `redhat.yaml` file as sole content. It may be something to address later.

PR #96, opened 2023-04-04
Default edpm_ceph_client_files_source to /etc/ceph

fultonj:
Remove task to fail role if edpm_ceph_client_files_source is empty/undefined. Default edpm_ceph_client_files_source and edpm_ceph_client_files_config_home to /etc/ceph.

If edpm_ceph_client_files_source does not exist on Ansible host (e.g. stat returns "No such file or directory"), task will warn ("Unable to find '/etc/ceph' in expected paths") but not fail.

The dataplane-operator will invoke this role only when the extraMounts has an extraVolType of Ceph. In that case it is expected that /etc/ceph is mounted in the AEE pod as it is for other pods which use Ceph (e.g. Cinder or Glance). This role will then copy files in that directory to the EDPM node.

This change makes it easier for users since they don't need to override the variables edpm_ceph_client_files_config_home and edpm_ceph_client_files_source in the ansibleVars of the OpenStackDataPlane CR.

### DIFF
--- a/roles/edpm_ceph_client_files/tasks/main.yml
+++ b/roles/edpm_ceph_client_files/tasks/main.yml
@@ -14,17 +14,6 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Fail if edpm_ceph_client_files_source is missing
-  ansible.builtin.fail:
-    msg: >-
-      edpm_ceph_client_files_source must be set to a path that
-      already exists on the Ansible host which contains Ceph
-      configuration and Cephx key files.
-  when:
-    - edpm_ceph_client_files_source is not defined or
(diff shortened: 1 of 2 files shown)


rebtoor (approved):

roles/edpm_ceph_client_files/tasks/main.yml
```diff
@@ -14,17 +14,6 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Fail if edpm_ceph_client_files_source is missing
```
fmount, line 17: In general this looks good to me, we don't have to fail and propagate the failure state to the `dataplane-operator` (the AEE execution Status).

fmount (approved):

PR #99, opened 2023-04-12
Fix molecule jobs for edpm-ansible roles

rebtoor:
Launching `molecule test` on roles often returned errors caused by different issues.

With this PR I'm trying to address the errors and fix them.

Molecule version used for tests: 4.0.4
Ansible version used for tests: 2.14.4


jpodivin (approved):
Great work. Some optimizations may still be possible, but I think we can leave that to followup work.

PR #110, opened 2023-04-14
Get rid of tenacity

kajinamit:
The ansible-core package in CentOS 8 Stream and CentOS Stream 9 will use python runtimes which are different from the system default one. Because of this change, ansible controller is no longer able to consume the libraries installed in system default.

This removes the dependency on tenacity library installed in system default python to avoid the breakage caused by that change.


kajinamit:
We can implement more generic retry function if needed but I'll leave it for now as we use tenacity only for one module and having that generic utility is now overkilling.

rebtoor (approved):

PR #114, opened 2023-04-20
Remove ansible-requirements.txt

kajinamit:
This file is not really a requirements file but a upper-constraints file but the current content does not define any constraints so can be removed.

We probably need this later when we pin ansible to a specific version but the file can be re-added at that time.

### DIFF
--- a/ansible-requirements.txt
+++ b/ansible-requirements.txt
@@ -1,4 +1,0 @@
-ansible
-metalsmith>=1.4.0 # Apache-2.0
-importlib-metadata
-jsonschema  # MIT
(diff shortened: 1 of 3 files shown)


rebtoor:
+2 for me, but I'd ask for a review from CI folks since zuul playbooks are involved.

cjeanner:
/lgtm

I don't see why we should keep any empty file. Let's clean the tree!

rebtoor (approved):

kajinamit:
Fixed the missing signed-off tag in the commit message

rebtoor (approved):

rebtoor (approved):

PR #116, opened 2023-04-20
Molecule/common containerfile

jpodivin:
Moving from per-role to centralized containerfile has several advantages. 
1. It's easier to globally change some aspect of our molecule images.
2. Minimum chance of missing a role when performing some change.
3. Cuts down on repo bloat.
4. Allows for greater generalization.

There are also some potential issues, which I nevertheless consider to be of lesser import. The file can be too general, for example, and require considerable number of set variables to work properly. This could increase confusion on the part of developer. On the other hand, if we are careful we can easily block the bloat, by reviewing changes to the containerfile.

Furthermore, if we end with situation where such increase of parameters is unavoidable, we can simply add another centralized containerfile, which would cover sufficient subset of use cases generating them.


rebtoor:
I like this approach, less duplicated code is always desirable. 

IMHO the majority of configuration changes between roles scenarios can be handled by prepare playbooks, so if we maintain a sane container base we should be fine.

rebtoor:
recheck

rebtoor (approved):

jpodivin:
> I like this approach, less duplicated code is always desirable.
> 
> IMHO the majority of configuration changes between roles scenarios can be handled by prepare playbooks, so if we maintain a sane container base we should be fine.

Considerable adjustments to individual containers can also be made in the `molecule.yml` files of their respective scenarios. Things like ports exposure, env vars and packages installed included. But we obviously must be careful not to overdo it. Otherwise we end up with Jinja template hundreds of lines long.

PR #117, opened 2023-04-20
Expanding build ignore list of the collection

jpodivin:
Number of files previously included in the collection tar will now be excluded. Most importantly this includes all files in molecule, and test directories, as defined by blob expression, as well as docs directory.

In case of the docs I was slightly conflicted. But since we are already excluding the finished documentation from build and packaging, I guess we don't really need the source either. So away it goes.


rebtoor:
I 100% agree about molecule files but I'm still unsure about the docs.

A docs folder should be included in a collection [1](https://docs.ansible.com/ansible/latest/dev_guide/developing_collections_structure.html#docs-directory) but probably we should change its content.

jpodivin:
> I 100% agree about molecule files but I'm still unsure about the docs.
> 
> A docs folder should be included in a collection [1](https://docs.ansible.com/ansible/latest/dev_guide/developing_collections_structure.html#docs-directory) but probably we should change its content.

Why do they have two divergent guides to the same thing?[2](https://docs.ansible.com/ansible/devel/dev_guide/developing_collections_documenting.html) Why?!

jpodivin:
Rereading the ansible docs, it seems like are documentions isn't compliant with what they expect of community collection at all. Good news is that we are not a community collection yet, so our docs don't have to follow their standard. If at some point we aim for this to be a published, or even certified collection, we should revisit this topic and establish compliant docs. Until then I would prefer we didn't package them at all. 

wdyt @rebtoor ?

rebtoor (approved):

rebtoor:
I agree. Let's drop it from the collection now.

PR #124, opened 2023-04-24
molecule tests for edpm-podman role

jpodivin:
Proper tests will require delegated driver, since the role deals with interfaces and services on the level we can't provide in pure container.


jpodivin:

PR #125, opened 2023-04-25
Modifying molecule settings for edpm_nova_libvirt role

jpodivin:


jpodivin:
The role will be removed later, so we don't have to test it.

PR #128, opened 2023-04-26
Molecule tests for edpm_module_load role

jpodivin:
Tests work apart from the `remove_module` scenario, which is failing on the removal step.


jpodivin:

PR #132, opened 2023-04-27
Removing superfluous .gitkeep files

jpodivin:
Relevant subdirectories of both `edpm_nova_compute` and `edpm_nova_libvirt` already contain other valid files, tracked by git. As such there is no need for dummy `.gitkeep` files to maintain the subdirs.


fao89 (approved):

rebtoor:
recheck

rebtoor (approved):

PR #136, opened 2023-05-02
Adding CodeQL badge to README.md

jpodivin:

PR #137, opened 2023-05-02
Turning README.rst into README.md

jpodivin:
The README.rst file isn't included in docs we are building, so there is no need to keep the rst syntax.

Move to markdown will also allow us to include badges showing current CI status etc. as in #136


kajinamit:
/lgtm

README looks good at https://github.com/jpodivin/edpm-ansible/tree/rst-to-md

rebtoor (approved):

PR #153, opened 2023-05-17
WIP: Importing Tripleo-IPA content

jpodivin:
So far without test coverage or any verification of basic function.


jpodivin:
On hold until it's clear we want this.

PR #155, opened 2023-05-18
Mocking imported module to avoid ansible-autodoc import error

jpodivin:
Eventually we will have to move away from ansible-autodoc, as it is no longer maintained and issues like this will only pile on. However this should enable us to build docs for the time being.


rebtoor (approved):
tested locally, LGTM!

PR #156, opened 2023-05-18
Manage swap partition as a hard requirement

bogdando:
Add swap partition management into the bootstrap role

The swap file/partition management logic is carried over from the existing EDPM pre-adoption implementation, but moved out of the nova compute specific scope into generic tasks for EDPM.

Depends-On: https://review.rdoproject.org/r/c/rdo-jobs/+/48683

Closes: [OSPRH-133](https://issues.redhat.com//browse/OSPRH-133)

### DIFF
--- a/roles/edpm_bootstrap/defaults/main.yml
+++ b/roles/edpm_bootstrap/defaults/main.yml
@@ -50,6 +50,12 @@ edpm_bootstrap_network_service: NetworkManager
 # String for SELinux state. One of: disabled, enforcing, permissive
 edpm_bootstrap_selinux_mode: enforcing
 
+# Swap management
+edpm_bootstrap_swap_size_megabytes: 4096
+edpm_bootstrap_swap_path: /swap
+edpm_bootstrap_swap_partition_enabled: false
+edpm_bootstrap_swap_partition_label: swap1
+
 # Shell command that is executed before any packages are installed by the role.
 # Can be used to register systems using any arbitrary registration command(s).
 edpm_bootstrap_command: ""
--- a/roles/edpm_bootstrap/tasks/bootstrap.yml
+++ b/roles/edpm_bootstrap/tasks/bootstrap.yml
@@ -177,6 +177,9 @@
         name: NetworkManager
         state: reloaded
 
+- name: Configure swap
+  ansible.builtin.import_tasks: swap.yml
+
 - name: Set GID hugetlbfs user to match kolla pre-set
   become: true
   group:
--- a/roles/edpm_bootstrap/tasks/swap.yml
+++ b/roles/edpm_bootstrap/tasks/swap.yml
@@ -1,0 +1,59 @@
+---
+# Copyright 2023 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Configure swap file
+  become: true
+  when:
+    - not edpm_bootstrap_swap_partition_enabled|bool
+    - edpm_bootstrap_swap_size_megabytes|int > 0
+  shell: |
+    #!/bin/bash
+    set -eu
+    if [ ! -f {{ edpm_bootstrap_swap_path }} ]; then
+      dd if=/dev/zero of={{ edpm_bootstrap_swap_path }} count={{ edpm_bootstrap_swap_size_megabytes }} bs=1M
+      chmod 0600 {{ edpm_bootstrap_swap_path }}
+      mkswap {{ edpm_bootstrap_swap_path }}
+      swapon {{ edpm_bootstrap_swap_path }}
+      if ! grep -qE "{{ edpm_bootstrap_swap_path }}\b" /etc/fstab; then
+        echo "{{ edpm_bootstrap_swap_path }} swap swap defaults 0 0" >> /etc/fstab
+      fi
+    fi
+
+- name: Configure swap partition
+  become: true
+  when:
+    - edpm_bootstrap_swap_partition_enabled|bool
+  shell: |
+    #!/bin/bash
+    set -eu
+    changed=1
+    if [ -e "/dev/disk/by-label/{{ edpm_bootstrap_swap_partition_label }}" ]; then
+      swap_partition=$(realpath /dev/disk/by-label/{{ edpm_bootstrap_swap_partition_label }})
+      if ! grep -qE "${swap_partition}\b" /etc/fstab; then
+        echo "$swap_partition swap swap defaults 0 0" >> /etc/fstab
+        changed=0
+      fi
+    else
+      for item in $(lsblk -f --output FSTYPE,UUID | awk '/swap/ {print $2}'); do
+        if ! grep -qe "${item}" /etc/fstab; then
+          echo -e "UUID=${item} swap swap defaults 0 0" >> /etc/fstab
+          changed=0
+        fi
+      done
+    fi
+    if [ $changed -eq 0 ]; then
+      swapon -a
+    fi
--- a/roles/edpm_bootstrap/molecule/default/prepare.yml
+++ b/roles/edpm_bootstrap/molecule/default/prepare.yml
@@ -24,6 +24,9 @@
         - { option: 'ExecStartPre', state: 'absent' }
         - { option: 'ExecStart', value: '/bin/true' }
 
+    - name: Mock swapon command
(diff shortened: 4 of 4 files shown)


bogdando:
The molecule testing is blocked as we cannot use ``swapon`` from podman containers, we'd need a delegated driver for that, likely. Any hints to make this unblocked, @raukadah perchance?

bogdando:
the rdo-project job is failed, probably by unrelated reason, because I can see the swap task completed there

https://logserver.rdoproject.org/56/156/814b11000ae0869a6e1559bee243ec11f332ff28/github-check/edpm-ansible-github-rdo-integration-centos-8-crc-singlenode-centos-9-external-compute/79eb0cc/controller/controller/pod/deploy-external-dataplane-compute-n4zc8-logs.txt

bogdando:
check-rdo

bogdando:
check-rdo

jpodivin:
Just to clarify. The molecule in podman seems to work with this patch after all https://github.com/openstack-k8s-operators/edpm-ansible/actions/runs/5014407281/jobs/8988616358?pr=156 . Do you still want to replace it with delegated driver?

roles/edpm_bootstrap/defaults/main.yml
```diff
@@ -49,3 +49,9 @@ edpm_bootstrap_network_service: NetworkManager
 
 # String for SELinux state. One of: disabled, enforcing, permissive
 edpm_bootstrap_selinux_mode: enforcing
+
+# Swap management
+edpm_bootstrap_swap_size_megabytes: 4096
+edpm_bootstrap_swap_path: /swap
+edpm_bootstrap_swap_partition: false
+edpm_bootstrap_swap_partition_label: swap1
```
kajinamit, line 57: I think we should hard-fail if the partition with the given label is not found.

roles/edpm_bootstrap/defaults/main.yml
```diff
@@ -49,3 +49,9 @@ edpm_bootstrap_network_service: NetworkManager
 
 # String for SELinux state. One of: disabled, enforcing, permissive
 edpm_bootstrap_selinux_mode: enforcing
+
+# Swap management
+edpm_bootstrap_swap_size_megabytes: 4096
+edpm_bootstrap_swap_path: /swap
+edpm_bootstrap_swap_partition: false
```
kajinamit, line 56: The var name does not represent well that this is a boolean flag to enable swap. Probably edpm_bootstrap_enable_swap (or swap_enabled) is better.

roles/edpm_bootstrap/tasks/bootstrap.yml
```diff
@@ -160,3 +160,47 @@
...
+    - edpm_bootstrap_swap_size_megabytes|int > 0
+  shell: |
+    #!/bin/bash
+    set -eu
+    if [ ! -f {{ edpm_bootstrap_swap_path }} ]; then
+      dd if=/dev/zero of={{ edpm_bootstrap_swap_path }} count={{ edpm_bootstrap_swap_size_megabytes }} bs=1M
+      chmod 0600 {{ edpm_bootstrap_swap_path }}
+      mkswap {{ edpm_bootstrap_swap_path }}
+      swapon {{ edpm_bootstrap_swap_path }}
+      if ! grep -qE "{{ edpm_bootstrap_swap_path }}\b" /etc/fstab; then
```
kajinamit, line 177: Can't we use the native module to manipulate fstab entries ?

roles/edpm_bootstrap/tasks/bootstrap.yml
```diff
@@ -160,3 +160,47 @@
       ansible.builtin.service:
         name: NetworkManager
         state: reloaded
+
+- name: Configure swap file
```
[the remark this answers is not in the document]
bogdando: According to the original [tripleo design spec](https://opendev.org/openstack/tripleo-specs/commit/59e352f1a928c8469cb46756eac6113397d06de4) which established the standalone roles aka EDPM for NG, there are the following tasks (stages):

* bootstrap
* install
* pre-network
* network
* configure
* container-config
* service-bootstrap

While those are likely a subject of changes, the idea behind, I believe is to reduce the need for future migrations, like using mutating hooks to update all CRs for EDPM objects, when changing the tasks and roles (if we expose it as a public/partners integration interface)

So, no, we should keep it in the bootstrap.yaml tasks

roles/edpm_bootstrap/tasks/bootstrap.yml
bogdando: Those are valid points, just please note that this change ports the unmodified implementation that we used to have in tripleo, and still do have in the layered product (where this feature is a release blocker for the upcoming release relying now on EDPM). This implementation has been suppored. We need to unblock the upcoming release by porting it, then we can think on changing it

roles/edpm_bootstrap/tasks/bootstrap.yml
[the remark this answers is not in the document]
slagle: I don't have any strong opinion on splitting it out into a separate role. A separate task file would be fine though that could just be included here. It might make things a bit cleaner, and adds the ability to execute just the swap tasks if needed.

roles/edpm_bootstrap/defaults/main.yml
bogdando: There is a fallback path https://github.com/bogdando/edpm-ansible/blob/f6144dd7cc0d767d2469a8922d93042a92bbaeaf/roles/edpm_bootstrap/tasks/swap.yml#L50 in such a case.
Please rememver, this is a direct port of the tripleo supported feature. We can adjust it in followups

bogdando:
Ci failure looks unrelated https://github.com/openstack-k8s-operators/edpm-ansible/actions/runs/5199776741/jobs/9377637527#step:6:249

roles/edpm_bootstrap/tasks/swap.yml
```diff
@@ -0,0 +1,59 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Configure swap file
+  become: true
+  when:
+    - not edpm_bootstrap_swap_partition_enabled|bool
+    - edpm_bootstrap_swap_size_megabytes|int > 0
+  shell: |
```
bshephar, line 22: We should probably use `ansible.builtin.shell` for these. I think we were trying to get consistency around the FQCN's with this project.

roles/edpm_bootstrap/tasks/swap.yml
```diff
@@ -0,0 +1,59 @@
...
+      if ! grep -qE "{{ edpm_bootstrap_swap_path }}\b" /etc/fstab; then
+        echo "{{ edpm_bootstrap_swap_path }} swap swap defaults 0 0" >> /etc/fstab
+      fi
+    fi
+
+- name: Configure swap partition
+  become: true
+  when:
+    - edpm_bootstrap_swap_partition_enabled|bool
+  shell: |
```
bshephar, line 39: And here `ansible.builtin.shell`.

roles/edpm_bootstrap/tasks/swap.yml
```diff
@@ -0,0 +1,59 @@
...
+    #!/bin/bash
+    set -eu
+    if [ ! -f {{ edpm_bootstrap_swap_path }} ]; then
+      dd if=/dev/zero of={{ edpm_bootstrap_swap_path }} count={{ edpm_bootstrap_swap_size_megabytes }} bs=1M
+      chmod 0600 {{ edpm_bootstrap_swap_path }}
+      mkswap {{ edpm_bootstrap_swap_path }}
+      swapon {{ edpm_bootstrap_swap_path }}
+      if ! grep -qE "{{ edpm_bootstrap_swap_path }}\b" /etc/fstab; then
+        echo "{{ edpm_bootstrap_swap_path }} swap swap defaults 0 0" >> /etc/fstab
+      fi
```
bshephar, line 32: I see this has already been discussed, but it would be nice if we can follow up after this is merged to reimplement using: https://docs.ansible.com/ansible/latest/collections/ansible/posix/mount_module.html

bshephar (commented):
I can see this role has some inconsistencies with FQCN usage. But I think we should probably aim for using the FQCN's with these new PR's. We can clean up the rest of the role later.

I see you mentioned in the comments that you're just porting a pre-existing thing to edpm-ansible in order to unblock some work. So I'm happy to merge this one and address the comments I made in a follow up PR if you prefer.

bogdando:
@kajinamit PTAL

roles/edpm_bootstrap/tasks/swap.yml
bogdando: ack, let's keep this for follow-ups

roles/edpm_bootstrap/tasks/swap.yml
bogdando: ack, let's keep this for follow-ups

roles/edpm_bootstrap/tasks/swap.yml
bogdando: ack, let's keep this for follow-ups

kajinamit (approved):
I don't know why github doesn't remove my change request.

rebtoor:
recheck

PR #157, opened 2023-05-18
Fix applying sysctl kernel settings

bogdando:
Extend the pattern used for merging edpm_kernel_modules and edpm_kernel_extra_modules, for edpm_kernel_sysctl_settings and edpm_kernel_sysctl_extra_settings.

Provide a cleaner and consistend with ketnel modules user interface, for systcl overrides as well. Merge it with opinionated extra sysctl defaults that we ship with the ansible role.

Fix the missing combine() logic for the existing sysctl role defaults and extras in role vars. Define opnionated defaults in the role vars instead of splitting them into different places in the role.

Add the former nova compute specific setting for
vm.unprivileged_userfaultfd to extend it for all EDPM hosts.

Closes: OSPRH-155

### DIFF
--- a/roles/edpm_kernel/vars/main.yml
+++ b/roles/edpm_kernel/vars/main.yml
@@ -20,51 +20,87 @@ edpm_kernel_modules:
   nf_conntrack: {}
 
 edpm_kernel_sysctl_settings:
+  fs.inotify.max_user_instances:
+    value: 1024
+  fs.suid_dumpable:
+    value: 0
+  kernel.dmesg_restrict:
+    value: 1
+  kernel.pid_max:
+    value: 1048576
+  net.bridge.bridge-nf-call-arptables:
+    value: 1
+  net.bridge.bridge-nf-call-ip6tables:
+    value: 1
+  net.bridge.bridge-nf-call-iptables:
+    value: 1
+  net.core.netdev_max_backlog:
+    value: 10000
+  net.ipv4.conf.all.arp_accept:
+    value: 1
+  net.ipv4.conf.all.arp_notify:
+    value: 1
+  net.ipv4.conf.all.log_martians:
+    value: 1
+  net.ipv4.conf.all.rp_filter:
+    value: 1
+  net.ipv4.conf.all.secure_redirects:
+    value: 0
+  net.ipv4.conf.all.send_redirects:
+    value: 0
+  net.ipv4.conf.default.accept_redirects:
+    value: 0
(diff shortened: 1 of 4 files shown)


bogdando:
To fully close it, we'll need to add a PR to the dataplane operator to expose the ``edpm_kernel_sysctl_extra_settings``, or rather ``edpm_kernel_sysctl_settings``?

SeanMooney:
> To fully close it, we'll need to add a PR to the dataplane operator to expose the `edpm_kernel_sysctl_extra_settings`, or rather `edpm_kernel_sysctl_extra_settings`?

technically no

i would prefer there to be a sysctl:{"key":"value"} filed to be on the dataplaneRole/Node

but you can define edpm_kernel_sysctl_extra_settings or edpm_kernel_sysctl_extra_settings

using the AnsibleVars filed on the node/role so i think we can close the jira issue for now with that and defer the conversation of if we ant to keep AnsibleVars (in which case we have nothing else to do ) or if we want to have CRD fields for sysctl.

roles/edpm_kernel/vars/main.yml
```diff
@@ -20,51 +20,87 @@ edpm_kernel_modules:
...
+  net.ipv6.ip_nonlocal_bind:
     value: 1
-  net.core.netdev_max_backlog:
-    value: 10000
-  kernel.dmesg_restrict:
+  net.netfilter.nf_conntrack_max:
+    value: 500000
+  net.nf_conntrack_max:
+    value: 500000
+  vm.unprivileged_userfaultfd:
```
SeanMooney, line 105: as a follow up you can remove this for the nova role.

SeanMooney (commented):
in general this looks good to me.
our default values are all in the role vars keeping edpm_kernel_sysctl_extra_setting purely for end user overrides

that simplifies the ux of this role since  they just add additional parameter instead of needing to know what was in edpm_kernel_sysctl_extra_setting and merging them.

this will be a lot less error prone as a result

roles/edpm_kernel/vars/main.yml
bogdando: we can do in a single swath, by https://github.com/openstack-k8s-operators/edpm-ansible/pull/94, as we address all the comments there

slagle (approved):

bogdando:
FTR, related tripleo Wallaby sync https://review.opendev.org/c/openstack/tripleo-ansible/+/883659

PR #159, opened 2023-05-19
Remove edpm_nvdimm role as no longer supported

bogdando:
Intel discontinued supported for there intel optane datacenter memory aka vPMEM aka persistent memory in august of last year.

As a result we decided not to graduate it form tech preview in 16.1 to GA in 17.0 (thus the NextGen 18.0 as well) and instead remove support for it in our product instead.

Closes: [OSPRH-139](https://issues.redhat.com//browse/OSPRH-139)


scripts/test_roles.py
```diff
@@ -13,7 +13,6 @@
     "edpm_module_load",
     "edpm_chrony",
     "edpm_podman",
```
kajinamit, line 15: we can remove the tailing comma.

kajinamit (commented):
Looks almost good to me. Leaving a minor comment in case you can fix it quickly.

jpodivin (approved):
The PR is fine from the edpm pov. Thanks for handling this @bogdando .

Shame about Optane. It seemed like an interesting tech.

slagle (approved):

bogdando:
> The PR is fine from the edpm pov. Thanks for handling this @bogdando . One last thing before we can merge this. You need to sign your commits, otherwise we can't get it in our main.
> 
> Shame about Optane. It seemed like an interesting tech.

hm, but I do sign all my commits, don't I? not sure what is missing?

PR #160, opened 2023-05-19
Extract KSM management task

bogdando:
KSM is not Nova compute specific, manage it in the
generic EDPM kernel role's deidcated task instead.

Also drop the RHEL8 family specific code path as
there will only be RHEL>=9 in NextGen EDPM, and
rework the tripleo-specific approach for using handlers
to remove the KSM package after disabling the services.

Closes: [OSPRH-137](https://issues.redhat.com//browse/OSPRH-137)


roles/edpm_bootstrap/defaults/main.yml
```diff
@@ -49,3 +49,6 @@ edpm_bootstrap_network_service: NetworkManager
 
 # String for SELinux state. One of: disabled, enforcing, permissive
 edpm_bootstrap_selinux_mode: enforcing
+
+# KSM control
+edpm_bootstrap_enable_ksm: false
```
SeanMooney, line 54: hum ok i was expecting this to be in its own role.

i guess bootstrap si not terrible but lets see what others think

SeanMooney (commented):
overall i think this is ok but I'm not sure how large the scope of Bootstrap is intended to be. currently its pretty small so its probably ok.

i also agree with keeping ksm off by default.
it can be useful in some cases but for most of our customer i think off is the right default as they wont want the CPU overhead and non-determinisum cause by ksm.

bogdando:
> overall i think this is ok but I'm not sure how large the scope of Bootstrap is intended to be. currently its pretty small so its probably ok.
> 
> i also agree with keeping ksm off by default. it can be useful in some cases but for most of our customer i think off is the right default as they wont want the CPU overhead and non-determinisum cause by ksm.

yes, I believe such tiny tasks do not need dedicated roles, and fits the (host) bootstrap task ideally.

roles/edpm_bootstrap/defaults/main.yml
kajinamit: +1 to creating a seprate role (edpm_ksm ?). We can include it from bootstrap but adding more service logics directly to bootstrap playbook makes future maintenance hard.

kajinamit:
I should have checked the comments here before I posted the review.

> yes, I believe such tiny tasks do not need dedicated roles, and fits the (host) bootstrap task ideally. This also aligns with the "Tripleo future" spec, where we agreed to cover things like that by the bootstrap task

I'm not too sure this is really true. For example we have a separate role for kernel parameters. What you are saying would mean we agreed to merge even edpm_kernel role to bootstrap role but I don't think implementing more logics directly in boostrap is better for readability and maintainability.

roles/edpm_bootstrap/defaults/main.yml
slagle: Why not the edpm_kernel role in a ksm.yml tasks file?

slagle (changes requested):

roles/edpm_bootstrap/defaults/main.yml
jpodivin: I would also argue for keeping this functionality separate. However, we could implement this as it's own role, and invoke that, conditionally, from edpm_bootstrap. 

How about that?

bogdando: I would prefer @slagle's approach. No need for a dedicated role for such a tiny config bit?

SeanMooney:
> this seems need a full rewrite after the change in edpm roles structure (no roles, but playbooks?)

this has not changed we still have roles we just also have playbooks for the openstackdataplane services

so this does not need a full rewrite i just need to be extracted either into its own role or into one of the existing roles.

if you extract it into its own role then that role will need to be called form the bootstrap playbook.

bogdando:
> Why not the edpm_kernel role in a ksm.yml tasks file?
@slagle is it OK to use it for bootstrap role? Do you insist to move this logic into edpm_kernel?
I'm not sure, the latter looks very much specific to managing kernel modules and sysctl vars, not some arbitrary actions

slagle:
> > Why not the edpm_kernel role in a ksm.yml tasks file?
> 
> @slagle is it OK to use it for bootstrap role? Do you insist to move this logic into edpm_kernel? I'm not sure, the latter looks very much specific to managing kernel modules and sysctl vars, not some arbitrary actions

I do think it should be in a separate role from bootstrap. Otherwise bootstrap is just a catch all for a bunch of unrelated tasks.

bogdando:
> > > Why not the edpm_kernel role in a ksm.yml tasks file?
> > 
> > 
> > @slagle is it OK to use it for bootstrap role? Do you insist to move this logic into edpm_kernel? I'm not sure, the latter looks very much specific to managing kernel modules and sysctl vars, not some arbitrary actions
> 
> I do think it should be in a separate role from bootstrap. Otherwise bootstrap is just a catch all for a bunch of unrelated tasks.

ok, I'll update this then, thanks!

bogdando:
All done

roles/edpm_kernel/tasks/ksm.yml
```diff
@@ -0,0 +1,61 @@
...
+        - name: Delete PageKSM after disable ksm
+          ansible.builtin.command: echo 2 >/sys/kernel/mm/ksm/run
+          when:
+            - ksm_service_check.rc is defined
+            - ksm_service_check.rc == 0
+            - ksmdisabled is changed
+
+    - name: Enable KSM
+      when: edpm_kernel_enable_ksm|bool
+      ansible.builtin.service:
```
rebtoor, line 54: use `ansible.builtin.systemd`

roles/edpm_kernel/tasks/ksm.yml
```diff
@@ -0,0 +1,61 @@
...
+      when: not edpm_kernel_enable_ksm|bool
+      block:
+        - name: Check for ksm
+          ansible.builtin.shell: systemctl is-active ksm.service || systemctl is-enabled ksm.service
+          become: true
+          failed_when: false
+          register: ksm_service_check
+
+        - name: Disable KSM services
+          ansible.builtin.service:
```
rebtoor, line 29: use `ansible.builtin.systemd`

roles/edpm_kernel/tasks/ksm.yml
```diff
@@ -0,0 +1,61 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Enable/disable ksm
+  block:
+    - name: Disable KSM
+      when: not edpm_kernel_enable_ksm|bool
+      block:
+        - name: Check for ksm
+          ansible.builtin.shell: systemctl is-active ksm.service || systemctl is-enabled ksm.service
```
rebtoor, line 23: Since you're going to stop and disable the services in the next task, and there isn't a `when` condition in this one, i'd remove this task.

Shell is one of the slowest module in ansible so in the worst case you're going to execute 2 tasks for an operation that you are doing anyway.

roles/edpm_kernel/tasks/ksm.yml
```diff
@@ -0,0 +1,61 @@
...
+          become: true
+          failed_when: false
+          register: ksm_service_check
+
+        - name: Disable KSM services
+          ansible.builtin.service:
+            name: "{{ item }}"
+            state: stopped
+            enabled: false
+          with_items:
```
rebtoor, line 33: `with_items` use is discouraged, use `loop`

roles/edpm_kernel/tasks/ksm.yml
```diff
@@ -0,0 +1,61 @@
...
+            - ksm_service_check.rc == 0
+            - ksmdisabled is changed
+
+    - name: Enable KSM
+      when: edpm_kernel_enable_ksm|bool
+      ansible.builtin.service:
+        name: "{{ item }}"
+        state: started
+        enabled: true
+      with_items:
```
rebtoor, line 58: `with_items` use is discouraged, `use loop`

rebtoor (changes requested):
I left a few comments and, other than those, you should add `edpm_kernel_enable_ksm` variable in the `meta/argument_specs.yml` file (and obviously removing it from `edpm_nova_compute` role)

roles/edpm_kernel/tasks/ksm.yml
bogdando: we used to need this check in tripleo to trigger the package removal after disabling the service. Since we couldn't use handlers in tripleo by the design concerns, I guess. But here it seems I can remove that and notify a handler to remove the package whenever the service has been stopped

bogdando:
All done, PTAL

roles/edpm_kernel/meta/argument_specs.yml
```diff
@@ -52,3 +52,7 @@ argument_specs:
         type: dict
         default: {}
         description: Hugepages configuration dictionary.
+      edpm_kernel_enable_ksm:
+        default: false
+        description: ''
```
rebtoor, line 57: can you add a short description here, pls? :)

Even "Enable KSM" would be fine.

bogdando: good point, more docs never can hurt

rebtoor (approved):

rebtoor:
recheck

rebtoor:
@bogdando jobs is failing with this error:

```
2023-09-20T14:13:13.481258510+00:00 stdout F TASK [osp.edpm.edpm_kernel : Disable KSM services] *****************************
2023-09-20T14:13:13.482572161+00:00 stdout F Wednesday 20 September 2023  14:13:13 +0000 (0:00:00.054)       0:00:32.375 *** 
2023-09-20T14:13:14.231864503+00:00 stdout F [0;31mfailed: [edpm-compute-0] (item=ksm.service) => {"ansible_loop_var": "item", "changed": false, "item": "ksm.service", "msg": "Could not find the requested service ksm.service: host"}[0m
2023-09-20T14:13:14.876269491+00:00 stdout F [0;31mfailed: [edpm-compute-0] (item=ksmtuned.service) => {"ansible_loop_var": "item", "changed": false, "item": "ksmtuned.service", "msg": "Could not find the requested service ksmtuned.service: host"}[0m
```

Is the package who owns those files missing?

bogdando:
> @bogdando jobs is failing with this error:
> 
> ```
> 2023-09-20T14:13:13.481258510+00:00 stdout F TASK [osp.edpm.edpm_kernel : Disable KSM services] *****************************
> 2023-09-20T14:13:13.482572161+00:00 stdout F Wednesday 20 September 2023  14:13:13 +0000 (0:00:00.054)       0:00:32.375 *** 
> 2023-09-20T14:13:14.231864503+00:00 stdout F �[0;31mfailed: [edpm-compute-0] (item=ksm.service) => {"ansible_loop_var": "item", "changed": false, "item": "ksm.service", "msg": "Could not find the requested service ksm.service: host"}�[0m
> 2023-09-20T14:13:14.876269491+00:00 stdout F �[0;31mfailed: [edpm-compute-0] (item=ksmtuned.service) => {"ansible_loop_var": "item", "changed": false, "item": "ksmtuned.service", "msg": "Could not find the requested service ksmtuned.service: host"}�[0m
> ```
> 
> Is the package who owns those files missing?

it looks like ksmtuned is a leftover from rhel8 support https://opendev.org/openstack/tripleo-heat-templates/src/branch/master/deployment/nova/nova-compute-container-puppet.yaml#L1678
 I will probably remove it, if my assumption for no longer maintaining rhel8 for NG EDPM was correct. But it may be not, as we still have mixed-rhel adoption story. Putting this on hold

/hold

slagle (approved):

slagle:
> > @bogdando jobs is failing with this error:
> > ```
> > 2023-09-20T14:13:13.481258510+00:00 stdout F TASK [osp.edpm.edpm_kernel : Disable KSM services] *****************************
> > 2023-09-20T14:13:13.482572161+00:00 stdout F Wednesday 20 September 2023  14:13:13 +0000 (0:00:00.054)       0:00:32.375 *** 
> > 2023-09-20T14:13:14.231864503+00:00 stdout F �[0;31mfailed: [edpm-compute-0] (item=ksm.service) => {"ansible_loop_var": "item", "changed": false, "item": "ksm.service", "msg": "Could not find the requested service ksm.service: host"}�[0m
> > 2023-09-20T14:13:14.876269491+00:00 stdout F �[0;31mfailed: [edpm-compute-0] (item=ksmtuned.service) => {"ansible_loop_var": "item", "changed": false, "item": "ksmtuned.service", "msg": "Could not find the requested service ksmtuned.service: host"}�[0m
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is the package who owns those files missing?
> 
> it looks like ksmtuned is a leftover from rhel8 support https://opendev.org/openstack/tripleo-heat-templates/src/branch/master/deployment/nova/nova-compute-container-puppet.yaml#L1678 I will probably remove it, if my assumption for no longer maintaining rhel8 for NG EDPM was correct. But it may be not, as we still have mixed-rhel adoption story. Putting this on hold
> 
> /hold

There is no mixed-rhel support that includes rhel 8 for NG. You can drop any rhel8 specific tasks.

rebtoor (approved):

PR #162, opened 2023-05-23
Basic documentation for our filters

jpodivin:
Our filter plugins don't have ansible compliant docs yet. That complicates things during the doc build process using antsibull-docs. These files will stand in for module attributes usually leveraged to define documentation of ansible plugins. 

This practice is comparable to the one employed by ansible core plugins[0].

[0] https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/filter/to_json.yml


jpodivin:
recheck

rebtoor (approved):

PR #163, opened 2023-05-23
Argspecs for edpm-ansible roles

jpodivin:
Basic argument specs for our roles. In most cases they are left to be filled out later, although they are still possible to verify. In long run this should simplify maintenance, as we will, in effect, have a well defined interface for our roles.

Practical impact at this point is mostly with enabling antsibull-docs functionality.


Jaganathancse (commented):
/lgtm

rebtoor (approved):

PR #167, opened 2023-05-25
Add NB DB certs config for ovn-bgp-agent NB driver

luis5tb:
There is a new driver for ovn-bgp-agent that uses NB instead of SB OVN DBs, thus it needs extra information/configuration to connect to NB in case of TLS being setup


raukadah:
recheck

rebtoor (approved):

PR #169, opened 2023-05-25
Added the telemetry role and playbooks to deploy telemetry agents

jlarriba:


jpodivin (changes requested):
Two minor points:
1. Could you please set up a zuul job for the molecule test? Otherwise we can't be sure how it will behave, not to mention it would have to be implemented in a followup.
2. Adding an argspec[0] would be a great way to formalize the role. It doesn't have to be anything fancy, just something that would pass verification stage.

[0]https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#specification-format

jpodivin (approved):
Thanks for the changes. To answer your question about zuul. I was slightly suspicious of the container use in the role. Molecule doesn't always do well when it comes to nested containers. But since it seems to be working I don't see any need for zuul either.

The zuul jobs are only used to test a couple of roles at this time, that's why you probably didn't see them. As for the argspecs, those were a recent (last week) addition. Thanks for pointing out that they are missing in the skeleton. I'll take a look at that soon.

rebtoor:
recheck

rebtoor (approved):

PR #174, opened 2023-05-30
Replacing IMG_BASE_TAG with IMAGE_TAG_BASE in README.md

jpodivin:
The guidelines offered by README.md for building and pushing an image were referring to environment variable IMG_BASE_TAG, which is not considered by the Makefile or the related scripts.

Following the guide would lead to unauthorized access exception upon push, since the default value of the IMAGE_TAG_BASE would be used instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,12 +43,12 @@ image can be rebuilt and pushed to a container repository.
 
 To build the image:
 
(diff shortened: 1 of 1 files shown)


raukadah (approved):
/lgtm

slagle (approved):

PR #176, opened 2023-05-31
antsibull-docs introduction

jpodivin:
Switching our doc build to antsibull-docs based approach. The original ansible-autodoc was removed, instead the documentation for all of our roles, plugins and content, will be generated by antsibull-docs. The existing approach of integrating handwritten and automatically generated content will be maintained. Roles documentation can still include further information defined purely in the `docs` directory. 

The antsibull-docs script requires some modification in order to avoid issues antsibull-docs standards, which are more strict than those of ansible itself. This concerns chiefly the question of `description` keys in argspec and their data type. Their inclusion would be time prohibitive and would bring little in a way of value to the generated documentation.

Since this collection will not be published on the Ansible Galaxy storefront, we don't have to follow the standard completely. For the same reason a `sed` script will remove disclaimers about presence of our collection on Ansible Galaxy. 
Doc build script now also includes installation of the collection itself, as antsibull-docs uses local installation of the collection to obtain necessary information for docs build.

Files generated by antsibull-docs runtime will not be scanned for issues by `doc8` as they often contain issues with whitespace. Instead the subdirectory will be skipped and only static files will be subjected to linter.

The `conf.py` script used by the sphinx documentation tool was simplified, since there is no need to mock some of the modules imported. Intersphinx[0] settings were added to improve reading experience and information retrieval.

New dependencies were added in the `docs-requirements.txt`.

Filter plugins now have `short_description` key in their documentation to allow rendering by `antsibull-docs`.

Documentation for the edpm_container_manage role was adjusted to remove things covered by the `antsibull-docs` files.

[0]https://docs.readthedocs.io/en/stable/guides/intersphinx.html


jpodivin:
The docs will have to wait until either the argspecs are expanded, or proposed fix is merged into antsibull-docs[0] and released.

[0]https://github.com/ansible-community/antsibull-docs/pull/154

rebtoor (approved):
Tested locally, it works!

Great work @jpodivin ++

PR #177, opened 2023-06-02
Option to select the required DDP package

Jaganathancse:
This changes to provide the option to choose the required DDP package when multiple different packages are available.


rebtoor (changes requested):
I'd prefer to have a role with all these tasks integrated in an existing playbook or in a new one. It would give us more flexibility in using it.

Strictly speaking about the code, you should use FQCN for all used ansible modules.

jpodivin (changes requested):
I believe `edpm_module_load` would be useful in this case. Also it may be for the best to turn other tasks into separate role of it's own, with properly exposed parameters, I assume similar installation procedure could be useful for other Intel drivers?

playbooks/deploy_edpm_kernel_ddp_pkg.yaml
slagle, line 1: I agree with the other comments that these tasks need to be added to an existing role, or a new role as opposed to a new playbook. EDPM consumes the roles, not the playbooks.

slagle (changes requested):

playbooks/deploy_edpm_kernel_ddp_pkg.yaml
Jaganathancse: Ack

rebtoor (approved):

PR #178, opened 2023-06-06
Correct the enable-chassis-as-gw logic

luis5tb:
The enable-chassis-as-gw is to enable the nodes to host ovn router gateway ports, not to enable/disable DVR [1].

This patch renames the option so that it is clear that is for making compute nodes able to host gw ports

[1] https://github.com/openstack-k8s-operators/neutron-operator/pull/153

### DIFF
--- a/roles/edpm_ovn/defaults/main.yml
+++ b/roles/edpm_ovn/defaults/main.yml
@@ -31,7 +31,7 @@ edpm_ovn_chassis_mac_mapping_prefixes:
 edpm_ovn_chassis_mac_mapping_seed: "{{ ansible_machine_id }}"
 edpm_ovn_encap_type: geneve
 edpm_ovn_dbs: []
-edpm_enable_dvr: true
+edpm_enable_chassis_gw: false
 edpm_enable_hw_offload: false
 edpm_ovn_multi_rhel: false
 edpm_enable_internal_tls: false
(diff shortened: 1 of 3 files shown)


karelyatin (approved):
Thanks LGTM

roles/edpm_ovn/defaults/main.yml
```diff
@@ -31,7 +31,7 @@ edpm_ovn_chassis_mac_mapping_prefixes:
 edpm_ovn_chassis_mac_mapping_seed: "{{ ansible_machine_id }}"
 edpm_ovn_encap_type: geneve
 edpm_ovn_dbs: []
-edpm_enable_dvr: true
+edpm_enable_chassis_gw: false
```
jpodivin, line 34: Everything seems good to me. But I'm wondering about this flip. Is the fact that the chassis gw is disabled by default, rather than previous situation when it was enabled, properly documented?

luis5tb: I thought the intention here was to have DVR enabled by default and not chassis-gws. So I enabled by default DVR in https://github.com/openstack-k8s-operators/neutron-operator/pull/153, and make this only about the chassis gw -- which I think the intention was to only have it by default on the controllers. 
It was in fact highlighed by Yatin that unless that option is set on the controllers (enable-chassis-as-gw), all the nodes (computes and controllers) will be used anyway for this.

jpodivin: >I thought the intention here was to have DVR enabled by default and not chassis-gws. So I enabled by default DVR in https://github.com/openstack-k8s-operators/neutron-operator/pull/153, and make this only about the chassis gw 
In that case it's all fine. Maybe it would be better to add that info commit message next time.

PR #183, opened 2023-06-08
Refactored `edpm_config_network` role to support `linux-system-role.network` in place of `os-net-config`

rebtoor:
In the `edpm_config_network` role we have added the `edpm_network_config_role` variable that control the way how networking should be configured on edpm compute nodes.

The default value, `os-net-config`, will continue to configure the network with `os-net-config`.

Setting the value to `nmstate` the `network` role from linux-system-roles/rhel-system-roles will be used. This allow to configure interfaces using `nmstate` as its backend. See the [examples](https://nmstate.io/) on nmstate website to learn how to configure interfaces.

A short example is provided in molecule/nmstate scenario.

### DIFF
--- a/roles/edpm_bootstrap/defaults/main.yml
+++ b/roles/edpm_bootstrap/defaults/main.yml
@@ -23,7 +23,6 @@ edpm_bootstrap_packages_bootstrap:
   - lvm2
   - crudini
   - jq
-  - NetworkManager-ovs
   - nftables
   - openvswitch
   - openstack-selinux
--- a/roles/edpm_bootstrap/defaults/main.yml
+++ b/roles/edpm_bootstrap/defaults/main.yml
@@ -23,6 +23,7 @@ edpm_bootstrap_packages_bootstrap:
   - lvm2
   - crudini
   - jq
+  - NetworkManager-ovs
   - nftables
   - openvswitch
   - openstack-selinux
--- a/roles/edpm_bootstrap/defaults/main.yml
+++ b/roles/edpm_bootstrap/defaults/main.yml
@@ -27,7 +27,6 @@ edpm_bootstrap_packages_bootstrap:
   - nftables
   - openvswitch
   - openstack-selinux
-  - os-net-config
   - python3-libselinux
   - python3-pyyaml
   - rsync
--- a/roles/edpm_bootstrap/defaults/main.yml
+++ b/roles/edpm_bootstrap/defaults/main.yml
@@ -25,6 +25,7 @@ edpm_bootstrap_packages_bootstrap:
   - jq
   - nftables
   - openvswitch
+  - NetworkManager
   - openstack-selinux
   - python3-libselinux
   - python3-pyyaml
--- a/roles/edpm_network_config/defaults/main.yml
+++ b/roles/edpm_network_config/defaults/main.yml
@@ -18,7 +18,15 @@
 # All variables intended for modification should be placed in this file.
 
 # All variables within this role should have a prefix of "edpm_network_config"
-edpm_network_config_role: os-net-config
+
+# This variable let the use choose which tool will be used to configure networking on hosts
+# Accepted values are:
+# * os-net-config (default) -> the os-net-config will be used
+# * nmstate -> the network role from linux-system-roles/rhel-system-roles will be used to
+#   configure networking on hosts. In particular, the nmstate tool should cover all the
+#   supported scenario. Refer to nmstate.io for configuration snippets.
+edpm_network_config_tool: os-net-config
+
 edpm_network_config_update: false
 edpm_network_config_async_poll: 3
 edpm_network_config_async_timeout: 300
--- a/roles/edpm_network_config/defaults/main.yml
+++ b/roles/edpm_network_config/defaults/main.yml
@@ -18,6 +18,7 @@
 # All variables intended for modification should be placed in this file.
 
 # All variables within this role should have a prefix of "edpm_network_config"
+edpm_network_config_role: os-net-config
 edpm_network_config_update: false
 edpm_network_config_async_poll: 3
 edpm_network_config_async_timeout: 300
--- a/roles/edpm_network_config/tasks/main.yml
+++ b/roles/edpm_network_config/tasks/main.yml
@@ -19,7 +19,11 @@
     - name: Retrieve role name [nmstate]
       ansible.builtin.set_fact:
         systemrolename:
-          "{{ ansible_facts['distribution'] | lower }}.
+          "{%- if ansible_facts['distribution'] == 'RedHat' -%}
+            redhat
+          {%- else -%}
+            fedora
+          {%- endif -%}.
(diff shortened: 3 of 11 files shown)


roles/edpm_bootstrap/defaults/main.yml
```diff
@@ -23,6 +23,7 @@ edpm_bootstrap_packages_bootstrap:
   - lvm2
   - crudini
   - jq
+  - NetworkManager-ovs
```
slagle, line 26: Does the linux-system-role not make sure this is installed? If not, would that be considered a bug?
I'd move the package install into edpm_network_config when the system role is used.
And move the os-net-config package install into edpm_network_config and only installed when os-net-config is used.

roles/edpm_network_config/defaults/main.yml
```diff
@@ -18,6 +18,7 @@
 # All variables intended for modification should be placed in this file.
 
 # All variables within this role should have a prefix of "edpm_network_config"
+edpm_network_config_role: os-net-config
```
slagle, line 21: As this isn't an actual role name, I not use the _role suffix. Perhaps call it edpm_network_config_tool. I'd also document the choices as either os-net-config or nmstate.

slagle (commented):
couple comments, looking good to me otherwise.

roles/edpm_network_config/defaults/main.yml
rebtoor: ack.

roles/edpm_bootstrap/defaults/main.yml
jpodivin: We may want to keep a list of those. Maybe a jira task under some general NG topic?

roles/edpm_network_config/tasks/main.yml
```diff
@@ -14,136 +14,25 @@
...
+        systemrolename:
+          "{{ ansible_facts['distribution'] | lower }}.
+          {%- if ansible_facts['distribution'] == 'RedHat' -%}
+            rhel
+          {%- else -%}
+            linux
+          {%- endif -%}
+          _system_role.network"
+      delegate_to: localhost
+    - name: Configure network with nmstate
```
jpodivin, line 34: Name of the task should indicate that the role has been dynamically selected. It may cut down on the time needed to troubleshoot things.

roles/edpm_network_config/tasks/network_config.yml
```diff
@@ -0,0 +1,149 @@
...
+      (os_net_config_returncode_stat.stat.exists and
+      ((os_net_config_returncode_slurp.content | b64decode | int) != 0)) or
+       (not os_net_config_returncode_stat.stat.exists)
+  block:
+    - name: Create /etc/os-net-config directory
+      become: true
+      ansible.builtin.file:
+        path: /etc/os-net-config
+        state: directory
+        recurse: true
```
jpodivin, line 95: Setting permissions explicitly would be a good idea. We don't know what kind of umask settings the system has.

roles/edpm_network_config/tasks/network_config.yml
```diff
@@ -0,0 +1,149 @@
...
+        net_config_data_lookup:
+          "{{ edpm_network_config_os_net_config_mappings }}"
+      when: not ansible_check_mode|bool
+      register: os_net_config_mappings_result
+
+    - name: Write os-net-config mappings file /etc/os-net-config/mapping.yaml
+      ansible.builtin.copy:
+        content: "{{ os_net_config_mappings_result.mapping | to_nice_yaml }}"
+        dest: /etc/os-net-config/mapping.yaml
+        mode: 0644
```
jpodivin, line 108: Since this is a refactor, changing the permissions type to string, as per recommendations in ansible documentation [0] may be opportune. 


[0] https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html#parameter-mode

roles/edpm_network_config/tasks/network_config.yml
```diff
@@ -0,0 +1,149 @@
...
+
+    - name: Write rc of NetworkConfig script
+      ansible.builtin.copy:
+        content: "{{ network_config_result.rc }}"
+        dest: /var/lib/edpm-config/os-net-config.returncode
+        mode: 0644
+      when:
+        - network_config_result.rc is defined
+
+    # LP Bug: https://bugs.launchpad.net/edpm/+bug/1958332
```
jpodivin, line 129: This bug doesn't exist. Because it used to say tripleo. I suspect this is outcome of sed application. Maybe we should link back to the original [1]. Or we can replace this with a comment explaining why is this being done.

[1]https://bugs.launchpad.net/tripleo/+bug/1958332

jpodivin (changes requested):

roles/edpm_bootstrap/defaults/main.yml
rebtoor: Do you mean a list of required packages?

roles/edpm_network_config/tasks/main.yml
rebtoor: you're right, let me fix this

roles/edpm_network_config/tasks/network_config.yml
rebtoor: I didn't touch original code but I guess you're right.

rebtoor: also i don't understand why there's a `recurse:true` on a folder on root directory.

roles/edpm_network_config/tasks/network_config.yml
rebtoor: ack!

roles/edpm_network_config/tasks/network_config.yml
```diff
@@ -0,0 +1,154 @@
...
+    state: latest
+
+- name: Ensure /var/lib/edpm-config directory exists
+  become: true
+  ansible.builtin.file:
+    path: /var/lib/edpm-config
+    state: directory
+    setype: container_file_t
+    selevel: s0
+    mode: 0750
```
jpodivin, line 53: Sorry once again the mode should be quoted. But it is just a nitpick

jpodivin (approved):
One minor quibble. But other than that all has been resolved.

roles/edpm_network_config/tasks/network_config.yml
rebtoor: argh! I missed this one :facepalm:

rebtoor:
recheck

rebtoor:
recheck

roles/edpm_bootstrap/defaults/main.yml
jpodivin: Yeah. I mean we kind of do so already, but this one would be attached to a task to make their installation automated.

jpodivin (approved):

PR #184, opened 2023-06-08
Add edpm_bootstrap_command

slagle:
edpm_bootstrap_command is a variable for a shell command that will be
run early in the edpm_boostrap role tasks, before any package installs.
It can be used to register systems in order to enable package
repositories.

### DIFF
--- a/roles/edpm_bootstrap/tasks/bootstrap.yml
+++ b/roles/edpm_bootstrap/tasks/bootstrap.yml
@@ -72,6 +72,22 @@
   tags:
     - always
 
+- name: Bootstrap command
+  become: true
+  when: edpm_bootstrap_command != ""
+  block:
+    - name: Bootstrap command
+      shell: "{{ edpm_bootstrap_command }}"
+      register: bootstrap_cmd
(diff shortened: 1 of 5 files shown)


fultonj (commented):
+2

roles/edpm_bootstrap/tasks/bootstrap.yml
```diff
@@ -90,6 +90,22 @@
       when:
         - rpm_query_result.rc > 0
 
+- name: Bootstrap command
```
kajinamit, line 93: If we assume this is used to run some commands like subscription-manager then this should be executed before L76, because that task installs the rhosp-release package

kajinamit (changes requested):

roles/edpm_bootstrap/tasks/bootstrap.yml
slagle: Indeed! Not sure how I missed that. Thanks, i'll move it earlier.

rebtoor (approved):

PR #185, opened 2023-06-09
Remove var files for CentOS 8 and Fedora

kajinamit:
... because we do not support these operating systems.


jpodivin:
Rebase may be in order.

jpodivin:
In fact it may be better to close this entirely. The #182 will remove the role altogether when it is merged, instead we will be using the ci-framework equivalent.

kajinamit:
Closing in favor of https://github.com/openstack-k8s-operators/edpm-ansible/pull/182

PR #187, opened 2023-06-09
Add tripleo-repos package to Containerfile.j2 template

Deydra71:
The tripleo-repos are now installed in container generally, in contrary with the test_deps role. Template was generated and successfully tested in local CentOS9 container, and the installation of required repos was successfull.

### DIFF
--- a/molecule/common/Containerfile.j2
+++ b/molecule/common/Containerfile.j2
@@ -5,8 +5,10 @@ FROM {{ item.registry.url }}/{{ item.image }}
 FROM {{ item.image }}
 {% endif %}
 
-RUN dnf upgrade -y && \
+RUN curl -o /etc/yum.repos.d/delorean.repo https://trunk.rdoproject.org/centos9-wallaby/current/delorean.repo && \
+    dnf upgrade -y && \
     dnf -y install sudo python3-libselinux selinux-policy {{ item.pkg_extras | default('') }} && \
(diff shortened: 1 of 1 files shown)


molecule/common/Containerfile.j2
```diff
@@ -9,4 +9,8 @@ RUN dnf upgrade -y && \
     dnf -y install sudo python3-libselinux selinux-policy {{ item.pkg_extras | default('') }} && \
     dnf clean all -y
 
+RUN curl -o /etc/yum.repos.d/delorean.repo https://trunk.rdoproject.org/centos9-wallaby/current/delorean.repo
+
+RUN dnf install -y python*tripleo-repos
```
rebtoor, line 14: I'd move the `dnf clean all -y` from the first RUN block here after the installation of `python*tripleo-repos` package.

jpodivin: Yes that may be for the better.

jpodivin (changes requested):
Maybe you could first install repos, then all the packages.

molecule/common/Containerfile.j2
Deydra71: Agree, I will merge it into one RUN block also preserve readiness. Thanks!

jpodivin (approved):

PR #188, opened 2023-06-12
Added task to restart NM after plugin installation and improved `systemrolename` gathering

rebtoor:
The restart task is needed to avoid the error: 

```libnmstate.error.NmstateDependencyError: Manager(MissingPlugin):NetworkManager plugin for 'ovs-bridge' unavailable"```

despite the plugin is already installed.

The `systemrolename` fact gathering has been improved since an if/else conditional was redundant.


bshephar (approved):
lgtm although, do we need to have fedora in here at all?

PR #192, opened 2023-06-13
Setting correct defaults for all compute nodes on telemetry playbooks…

jlarriba:
… execution


rebtoor (approved):

jlarriba:
It seems that the github-check is failing on provisioning the environment, and does not seem to be related to my change.

PR #194, opened 2023-06-14
Restart NetworkManager only when needed and renamed task

rebtoor:
It's not necessary and potentially dangerous to restart NetworkManager if it's not needed, so we will check the state of the dnf task to monitor if it's changed.

Also, the task name now explains what is going to be installed.


rebtoor:
recheck

rebtoor:
recheck

PR #195, opened 2023-06-21
Added sentinel file to check if rules are changed so they can be applied

rebtoor:
Also:
* add owner/group/mode for each created file to improve security
* perform rules validation using privilege escalation otherwise ruleset files couldn't be read from standard user

This fix is needed since we are now executing `bootstrap`, `configure` and `run` steps in different ansible executions, so the variables (`nft_ruleset` in this case) aren't passed to next steps.

### DIFF
--- a/roles/edpm_nftables/tasks/configure.yml
+++ b/roles/edpm_nftables/tasks/configure.yml
@@ -23,7 +23,7 @@
         state: directory
         owner: root
         group: root
-        mode: 0750
+        mode: '0750'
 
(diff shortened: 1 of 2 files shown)


rebtoor:
recheck

rebtoor:
recheck

rebtoor:
recheck

fultonj (approved):

rebtoor:
recheck

jpodivin (approved):

PR #198, opened 2023-06-27
Multiple roles were added to the local test skip list

jpodivin:
Several roles were added to the local test script skip list as their successful testing in local environment couldn't be guaranteed.


rebtoor:
recheck

PR #200, opened 2023-06-28
Changed root size in growvols defaults

rebtoor:
Since we're creating a swapfile when a swap partition isn't present due to https://github.com/openstack-k8s-operators/edpm-ansible/pull/156, let's extend the root partition size to reserve some space for the swapfile itself.

### DIFF
--- a/roles/edpm_growvols/defaults/main.yml
+++ b/roles/edpm_growvols/defaults/main.yml
@@ -16,7 +16,7 @@
 
 
 growvols_args: >
-  /=8GB
+  /=15GB
   /tmp=1GB
   /var/log=10GB
   /var/log/audit=2GB


roles/edpm_growvols/defaults/main.yml
```diff
@@ -16,7 +16,7 @@
 
 
 growvols_args: >
-  /=8GB
+  /=15GB
```
rabi, line 19: That won't work, I guess the edpm job failed for that.

rabi: The failure is different though. We've a compute disk size of 20GB[1] for pre-provisioned nodes it seems.

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/3813682af65b599aba1e8e17e638854ee27ff5b4/devsetup/scripts/gen-edpm-compute-node.sh#L27

rebtoor: ok! 

So it could have more sense to reduce the default swapfile size to 1/2 GB instead of changing the % of space allocated to var?

steveb: Could you clarify that swap is only required on CI jobs which have constrained memory? If so, changing the defaults here would not be appropriate. These values are the defaults for real deployments with appropriate amounts of memory where swap is not required (or desirable?).

I would suggest job definitions should pass in custom values of growvols_args based on the available storage and memory of the target test resources.

steveb (changes requested):

roles/edpm_growvols/defaults/main.yml
steveb: 8GB is aimed to be a big enough root for it not filling up in the lifecycle of a deployed node. If there is an argument for making that bigger then we could increase for those reasons, and that would give some extra space for a swapfile in short-lived resource constrained CI nodes.

rabi: Now it won't work for edpm baremetal jobs[1]. Earlier result was false positive as PRs were not tested.

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/Makefile#L11

rabi:
recheck

roles/edpm_growvols/defaults/main.yml
rebtoor: I reduced the swapfile size from 4Gb to 1Gb (https://github.com/openstack-k8s-operators/edpm-ansible/pull/205/files) but it's still not enough.

@rabi can we increase BMAAS_DISK_SIZE or is it a constraint?

PR #201, opened 2023-06-28
Full documentation for edpm_network_config role endpoint

jpodivin:
Each variable now has a type, default value and basic info.

Several variables were removed, as they are not used by the role.

Namely:
 - edpm_network_config_with_ansible

Typo in the edpm_os_net_config module was fixed.

### DIFF
--- a/roles/edpm_network_config/defaults/main.yml
+++ b/roles/edpm_network_config/defaults/main.yml
@@ -38,7 +38,6 @@ edpm_network_config_manage_service: true
 edpm_network_config_nmstate: false
 edpm_network_config_os_net_config_mappings: {}
 edpm_network_config_safe_defaults: true
-edpm_network_config_with_ansible: true
 edpm_network_config_template: templates/single_nic_vlans/single_nic_vlans.j2
 edpm_network_config_override: {}
 edpm_bond_interface_ovs_options: "bond_mode=active-backup"
--- a/plugins/modules/edpm_os_net_config.py
+++ b/plugins/modules/edpm_os_net_config.py
@@ -64,7 +64,7 @@ options:
   safe_defaults:
     description:
       - If enabled, safe defaults (DHCP for all interfaces) will be applied in
-        case of failuring while applying the provided net config.
+        case of failing while applying the provided net config.
     type: bool
     default: false
   use_nmstate:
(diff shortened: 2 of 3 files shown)


roles/edpm_network_config/defaults/main.yml
```diff
@@ -30,15 +30,12 @@ edpm_network_config_tool: os-net-config
 edpm_network_config_update: false
 edpm_network_config_async_poll: 3
 edpm_network_config_async_timeout: 300
-edpm_network_config_bridge_name: br-ex
```
rebtoor, line 33: could it be that we should rename `neutron_physical_bridge_name` ?

roles/edpm_network_config/defaults/main.yml
```diff
@@ -30,15 +30,12 @@ edpm_network_config_tool: os-net-config
...
 edpm_network_config_debug: "{{ (ansible_verbosity | int) >= 2 | bool }}"
 edpm_network_config_hide_sensitive_logs: true
-edpm_network_config_interface_name: nic1
```
rebtoor, line 36: as above but related to `neutron_public_interface_name`

rebtoor (commented):
left a couple of comments

roles/edpm_network_config/defaults/main.yml
jpodivin: I would argue that it's easier to rename it here, in one place, rather than everywhere else. The `neutron_physical_bridge_name` also seems more descriptive.

rebtoor: yeah but it would be better to have the `edpm_network_config_` prefix just like the other vars

rebtoor: I agree that we should put this PR on hold, and i'm surprised to see that the var is widely used but it's not documented and doesn't have a "default" in the role where it should be used.

jpodivin:
Ok. All of the dependencies have PR proposed. The director operator and director operator dev tools are only exceptions because those have files explicitly marked as 17.0

jpodivin:
recheck

jpodivin:
recheck

rebtoor:
recheck

rebtoor:
recheck

jpodivin:
recheck

slagle (approved):

jpodivin:
recheck

jpodivin:
recheck

jpodivin:
recheck

slagle (approved):

PR #204, opened 2023-06-29
Check if we have required python version available before running tests

jpodivin:
Collection was built with python 3.9 in mind, however it is possible that when molecule tests are executed on a local machine other python versions may be used to create the venv.

This patch introduces a mechanism, which checks if the appropriate python version is present on the system, and if not prints a warning indicating that the default will be used instead.


rebtoor (approved):

rebtoor:
recheck

jpodivin:
recheck

rebtoor:
recheck

jpodivin:
recheck

jpodivin:
recheck

PR #205, opened 2023-06-29
Refactored swap Playbook in ansible style and changed swapfile default size

rebtoor:
* "swap" playbook has been refactored to avoid shell/command tasks when/if possible
* Swapfile default size has been reduced to 1Gb


jpodivin (approved):
/lgtm

jpodivin (approved):

PR #206, opened 2023-07-03
Removing ununsed debug args

jpodivin:
Many of our roles have debug arguments with no effect on the role behavior. These arguments are often set by referring to the ansible runtime verbosity level. In cases when these arguments are not used anywhere else in edpm_ansible collection or in the wider codebase, I believe it's safe to remove them.


rebtoor:
+2

since those variables didn't trigger an actual "debug" or, at least, "increased verbosity", i totally agree to remove them.

rebtoor (commented):
+2

since those variables didn't trigger an actual "debug" or, at least, "increased verbosity", i totally agree to remove them.

fao89 (approved):

PR #209, opened 2023-07-07
Enhance instructions for testing local changes

ASBishop:
- Add instructions for testing individual roles with molecule
- Add instructions to test local changes via ansibleee using NFS

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,15 @@ To push the image:
 Depending on the repository, a ``podman login quay.io/<user>`` may be required
 before pushing.
 
+Run ansibleee pods with a local copy of edpm-ansible
+----------------------------------------------------
+
+Local changes to edpm-ansible can also be tested by adding a volume mount to
+the ansibleee pods. This will be faster than building a new
+openstack-ansibleee-runner container image. See the [edpm-ansible
+documentation](https://openstack-k8s-operators.github.io/edpm-ansible/) for
+more information.
+
 License
 -------
 
--- a/docs/source/contributing_roles.rst
+++ b/docs/source/contributing_roles.rst
@@ -22,13 +22,44 @@ stub at the `../collections/osp/edpm/edpm_<NEWROLENAME>_role.rst` path.
 Optionally you can write further information about role operation.
 Including section of examples and molecule tests.
 
-Local testing of new roles
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+Testing roles with molecule
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Launch `make execute_molecule` to test all collections roles.
 
-TBD: test single role
+Roles can be tested individually using the *molecule-venv* created by running
+`make execute_molecule`. The *molecule-venv* can also be manually created.
 
+.. code-block:: console
+
+    % python3 -m venv molecule-venv
+    % source molecule-venv/bin/activate
+    (molecule-venv) % pip install --upgrade pip
+    (molecule-venv) % pip install molecule molecule-podman jmespath
+
+Use the *molecule-venv* to test a specific role.
+
+.. code-block:: console
+
+    (molecule-venv) % cd roles/edpm_<role>
+    (molecule-venv) % molecule test # tests default scenario
+    (molecule-venv) % molecule test --all
+    (molecule-venv) % molecule test --scenario-name <specific scenario>
+
+Testing roles with ansibleee
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+edpm-ansible is included in the openstack-ansibleee-runner container image,
+which is key component used by the dataplane-operator that deploys EDPM
+nodes. The dataplane-operator's CRD includes support for specifying additional
+volume mounts for the ansibleee pods, which provides a mechanism for accessing
+a local copy of edpm-ansible. This makes it possible to develop and test local
+changes edpm-ansible without having to build and deploy a new
+openstack-ansibleee-runner container image.
+
+.. toctree::
+
+   testing_with_ansibleee
 
 Contributing plugins
 ~~~~~~~~~~~~~~~~~~~~
--- a/docs/source/testing_with_ansibleee.rst
+++ b/docs/source/testing_with_ansibleee.rst
@@ -1,0 +1,113 @@
+======================
+Testing with ansibleee
+======================
+
+Provide NFS access to your edpm-ansible directory
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The technique described here uses NFS to access the edpm-ansible directory on
+your development system, so you'll need to install an NFS server and create
+an appropriate export on your development system. Of course, this implies
+your OpenShift deployment that runs the dataplane-operator has access to
+the NFS server, including any required firewall rules.
+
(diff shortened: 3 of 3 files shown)


testing_local_changes.md
```diff
@@ -0,0 +1,115 @@
...
+your OpenShift deployment that runs the dataplane-operator has access to
+the NFS server, including any required firewall rules.
+
+* [EL 8 instructions](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/deploying_different_types_of_servers/exporting-nfs-shares_deploying-different-types-of-servers#assembly_configuring-the-nfs-server-to-run-behind-a-firewall_exporting-nfs-shares)
+* [EL 9 instructions](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/managing_file_systems/exporting-nfs-shares_managing-file-systems#assembly_configuring-the-nfs-server-to-run-behind-a-firewall_exporting-nfs-shares)
+
+When using OpenShift Local (aka CRC), your export will be something like this:
+
+```
+% cat &lt;&lt;EOF &gt;/etc/exports
```
rebtoor, line 25: is it `cat <<EOF >/etc/exports` ?

rebtoor (changes requested):
+2 for the general idea, but I'd like that the `testing_local_changes.md` file become a page published under https://openstack-k8s-operators.github.io/edpm-ansible/
so I'd convert it in a .rst file under docs/source folder.

testing_local_changes.md
ASBishop: It will be when viewed as html. The problem with markdown is '<' characters get confused with html tags (even in literal code blocks), and using "&lt;" allows the page to be rendered correctly.

But I'll follow your suggestion and move this to an rst file in the docs area, where the conversion to html isn't an issue.

testing_local_changes.md
```diff
@@ -0,0 +1,115 @@
...
+% NFS_SHARE=&lt;Path to your edpm-ansible directory&gt;
+% NFS_SERVER=&lt;IP of your NFS server&gt
+% cat &lt;&lt;EOF &gt;edpm-ansible-storage.yaml
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: edpm-ansible
+spec:
+  capacity:
+    storage: 1Gi
```
SeanMooney, line 61: i assume the size here dose not really matter as long as the PVC matches correct.

SeanMooney:
i took a slightly different approach to this type of testing in the nova operator by the way.

the normal workflow given the nova operator create the playbooks as a config map by reading them from disk when it starts was just run the operator then watch it reconsile.
that is fine but it does not actually allow you to write tests for the playbook functionality.

to speed up operator development and improve testing i also create a moldecue scenario to emulate the way they are exctuted

https://github.com/openstack-k8s-operators/nova-operator/tree/main/molecule/default

so that creates 2 vms one to act as the AEE/DAtapalen operator and one to ack as the edpm node.

the workflow in the converge playbook was pretty simple
first we stage everything in the AEE VM
https://github.com/openstack-k8s-operators/nova-operator/blob/main/molecule/default/converge.yml#L2-L50

then we did the minium amount of prep on the edpm VM to fullfle the requirement of our playbook
https://github.com/openstack-k8s-operators/nova-operator/blob/main/molecule/default/converge.yml#L52C1-L75C1

and finally we just run the AEE container with the playbook our play vai podman
https://github.com/openstack-k8s-operators/nova-operator/blob/main/molecule/default/converge.yml#L77-L102

so basically you could molecule converge and that would create the dev env to allow you to develop a playbook/role using the local content and would execute it against the edpm VM.

this could be simply recreated by doing molecule destroy and molecule converge to start clean if needed and you could ssh in and inspect the ndoes with molecule login.

i also added a verification plybook so you could test the role with molecule test https://github.com/openstack-k8s-operators/nova-operator/blob/main/molecule/default/verify.yml

and some test helpers.

The procedure in this pr is useful  if you want to test using the dataplane operator but i also belive we could improve the use of molecule in the roles directly and not require the dataplane operator at all.

testing_local_changes.md
```diff
@@ -0,0 +1,115 @@
...
+          - extraVolType: edpm-ansible
+            mounts:
+            - mountPath: /usr/share/ansible/collections/ansible_collections/osp/edpm
+              name: edpm-ansible
+            volumes:
+            - name: edpm-ansible
+              persistentVolumeClaim:
+                claimName: edpm-ansible
+                readOnly: true
+```
```
SeanMooney, line 115: ack so with this extra mount deffitnion you have effectly replace the ansibel collection in the AEE pod that is spawned win a role invokes a OpenStackDatapalne service.

it will instead use the source form the nfs share via the pvc.

testing_local_changes.md
```diff
@@ -0,0 +1,115 @@
...
+  capacity:
+    storage: 1Gi
+  volumeMode: Filesystem
+  accessModes:
+    - ReadOnlyMany
+  # IMPORTANT! The persistentVolumeReclaimPolicy must be "Retain" or else
+  # your code will be deleted when the volume is reclaimed!
+  persistentVolumeReclaimPolicy: Retain
+  storageClassName: edpm-ansible
+  mountOptions:
```
SeanMooney, line 69: when we dicussed this breifly on the call you also confirmed that
while this example is using nfs you could in theory use cephfs or any other volume driver that could fullfile the access mode in this case ReadOnlyMany (ROX)

NFS is definally on of the simpler solution to meet that crtia 

in terms of the mount option you could is - hard actually required
hard mounts could cause block io issue in the oepnshif cluster if the nfs connection was to drop but for development that probably ok.

testing_local_changes.md
ASBishop: Correct.

testing_local_changes.md
ASBishop: Yes, the ansibleee pod will use your local edpm-ansible sources.

testing_local_changes.md
ASBishop: NFS was the simplest, logical choice for me because I already had the NFS server infrastructure in place (for testing an NFS cinder backend). I'm sure there are other possibilities.

Yeah, I think the "hard" mount option was just part of another example I looked at when I started playing with this, and agree it's not something you'd want to do in a production environment.

ASBishop:
> i took a slightly different approach to this type of testing in the nova operator by the way.

I think this is a case of "different horses for different courses." I feel I have a decent environment for molecule testing, and want a way to test my edpm_changes in something approaching a "real world" environment. All of these approaches are valid, and it might be useful to document yours as an alternative to mine.

testing_local_changes.md
SeanMooney: ack

SeanMooney:
> > i took a slightly different approach to this type of testing in the nova operator by the way.
> 
> I think this is a case of "different horses for different courses." I feel I have a decent environment for molecule testing, and want a way to test my edpm_changes in something approaching a "real world" environment. All of these approaches are valid, and it might be useful to document yours as an alternative to mine.

I'm starting on a new edpm_libvirt role currently and i plan to use the vagrant VM approach for its molecule even when I'm porting the playbook form the nova operator into a role.

so can can serve as a reference i think.

i was not suggesting we don't document your approach by the way i just wanted to share what i did incase it was useful to others.

rebtoor (changes requested):
Hey @ASBishop thanks a lot! 🙇 

Can you add `testing_with_ansibleee` in the index.rst file so it will be linked in the home page?

rebtoor (approved):

rebtoor:
recheck

PR #211, opened 2023-07-11
Use ansible_hostname rather than fqdn for hostname

rabi:
I think it should be the nodeName[1] and not fqdn. 
$$echo $(oc get pod ovn-controller-6l7t2 -o "jsonpath={.spec.nodeName}")
crc-8cf2w-master-0

[1] https://github.com/openstack-k8s-operators/ovn-operator/blob/main/pkg/ovncontroller/configjob.go#L68


bshephar (approved):

rabi:
recheck

PR #214, opened 2023-07-11
Consolidating edpm_podman arguments

jpodivin:
edpm_buildah_package and edpm_podman_package were transformed from list to string variables

edpm_podman_insecure_registries was merged with edpm_container_registry_insecure_registries

edpm_podman_purge_packages var was removed, as it wasn't used by the role at all


rebtoor (commented):
+2

jpodivin:
recheck

fao89 (approved):

PR #215, opened 2023-07-14
port deploy-libvirt playbook to edpm_libvirt role

SeanMooney:
This change ports the deploy-libvirt playbook form
the nova operator and converts it into a standalone
role.

This includes a molecule testing leveraging the
vagrant molecule plugin to create a vm for testing.
basic verifcation tests are also provided to ensure
the container are created, running and defiend as
systemd services and ensure the expected host
directories are created.


roles/edpm_libvirt/molecule-requirements.txt
SeanMooney, line 1: if perople are wondering why i have a second molecule-requiments.txt and i am not just adding vagrant to the other one its because for some reason with the newer ansible the vagrant plugin cant find ansible.molecule_utils.common.yaml
even though i can create a python shell and import it and even though i can see the file in the right location.

i spend two days in the last 3 weeks trying to make that work so for the sake fo time i am just clamping to a know working version to move this forward but we can loop back at a later date.

zuul.d/jobs.yaml
```diff
@@ -56,3 +56,24 @@
     name: edpm-ansible-crc-podified-edpm-baremetal
     parent: cifmw-crc-podified-edpm-baremetal
     irrelevant-files: *molecule_irrelevant_files
+
+# non-containerized role jobs
+- job:
+    name: edpm-ansible-edpm_libvirt
+    description: |
+      This job tests installing libvirt via the edpm_libvirt role.
+    parent: cifmw-molecule-base
```
SeanMooney, line 65: i think this need required-project for the ci framework

https://github.com/openstack-k8s-operators/ci-framework/blob/d778a092ab6f6c0031f9edce6804573188a2af26/ci/playbooks/molecule-prepare.yml#L22-L25

slagle (commented):
+2, LGTM.

jpodivin:
recheck

rebtoor:
recheck

PR #217, opened 2023-07-17
edpm nova

SeanMooney:
- port deploy-libvirt playbook to edpm_libvirt role
- add tags and update bindmoutnts
- Port nova playbook to edpm_nova role


roles/edpm_libvirt/meta/argument_specs.yml
```diff
@@ -0,0 +1,7 @@
+---
+argument_specs:
+  # ./roles/edpm_libvirt/tasks/main.yml entry point
+  main:
+    short_description: The main entry point for the edpm_libvirt role.
+    description: Multiple lines description
+    options: {}
```
jpodivin, line 7: Role arg should be documented here. If they are all optional we can mark them as such.

roles/edpm_libvirt/molecule-requirements.txt
```diff
@@ -0,0 +1,8 @@
+ansible-core<2.12.0
```
jpodivin, line 1: If possible we should centralize requirements in common file we have for that purpose.

roles/edpm_libvirt/tasks/post-install.yml
```diff
@@ -0,0 +1,44 @@
...
+    </usage>
+    </secret>
+    EOF
+    podman cp secret.xml libvirt_virtqemud:/tmp/secret.xml
+    rm -f secret.xml
+    podman exec libvirt_virtqemud bash -c "virsh secret-undefine ${CLUSTER_FSID_MAP[${cluster}]}"
+    podman exec libvirt_virtqemud bash -c "virsh secret-define --file /tmp/secret.xml; rm -f /tmp/secret.xml"
+    keyring=$(awk '$1 == "key" {print $3}' /etc/ceph/${cluster}.client.${CLUSTER_USER_MAP[$cluster]}.keyring)
+    podman exec libvirt_virtqemud bash -c "virsh secret-set-value ${CLUSTER_FSID_MAP[${cluster}]} --base64 ${keyring}"
+    done
```
jpodivin, line 28: Please declare that this operation changes state of the machine.

roles/edpm_libvirt/tasks/post-install.yml
```diff
@@ -0,0 +1,44 @@
...
+- name: Copy qemu vnc firewall config
+  tags:
+    - install
+    - post-libvirt
+  become: true
+  ansible.builtin.template:
+    src: "firewall.yaml"
+    dest: "/var/lib/edpm-config/firewall/vnc.yaml"
+    mode: "640"
```
jpodivin, line 37: Mode should be declared in quotes, but more importantly it should be declared in the same way across the role.

roles/edpm_libvirt/templates/firewall.yaml
```diff
@@ -0,0 +1,7 @@
+---
```
jpodivin, line 1: Is this an example of the template rendering? If so, it should be moved into docs.

roles/edpm_nova/molecule-requirements.txt
```diff
@@ -0,0 +1,8 @@
+ansible-core<2.12.0
```
jpodivin, line 1: Reqs should reside in one place, if possible.

roles/edpm_nova/tasks/configure.yml
```diff
@@ -0,0 +1,75 @@
...
+    owner: "{{ item.owner | default(omit) }}"
+    group: "{{ item.group | default(omit) }}"
+    mode: "{{ item.mode | default(omit) }}"
+  with_items:
+  - { 'path': /var/lib/nova}
+  - { 'path': /var/lib/_nova_secontext}
+  - { 'path': /var/lib/nova/instances}
+  - { 'path': /var/log/containers/nova, 'mode': '0750' }
+  - { 'path': /var/log/containers/stdouts }
+  - { "path": /etc/ceph, "mode": "0750", "owner": "root", "group": "root"}
```
jpodivin, line 33: Nitpick, but the quotes should be same across the file. It's easier on the eye.

roles/edpm_nova/tasks/post-install.yml
jpodivin, line 1: Is this empty file necessary?

roles/edpm_nova/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+- name: configure nova
+  include_tasks: configure.yml
+
+- name: install nova
+  include_tasks: install.yml
+
+- name: Flush handlers
+  meta: flush_handlers
+
+- name: post-install nova
```
jpodivin, line 27: The file is empty so nothing is executed in this step.

jpodivin (changes requested):
Roles need testing in CI. Without that we can't merge them. I would suggest zuul, it's probably easier to fit there than in containers.

roles/edpm_libvirt/molecule-requirements.txt
SeanMooney: we cant in this case as i noted in the libvirt PR.
there is a depenciy issue with the vagrant molecule plugin.
https://github.com/openstack-k8s-operators/edpm-ansible/pull/215#discussion_r1263630455

roles/edpm_libvirt/templates/firewall.yaml
SeanMooney: no it is not.
it is the template that is used to render the firewall config
not an example

roles/edpm_libvirt/tasks/post-install.yml
SeanMooney: you mean add changed=true

roles/edpm_nova/meta/argument_specs.yml
```diff
@@ -0,0 +1,10 @@
+---
+argument_specs:
+  # ./roles/edpm_nova/tasks/main.yml entry point
+  main:
+    short_description: The main entry point for the edpm_nova role.
+    description: Multiple lines description
+    options:
+       edpm_nova_config_src:
+        type: str
```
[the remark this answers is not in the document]
SeanMooney: yep its on my todo list.

i will be updating the libvirt part in the libvirt role PR #215 . the nova PR (this one) will be updated after

roles/edpm_nova/molecule-requirements.txt
SeanMooney: yep as noted its not possibel currently as we cannot use the same version fo ansible due to python path issues

https://github.com/openstack-k8s-operators/edpm-ansible/pull/215#discussion_r1263630455

i do not have time to try and unify them in the short term.

roles/edpm_nova/tasks/post-install.yml
SeanMooney: i wanted to keep the pattern constient so i intnially included this empty file

roles/edpm_nova/tasks/main.yml
SeanMooney: yep that is intentional
i wanted to keep the pattern consistent

SeanMooney:
> Roles need testing in CI. Without that we can't merge them. I would suggest zuul, it's probably easier to fit there than in containers.

the molecule tests are using libvirt vms precicly because we cannot test them in containers.
i was considering how to run this in ci as i dotn think the ci framework jobs will work.

given we do not have ci coverage for the existing libvirt and nova roles i was hoping that would not be a blocker initially but i can look at creating a zuul job.

roles/edpm_libvirt/tasks/post-install.yml
jpodivin: Yep.

roles/edpm_nova/templates/config.json.j2
```diff
@@ -0,0 +1,41 @@
+{
+    "command": "nova-compute",
```
slagle, line 2: doesn't this need to be updated so that it reads all config files from /etc/nova, or does that happen by default?

slagle:
I'm testing this with https://github.com/openstack-k8s-operators/dataplane-operator/pull/297. It looks like this includes https://github.com/openstack-k8s-operators/edpm-ansible/pull/215, is that correct?

SeanMooney:
@slagle yes it contains #215 but this PR is out of date.
i have not made any changes that you need to care about in #215 that are not in #217.

i pushed up libvirt and nova as two separate PRs.
i have been working on ci and arg specs in #215 and then i was going to rebase this after once i had that working.

for your integration i think you can use this slightly older version and it should not matter.
i will keep working on #215 and once that is ready ill rebase and update this but you have both roles in this PR.

roles/edpm_nova/templates/config.json.j2
slagle: Looks like no. It seems --config-dir /etc/nova/nova.conf.d is the default behavior, is that right?

roles/edpm_libvirt/meta/argument_specs.yml
SeanMooney: yep as noted prevsiouly this role is part of a separate PR please review it there #215 instead not here

roles/edpm_nova/templates/config.json.j2
SeanMooney: corret it does not.

this is part of oslo oslo will look in /etc/nova and /etc/nova/nova.conf.d by default.
when we register the config options we tell it the project name e.g. "nova"
it then searches a set of directors to check incluign /etc/<project> and /etc/<project>/<project>.conf.d/

so we don't need to pass it here provided we place the files in the default locations
we we do below.

roles/edpm_nova/tasks/post-install.yml
SeanMooney: ill proably remove this since it seams to be causing confution

slagle:
> @slagle yes it contains #215 but this PR is out of date. i have not made any changes that you need to care about in #215 that are not in #217.
> 
> i pushed up libvirt and nova as two separate PRs. i have been working on ci and arg specs in #215 and then i was going to rebase this after once i had that working.
> 
> for your integration i think you can use this slightly older version and it should not matter. i will keep working on #215 and once that is ready ill rebase and update this but you have both roles in this PR.

This (217) one lgtm. I've tested it in https://github.com/openstack-k8s-operators/dataplane-operator/pull/297, and that's working.

slagle (commented):
+2, LGTM.

jpodivin (approved):
lgtm

jpodivin:
recheck

molecule/test-helpers/verify_podman.yaml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Check the container stdout log exists {{ item }}
+      become: true
+      ansible.builtin.stat:
+        path: "/var/log/containers/stdouts/{{ item }}.log"
+      register: log_exists
+    - name: Assert the container stdout log exists {{ item }}
+      ansible.builtin.assert:
+        that:
+          - log_exists.stat.exists
+        fail_msg: "Container stdout log for {{ item }} does not exist"
```
bshephar, line 35: What's the iterator here? Using `{{ item }}` but no `loop:`?

SeanMooney: you cannot use loop with block, this is a well know limitation in ansible
so the convention is to put the blocks in separate task files and then invoke that in a loop

see the verify.yml for where this is used

roles/edpm_libvirt/molecule/default/verify.yml
```diff
@@ -0,0 +1,42 @@
...
+
+- name: Verify
+  hosts: compute-1
+  gather_facts: false
+  vars:
+    test_helper_dir: "../../../../molecule/test-helpers"
+  tasks:
+    - name: ensure expected directories exist
+      ansible.builtin.include_tasks: "{{test_helper_dir}}/verify_dir.yaml"
+      with_items:
```
SeanMooney, line 12: this with_items: is what defines the itmes var when we include the test helpers.

slagle (approved):

PR #219, opened 2023-07-19
Add new network template

cjeanner:
With the new CI layout, having a dedicated nodeset for CRC alongside a
dedicated Compute, we need to get a dedicated network template file in
order to consume the right interface - eth1, aka nic2.

That "nic2" interface is associated to a private, dedicated network in
order to allow network isolation to proceed, exactly like late OVB jobs
in TripleO CI ecosystem.

This patch therefore adds a new "ci" dedicated template we can then
consume from within the ci-framework, ensuring we're using the correct
interface.

Note: this is hopefully a temporary workaround - the ansibleVars we
should use is, for now, a string, and we can't edit it with kustomize.
There's a plan to change its type, allowing proper edition of the
mapping.

Signed-off-by: Cédric Jeanneret <cjeanner@redhat.com>

### DIFF
--- a/roles/edpm_network_config/templates/ci/extracted_crc_nic2_vlans.j2
+++ b/roles/edpm_network_config/templates/ci/extracted_crc_nic2_vlans.j2
@@ -1,0 +1,32 @@
+---
+{# This template is for extracted CRC layout -#}
+{% set mtu_list = [ctlplane_mtu] %}
+{% for network in role_networks %}
+{{ mtu_list.append(lookup('vars', networks_lower[network] ~ '_mtu')) }}
+{%- endfor %}
+{% set min_viable_mtu = mtu_list | max %}
+network_config:
+- type: ovs_bridge
+  name: {{ neutron_physical_bridge_name }}
+  mtu: {{ min_viable_mtu }}
+  use_dhcp: false
+  dns_servers: {{ ctlplane_dns_nameservers }}
+  domain: {{ dns_search_domains }}
+  addresses:
+  - ip_netmask: {{ ctlplane_ip }}/{{ ctlplane_subnet_cidr }}
+  routes: {{ ctlplane_host_routes }}
+  members:
+  - type: interface
+    name: nic2
+    mtu: {{ min_viable_mtu }}
+    # force the MAC address of the bridge to this interface
+    primary: true
+{% for network in role_networks %}
+  - type: vlan
+    mtu: {{ lookup('vars', networks_lower[network] ~ '_mtu') }}
+    vlan_id: {{ lookup('vars', networks_lower[network] ~ '_vlan_id') }}
+    addresses:
+    - ip_netmask:
+        {{ lookup('vars', networks_lower[network] ~ '_ip') }}/{{ lookup('vars', networks_lower[network] ~ '_cidr') }}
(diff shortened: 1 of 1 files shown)


roles/edpm_network_config/templates/ci/extracted_crc_nic2_vlans.j2
```diff
@@ -0,0 +1,31 @@
...
+  mtu: {{ min_viable_mtu }}
+  use_dhcp: false
+  dns_servers: {{ ctlplane_dns_nameservers }}
+  domain: {{ dns_search_domains }}
+  addresses:
+  - ip_netmask: {{ ctlplane_ip }}/{{ ctlplane_subnet_cidr }}
+  routes: {{ ctlplane_host_routes }}
+  members:
+  - type: interface
+    name: nic2
```
rabi, line 19: > order to consume the right interface - eth1, aka nic2.

You don't need a separate template in the role added for CI for that. It can be set with edpm_os_net_config_mappings/net_config_data_lookup ansible var[1].

[1] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/plugins/modules/edpm_os_net_config_mappings.py#L67-L69

rabi (changes requested):

roles/edpm_network_config/templates/ci/extracted_crc_nic2_vlans.j2
cjeanner: hm, overriding the "meaning" of nic1 then?

cjeanner:
Following @rabi 's indication, ci-framework now pushes a small snippet adding the mapping. I can close this PR then. Thanks for the hint!

cjeanner:
After some discussions and checks, we can't use kustomize against the `ansibleVars` because it's basically a string:
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/api/v1beta1/common.go#L54

There's apparently plan to change this parameter type, but until it's done, we have to get our proper mapping in in order to move on and get the new layout working.

Once the parameter type is changed, we'll be able to leverage kustomize in order to make the needed changes.

roles/edpm_network_config/templates/ci/extracted_crc_nic2_vlans.j2
cjeanner: As written bellow - we can't use that for now, because of the type associated to `ansibleVars`... basically, we'd override the whole value, because it's a string, not a proper mapping :(.

rabi:
> After some discussions and checks, we can't use kustomize against the ansibleVars because it's basically a string

I think just add a change to install_yamls[1] add the var should be enough

rebtoor (approved):

cjeanner:
/hold
`error: timed out waiting for the condition on openstackdataplanes/openstack-edpm`

We want to test that patch in ci-framework before landing it - that's done here mostly: https://github.com/openstack-k8s-operators/ci-framework/pull/361

cjeanner:
/unhold

That patch is working as-is - we're facing some "next issue" related to DNS, but it has nothing to do with this new file. So we're good to go.

@rebtoor @rabi if any of you could merge this one, even if zuul is crying (there's no reason it cries on that file addition - issues are the usual ones we want to correct with the new job layout), that would be really nice :).

rebtoor (approved):

PR #221, opened 2023-07-21
Exclude edpm_ddp_package molecule test from being executed locally

rebtoor:
It's almost useless to execute this test:

* it needs linux-firmware package installed (we can install it in a container but still...)
* it needs to load that module (we can mock that, but, again...)

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/398


rebtoor:
recheck

rebtoor:
recheck

jpodivin (approved):
This does make sense. I we absolutely have to test it locally we still can invoke it by hand.

PR #225, opened 2023-07-25
Use edpm_network_config_override when defined

bshephar:
Given the current method of interacting with the network_config role, this change moves to using edpm_network_config_override when it's not an empty string. This will allow users to provide jinja2 data instead of json data. If this variable isn't defined, we will fall back to using the default template.

### DIFF
--- a/roles/edpm_network_config/defaults/main.yml
+++ b/roles/edpm_network_config/defaults/main.yml
@@ -39,6 +39,5 @@ edpm_network_config_nmstate: false
 edpm_network_config_os_net_config_mappings: {}
 edpm_network_config_safe_defaults: true
 edpm_network_config_with_ansible: true
-edpm_network_config_template: templates/single_nic_vlans/single_nic_vlans.j2
-edpm_network_config_override: {}
+edpm_network_config_template: ""
 edpm_bond_interface_ovs_options: "bond_mode=active-backup"
--- a/roles/edpm_network_config/tasks/os_net_config.yml
+++ b/roles/edpm_network_config/tasks/os_net_config.yml
@@ -26,21 +26,11 @@
     - name: Render overidden network config
       no_log: "{{ edpm_network_config_hide_sensitive_logs | bool }}"
       ansible.builtin.copy:
-        content: "{{ edpm_network_config_override | to_yaml }}"
+        content: "{{ edpm_network_config_template | from_yaml | to_yaml }}"
         dest: "{{ nic_config_file }}"
         mode: '0644'
         backup: true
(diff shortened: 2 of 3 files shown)


roles/edpm_network_config/tasks/os_net_config.yml
```diff
@@ -26,12 +26,13 @@
     - name: Render overidden network config
       no_log: "{{ edpm_network_config_hide_sensitive_logs | bool }}"
       ansible.builtin.copy:
-        content: "{{ edpm_network_config_override | to_yaml }}"
```
rabi, line 29: Isn't `content` is expected a string and `to_yaml` return one?  I  guess it also validates that.

bshephar: I know we discussed this on Slack. But for continuity here. It ended up printing the contents of the file in a single line with all of the `\n`'s in the string. 

As you suggested, `| from_yaml | to_yaml` does indeed fix that.

roles/compute-inv.yaml
rebtoor, line 1: this file shouldn't be here

roles/net_config.yml
rebtoor, line 1: also this file, it seems like a test :sweat_smile:

rebtoor (changes requested):

roles/compute-inv.yaml
bshephar: haha yeah. My bad.

rabi:
This PR should have failed in CI without adding the inline network config to dataplane CR. But looks like we don't test changes in CI again (as we don't use built ansibleee image) :/

rebtoor (approved):

roles/edpm_network_config/defaults/main.yml
```diff
@@ -39,6 +39,5 @@ edpm_network_config_nmstate: false
 edpm_network_config_os_net_config_mappings: {}
 edpm_network_config_safe_defaults: true
 edpm_network_config_with_ansible: true
-edpm_network_config_template: templates/single_nic_vlans/single_nic_vlans.j2
-edpm_network_config_override: {}
+edpm_network_config_template: ""
```
rabi, line 42: `edpm_network_config` would probably be better to tell that it;s not a file.

roles/edpm_network_config/tasks/os_net_config.yml
```diff
@@ -26,21 +26,11 @@
     - name: Render overidden network config
```
rabi, line 26: This can be changed too

PR #226, opened 2023-07-25
Adding note about edpm_ssh_known_hosts not exposing parameters

jpodivin:


rebtoor (approved):

rebtoor:
I think I should add "roles/*/meta/*" to the list of `irrelevant-files` that triggers zuul...

jpodivin:
> I think I should add `roles/*/meta/*` to the list of `irrelevant-files` that triggers zuul...

That would save us time. But then again the argspec file is used when roles are invoked in playbooks to verify args.
If we modify it and don't test it appropriately we may run into unforeseen consequences.
It isn't likely, but I can see a scenario where a change in use of the role, coinciding with change to an argspec, could cause new bugs to appear.

PR #233, opened 2023-07-25
Argspec for edpm_iscsid role

jpodivin:
Patch also removes unused `edpm_iscsid_hide_sensitive_logs` variable.

The PR also removes MD5 from the list of approved hash algorithms.

### DIFF
--- a/roles/edpm_iscsid/defaults/main.yml
+++ b/roles/edpm_iscsid/defaults/main.yml
@@ -32,4 +32,4 @@ edpm_iscsid_volumes:
   - /etc/target:/etc/target:z
   - /var/lib/iscsi:/var/lib/iscsi:z
 
-edpm_iscsid_chap_algs: 'SHA3-256,SHA256,SHA1,MD5'
+edpm_iscsid_chap_algs: 'SHA3-256,SHA256,SHA1'
--- a/roles/edpm_iscsid/defaults/main.yml
+++ b/roles/edpm_iscsid/defaults/main.yml
@@ -20,8 +20,6 @@
 # All variables within this role should have a prefix of "edpm_iscsid"
 edpm_iscsid_debug: "{{ (ansible_verbosity | int) >= 2 | bool }}"
 
-edpm_iscsid_hide_sensitive_logs: true
-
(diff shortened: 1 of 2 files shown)


slagle (approved):

roles/edpm_iscsid/defaults/main.yml
```diff
@@ -34,4 +32,4 @@ edpm_iscsid_volumes:
   - /etc/target:/etc/target:z
   - /var/lib/iscsi:/var/lib/iscsi:z
 
-edpm_iscsid_chap_algs: 'SHA3-256,SHA256,SHA1,MD5'
+edpm_iscsid_chap_algs: 'SHA3-256,SHA256,SHA1'
```
rebtoor, line 35: :clap:

roles/edpm_iscsid/meta/argument_specs.yml
```diff
@@ -3,4 +3,34 @@ argument_specs:
...
+          Produce additional text messages describing role operation.
+          Set to false by default using following template:
+          {{ (ansible_verbosity | int) >= 2 | bool }}
+      edpm_iscsid_image:
+        type: str
+        default: "quay.io/podified-antelope-centos9/openstack-iscsid:current-podified"
+        description: URL of the iscsid container image.
+      edpm_iscsid_config_dir:
+        type: path
+        default: /var/lib/config-data/ansible-generated/iscsid
```
rebtoor, line 20: Description is missing, but this var isn't used at all. I'd remove it.

Directory is created [here](https://github.com/openstack-k8s-operators/edpm-ansible/blob/670640e21f4840408a7f3be26008733556e0fce6/roles/edpm_iscsid/tasks/install.yml#L19-L32), without referring to this directory.

rebtoor (changes requested):

rebtoor (approved):

roles/edpm_iscsid/defaults/main.yml
ASBishop: Why was this change made? It needs to be coordinated with similar changes in the controlplane, and should have been cross checked with someone in the storage team.

PR #243, opened 2023-07-26
Replace edpm_network_config_template with existing _override

bshephar:
Given the way this interface is leveraged via the dataplane-operator, it
doesn't make sense to define a file path. Instead, this change moves the
existing edpm_network_config_override to be the primary interface. It
has subsequently been renamed to edpm_network_config_template.
This allows users to define jinja2 strings for the variable, and they are
rendered out to compute nodes using the copy module.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/451

### DIFF
--- a/roles/edpm_network_config/defaults/main.yml
+++ b/roles/edpm_network_config/defaults/main.yml
@@ -38,6 +38,5 @@ edpm_network_config_manage_service: true
 edpm_network_config_nmstate: false
 edpm_network_config_os_net_config_mappings: {}
 edpm_network_config_safe_defaults: true
-edpm_network_config_template: templates/single_nic_vlans/single_nic_vlans.j2
-edpm_network_config_override: {}
+edpm_network_config_template: ""
 edpm_bond_interface_ovs_options: "bond_mode=active-backup"
--- a/roles/edpm_network_config/molecule/default/converge.yml
+++ b/roles/edpm_network_config/molecule/default/converge.yml
@@ -18,7 +18,39 @@
 - name: Converge
   hosts: all
   vars:
-    edpm_network_config_template: templates/standalone.j2
+    edpm_network_config_template: |
+            ---
+            {% set control_virtual_ip = net_vip_map.ctlplane %}
+            {% set public_virtual_ip = vip_port_map.external.ip_address %}
+            {% if ':' in control_virtual_ip %}
+            {%   set control_virtual_cidr = 128 %}
+            {% else %}
+            {%   set control_virtual_cidr = 32 %}
+            {%   endif %}
+            {% if ':' in public_virtual_ip %}
+            {%   set public_virtual_cidr = 128 %}
+            {% else %}
+            {%   set public_virtual_cidr = 32 %}
+            {%   endif %}
+            network_config:
+            - type: ovs_bridge
+              name: br-ctlplane
+              use_dhcp: false
+              mtu: {{ ctlplane_mtu }}
+              ovs_extra:
+              - br-set-external-id br-ctlplane bridge-id br-ctlplane
+              addresses:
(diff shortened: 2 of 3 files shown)


bshephar:
We need to fix CI to adjust for this change in the dataplane-operator CI pipeline.

slagle (commented):
I understand where this is going, and I think I can get behind it. It does however have the implication that you can't use our packaged templates in the role templates dir without reproducing the content into the DataPlane resource. Hard to say how many users have actually done that historically, which is why I say I can get behind this, but I do think it's worth raising for broader awareness. May want input from the hardprov folks.

bshephar:
> I understand where this is going, and I think I can get behind it. It does however have the implication that you can't use our packaged templates in the role templates dir without reproducing the content into the DataPlane resource. Hard to say how many users have actually done that historically, which is why I say I can get behind this, but I do think it's worth raising for broader awareness. May want input from the hardprov folks.

Yeah, good point. I'll loop them in for awareness.

I can't think of any situation where I have seen a user using our default network configuration files. But it would definitely be nice to get some more eyes on it to make sure we're not taking functionality away from anyone.

slagle:
What's holding this one up at this point? Do we just need the CI change? We might need both vars for a time to land this without breaking CI. Then we can come back around and delete the _override var once CI is updated. So the steps would be:

1. Update CI to use _override
2. Land this PR, with the 2 vars doing the same thing. Just set _template=_override
3. Switch CI to use _template
4. Remove _override

bshephar:
recheck

slagle (approved):

bshephar:
recheck

bshephar:
Hmm, this keeps failing before it launches the network job. I've seen it run once, but it has failed a couple of times consecutively with the dataplane-operator reporting:
```
<pre class="zuul-log-output">

2023-08-08T05:45:36.316658281+00:00 stderr F E0808 05:45:36.316399       1 leaderelection.go:330] error retrieving resource lock openstack-operators/e12e763d.openstack.org: Get "https://10.217.4.1:443/apis/coordination.k8s.io/v1/namespaces/openstack-operators/leases/e12e763d.openstack.org": context deadline exceeded
--


</pre><pre class="zuul-log-output"><span class="log-message zuul-log-sev-0">2023-08-08T05:45:41.983460172+00:00 stderr F I0808 05:45:41.979933       1 leaderelection.go:283] failed to renew lease openstack-operators/e12e763d.openstack.org: timed out waiting for the condition</span></pre>
```

bshephar:
recheck

bshephar:
recheck

slagle:
recheck

bshephar:
```
No resources found in openshift-nmstate namespace.
--
No resources found in openshift-nmstate namespace.
No resources found in openshift-nmstate namespace.
No resources found in openshift-nmstate namespace.
error: You must be logged in to the server (Unauthorized)


```

bshephar:
recheck

bshephar:
recheck

bshephar:
recheck

bshephar:
The last run passed the os-net-config execution, but the overall CI job failed. Re-running to see if that error is reproduced now since the changes in that CI job.

bshephar:
I'm not sure why we're using Ubuntu 16.04 for anything, but for whatever reason the CI env failed to pull that container. I can see that it is available though, so rechecking

bshephar:
recheck

bshephar:
recheck

slagle:
+2

But we need to wait for the Depends-On and to see it work with https://github.com/openstack-k8s-operators/dataplane-operator/pull/394

bshephar:
Broke this up into more backwards compatible PR's. The last of which is this one:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/367

PR #246, opened 2023-07-27
Argspecs for edpm_nova_compute role

jpodivin:


SeanMooney (commented):
given this role is due to be deleted soon I'm not sure if
we should proceed with this pr.

the same as true of the edpm_nova_libvirt role

roles/edpm_nova_compute/defaults/main.yml
```diff
@@ -18,6 +18,7 @@
 # All variables intended for modification should be placed in this file.
 
 # All variables within this role should have a prefix of "edpm_nova_compute"
+edpm_service_net_map: "{{ service_net_map }}"
```
SeanMooney, line 21: this is more a playbook var then a role var but i understand why you are adding this here.
this is not actully part of the edpm_nova_compute role which is why it does nto have edpm_nova_compute
as the prfix so this is correct.

roles/edpm_nova_compute/defaults/main.yml
```diff
@@ -186,7 +187,7 @@ edpm_nova_compute_libvirt_enabled_perf_events: ''
 edpm_nova_compute_libvirt_inject_key: false
 edpm_nova_compute_libvirt_inject_partition: '-2'
 edpm_nova_compute_libvirt_inject_password: false
-edpm_nova_compute_libvirt_live_migration_inbound_addr: '{{ hostvars[inventory_hostname]["fqdn_" ~ service_net_map["nova_libvirt_network"]] }}'
+edpm_nova_compute_libvirt_live_migration_inbound_addr: '{{ hostvars[inventory_hostname]["fqdn_" ~ edpm_service_net_map["nova_libvirt_network"]] }}'
```
SeanMooney, line 190: if we were keeping this role longterm this is somehting we woudl want to review.
in general we expect this to be an ip not an fqdn but both will work.
i assume the fqdn was used because of TLSe so this is not nessisarly wrong just not what we would normally expect to be popluatled

https://docs.openstack.org/nova/latest/configuration/config.html#libvirt.live_migration_inbound_addr

if you notice the upstream docs detail that this should be an ip addres

IP address used as the live migration address for this host.
"""
This option indicates the IP address which should be used as the target for live migration traffic when migrating to this hypervisor. This metadata is then used by the source of the live migration traffic to construct a migration URI.

If this option is set to None, the hostname of the migration target compute node will be used."""

im not entirly sure if '{{ hostvars[inventory_hostname]["fqdn_" ~ edpm_service_net_map["nova_libvirt_network"]] }}'
is returning an ip or an fqdn but if its the former and it works its technially relying on undocumented behavior

This is prexisting behavior and not related to this PR however just wanted to note that while reviewing.

roles/edpm_nova_compute/defaults/main.yml
```diff
@@ -214,8 +215,8 @@ edpm_nova_compute_vendordata_dynamic_auth_user_domain_name: Default
 edpm_nova_compute_vif_plug_ovs_ovsdb_connection: ''
 edpm_nova_compute_vnc_enabled: true
 edpm_nova_compute_vncproxy_host: ''   # EndpointMap, NovaPublic, host_nobrackets
-edpm_nova_compute_vnc_server_listen: '{{ hostvars[inventory_hostname][service_net_map["nova_libvirt_network"] ~ "_ip"] }}'
-edpm_nova_compute_vncserver_proxyclient_address: '{{ hostvars[inventory_hostname][service_net_map["nova_libvirt_network"] ~ "_ip"] }}'
+edpm_nova_compute_vnc_server_listen: '{{ hostvars[inventory_hostname][edpm_service_net_map["nova_libvirt_network"] ~ "_ip"] }}'
+edpm_nova_compute_vncserver_proxyclient_address: '{{ hostvars[inventory_hostname][edpm_service_net_map["nova_libvirt_network"] ~ "_ip"] }}'
```
SeanMooney, line 219: i would not expect either of thise to use the libvirt network by defaut by the way.

they should actully be using the internal_api network.

the libvirt network is not nessisarly expected to be reacable by the novnc poxy and in fact for security reasons it reallly should not be.
when a seperate network is used for libvirt it should be isolated form the contolers and only accesable on the compute nodes.

again this is preexisitng behvior so not relvent to this review just noting that this is not what we woudl expect form a nova/compute perspective.

roles/edpm_nova_compute/meta/argument_specs.yml
```diff
@@ -3,4 +3,924 @@ argument_specs:
   # ./roles/edpm_nova_compute/tasks/main.yml entry point
   main:
     short_description: The main entry point for the edpm_nova_compute role.
-    options: {}
+    options:
+      service_net_map:
```
SeanMooney, line 7: edpm_service_net_map is the paramter used by the role and service_net_map is the default that it is initalised form

roles/edpm_nova_compute/molecule/default/converge.yml
```diff
@@ -17,5 +17,29 @@
...
+        command: 'sleep 3600'
+  pre_tasks:
+    - ansible.builtin.set_fact:
+        internal_api:
+          nova_api_network: mock_network
+          nova_libvirt_network: mock_network
+        internal_api_ip: 10.0.0.1
+        fqdn_internal_api: mock_api
   roles:
+    - role: "osp.edpm.edpm_container_standalone"
```
SeanMooney, line 41: ok so this role is setting some default in teh env.

i would generaly consider that to be a bug in the interaction between these two roles but i guess that is ok for now.

SeanMooney (approved):
from the point of view of the args spec changes this pr looks ok to me.
this role likely wont remain in the repo for long but i dont see any harm in merging this for now to keep it inlien with the other roles until it is eventrually removed.

SeanMooney (approved):

SeanMooney (approved):
from the point of view of the args spec changes this pr looks ok to me.
this role likely wont remain in the repo for long but i dont see any harm in merging this for now to keep it inlien with the other roles until it is eventrually removed.

slagle (commented):
Given https://github.com/openstack-k8s-operators/edpm-ansible/pull/217 this role isn't even used anymore right?

SeanMooney:
> Given #217 this role isn't even used anymore right?

it wasn't used even before that since we started using the dataplane operator to deploy the nova via the novaexternalcompute CRD 

i have an old PR to delete it which ill revie after i port a couple of task form this but my plan is to remove this rule by the end of Q3 once https://github.com/openstack-k8s-operators/edpm-ansible/pull/160 and a few other bits are ported to other roles.

that partly why I'm ok with merging this and the edpm_nova_libvirt arg spec

they should not break anything. it seams ok on review and they should not currently used anywhere.

PR #247, opened 2023-07-27
Argspecs for edpm_nova_libvirt role

jpodivin:
This is a tough one. Defaults contain lot of TODOs and I'm not exactly sure if/how were they resolved.
I'm also seeing lot of things that may no long er be necessary.


SeanMooney (approved):
as noted on the edpm_nova_compute review this isn't used anywhere
but this does not break anything so I'm ok with merging this to be consistent.

this role will be removed in the near to medium term but I'm fine with keeping it update to date with the new role testing guidelines i.e argspecs since this is already written but i would suggest ignoring this for future refactors.

PR #248, opened 2023-07-28
Removing edpm_hosts_entries role

jpodivin:
The role isn't used anymore, so we shouldn't keep it around.


jpodivin:
recheck

jpodivin:
recheck

rebtoor:
recheck

rabi:
recheck

rebtoor (approved):

karelyatin (commented):
https://github.com/openstack-k8s-operators/edpm-ansible/blob/fd8c60e31d9e0a29e740e4f5770d9d23f5f03074/roles/edpm_nodes_validation/tasks/main.yml#L66C9-L80 part also looks irrelevant now if this role is being dropped and no option to use /etc/hosts.

slagle:
> /hold lets hold for now we have an issue with expose dns service on a 3 node cluster without dedicated workers. a workaround would be to use /etc/hosts.

@stuggi Is this still considered on hold?

abays:
> > /hold lets hold for now we have an issue with expose dns service on a 3 node cluster without dedicated workers. a workaround would be to use /etc/hosts.
> 
> @stuggi Is this still considered on hold?

@slagle I believe it turned out that DNS worked with a 3 node master/worker cluster after all.

stuggi:
> > > /hold lets hold for now we have an issue with expose dns service on a 3 node cluster without dedicated workers. a workaround would be to use /etc/hosts.
> > 
> > 
> > @stuggi Is this still considered on hold?
> 
> @slagle I believe it turned out that DNS worked with a 3 node master/worker cluster after all.

its no longer required to hold it. there was a missconfig which resulted in the dns service to not be reachable.

slagle (approved):

PR #249, opened 2023-07-28
Add templating support for nmstate defined networking configuration

rebtoor:
* Fixed `nmstate` molecule scenario for `edpm_network_config` role

### DIFF
--- a/roles/edpm_network_config/tasks/main.yml
+++ b/roles/edpm_network_config/tasks/main.yml
@@ -38,6 +38,9 @@
         name: NetworkManager
         state: restarted
       when: nm_ovs_status.changed  # noqa: no-handler
+    - name: Render network_state variable
+      ansible.builtin.set_fact:
+        network_state: "{{ edpm_network_config_template | from_yaml }}"
     - name: Load system-roles.network tasks [nmstate]
       ansible.builtin.include_role:
         name: "{{ systemrolename }}"
--- a/roles/edpm_network_config/molecule/default/prepare.yml
+++ b/roles/edpm_network_config/molecule/default/prepare.yml
@@ -15,6 +15,7 @@
 # under the License.
 - name: Prepare test_deps
   hosts: all
+  become: true
   roles:
     - role: ../../../../molecule/common/test_deps
       test_deps_setup_edpm: true
@@ -27,12 +28,13 @@
 
   tasks:
     - name: Ensure legacy scripts installed
-      become: true
       ansible.builtin.package:
         name: network-scripts
         state: present
       when:
         - ansible_facts['distribution_major_version'] is version('8', '==')
     - name: Create a dummy network interface
-      become: true
       ansible.builtin.command: "ip link add dummy0 type dummy"
+      register: ip_command_output
+      changed_when: ip_command_output.rc == 0
+      failed_when: ip_command_output.rc != 0
--- a/roles/edpm_network_config/molecule/nmstate/converge.yml
+++ b/roles/edpm_network_config/molecule/nmstate/converge.yml
@@ -18,14 +18,19 @@
 - name: Converge
   hosts: all
   vars:
-    edpm_network_config_tool: nmstate
-    network_provider: nm
-    network_state:
+    # edpm_network_config - nmstate
+    edpm_network_config_tool: 'nmstate'
+    edpm_network_config_template: |
+      ---
       interfaces:
-      - name: dummy0
-        type: dummy
-        state: down
-        ipv4:
-          enabled: false
+        - name: dummy0
+          type: dummy
+          state: up
+          ipv4:
+            address:
+            - ip: 192.168.180.2
+              prefix-length: 24
+            dhcp: false
+            enabled: true
   roles:
-    - role: "osp.edpm.edpm_network_config"
+    - role: osp.edpm.edpm_network_config


rebtoor:
it's useless to "recheck" the CI, job is already passed on baremetal.

roles/edpm_network_config/tasks/main.yml
```diff
@@ -36,6 +36,9 @@
         name: NetworkManager
         state: restarted
       when: nm_ovs_status.changed
+    - name: Render network_config variable
+      ansible.builtin.set_fact:
+        network_config: "{{ lookup('template', edpm_network_config_nmstate_template ) }}"
```
rabi, line 41: Where is this var used? Isn't `network_state` the var used by linux-system-reoles[1]? 

[1] https://github.com/linux-system-roles/network#variables

roles/edpm_network_config/templates/system_role_nmstate/single_nic_vlans.j2
```diff
@@ -0,0 +1,58 @@
+---
```
rabi, line 1: Do we still create these templates keeping in mind the move to use inline templates that @bshephar is proposing in the other PR?

jpodivin: Keeping templates as a separate file has some advantages. For example when it comes to questions of file history. If you want to know all the changes made to template, you just need to call git. But if the template is in line you get it mixed with everything else.

roles/edpm_network_config/tasks/main.yml
rebtoor: indeed, i thought `network_state` and wrote `network_config`. :shrug: 

I'm scared that the tests passed anyway.

roles/edpm_network_config/templates/system_role_nmstate/single_nic_vlans.j2
rebtoor: @rabi i agree with you but at the moment we're still with files. When we'll remove other templates, I'll replace this one too. :)

rebtoor:
recheck

roles/edpm_network_config/tasks/main.yml
```diff
@@ -36,6 +36,9 @@
         name: NetworkManager
         state: restarted
       when: nm_ovs_status.changed
+    - name: Render network_config variable
+      ansible.builtin.set_fact:
+        network_state: "{{ lookup('template', edpm_network_config_nmstate_template ) }}"
```
rabi, line 41: Should we use the fqcn `ansible.builtin.template` here?

rebtoor: You're right. I took the example from the webpage, but in the docs ( `ansible-doc -t lookup template` ) the fqcn is used.

Thanks for spotting @rabi !

roles/edpm_network_config/molecule/nmstate/dummy_config.j2
```diff
@@ -0,0 +1,7 @@
+---
+interfaces:
+- name: dummy0
+type: dummy
```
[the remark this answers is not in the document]
rebtoor: ah! I made this mistake with an extra set of ctrl+shift+tab when i removed a level of indendation due to the removal of `network_state:`

Thanks for spotting it @jpodivin !

roles/edpm_network_config/defaults/main.yml
```diff
@@ -40,5 +40,6 @@ edpm_network_config_os_net_config_mappings: {}
 edpm_network_config_safe_defaults: true
 edpm_network_config_with_ansible: true
 edpm_network_config_template: templates/single_nic_vlans/single_nic_vlans.j2
+edpm_network_config_nmstate_template: templates/system_role_nmstate/single_nic_vlans.j2
```
bshephar, line 43: Rather than making this a file path, which is only really useful for CI can we do a similar thing to this PR?
https://github.com/openstack-k8s-operators/edpm-ansible/pull/243

So instead of having a template file, we can in-line the yaml which will work better for the dataplane-operator CRD interface?

Like this:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/243/files#diff-89da6e02d8fbed04de229b13205144afd3d4bcedda05167972c6badac58102b1R21

Which means the Dataplane operator CRD interface would look like:
https://github.com/openstack-k8s-operators/dataplane-operator/pull/318/files#diff-7c43891912d609cdc0e86a9cb67c44e43264d6e6463d9afa2726a1292cf749caR57-R90

Since users can't realistically use a filepath template file.

bshephar (changes requested):

roles/edpm_network_config/defaults/main.yml
bshephar: I should have read the conversation history before my comment. I see this was discussed above too.

rebtoor:
/hold

we need to understand if we want to go on with the path described in: https://github.com/openstack-k8s-operators/edpm-ansible/pull/243/

roles/edpm_network_config/defaults/main.yml
rebtoor: I modified the role following your PRs examples. I also drop `edpm_network_config_nmstate_template` (re)using `edpm_network_config_template` var because otherwise we should handle this case here: https://github.com/openstack-k8s-operators/dataplane-operator/pull/284/files#diff-874aae70e738f15dc1bf30750cee0e87e3278c5c02d37e35c0943ca6c29393bb

slagle (approved):

PR #253, opened 2023-08-02
Replace package module with dnf module

rebtoor:
It's pointless to continue to use the `package` module (which itself rely on an action plugin to be executed) since edpm-ansible only supports EL 8/9 targets.


fao89 (approved):

rebtoor:
recheck

slagle (approved):

rebtoor:
recheck

jpodivin (approved):

rebtoor:
recheck

PR #256, opened 2023-08-03
Adding configure_frr playbook

jpodivin:


jpodivin:
recheck

rebtoor (changes requested):
We have used `ansible.builtin.import_roles` under `tasks` in all the other playbooks, but here we're using the `roles` directive. We should adapt this playbook to the style of the others IMHO.

slagle:
> We have used `ansible.builtin.import_role` under `tasks` in all the other playbooks, but here we're using the `roles` directive. We should adapt this playbook to the style of the others IMHO.

:+1:

slagle:
agreed. why change from what are already using in dataplane-operator?

slagle (approved):

PR #260, opened 2023-08-03
Adding install_frr playbook

jpodivin:


rebtoor (changes requested):

rebtoor (approved):

slagle (approved):

jpodivin:
recheck

PR #261, opened 2023-08-03
Adding install_os playbook

jpodivin:


rebtoor (changes requested):
`ansible.builtin.import_role`

rebtoor (approved):

rebtoor:
@jpodivin be aware of this: https://github.com/openstack-k8s-operators/dataplane-operator/pull/335

cc @slagle

slagle:
> @jpodivin be aware of this: [openstack-k8s-operators/dataplane-operator#335](https://github.com/openstack-k8s-operators/dataplane-operator/pull/335)
> 
> cc @slagle

Indeed, this will need to be updated to account for that change.

slagle (approved):

ASBishop:
Is the intention that the dataplane service will be updated to invoke this new playbook instead of executing the list of tasks as it currently does?

The reason I ask is that I need to update the install/configure/run os tasks (for multipathd) and I need to know where to make the change.

PR #262, opened 2023-08-03
Adding install_ovn_bgp_agent playbook

jpodivin:


playbooks/install_ovn_bgp_agent.yaml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Deploy EDPM OVN BGP Agent Install
+  hosts: all
+  strategy: linear
+  become: True
+  tasks:
+    - name: Install edpm_ovn_bgp_agent
+      import_role:
```
rebtoor, line 8: `ansible.builtin.import_role`

rebtoor (changes requested):

rebtoor (approved):

slagle (approved):

PR #270, opened 2023-08-04
mova edpm-ansible-crc-podified-edpm-baremetal to github-manual

SeanMooney:
This change moves the edpm-ansible-crc-podified-edpm-baremetal
job to the manual pipeline until it can be fixed.

### DIFF
--- a/zuul.d/projects.yaml
+++ b/zuul.d/projects.yaml
@@ -12,6 +12,11 @@
         - edpm-ansible-crc-podified-edpm-deployment:
             dependencies:
               - edpm-ansible-content-provider
+    github-manual:
+      jobs:
+        # Note: the baremetal job is very unstable as such
+        # it is moved to the manual pipeline until that can
+        # be fixed. comment "check-github" to run this job.
         - edpm-ansible-crc-podified-edpm-baremetal:
             dependencies:
               - edpm-ansible-content-provider


raukadah:
@SeanMooney Thank you for taking a look at the job. Based on the health of[ edpm baremetal job ](https://review.rdoproject.org/zuul/builds?job_name=edpm-ansible-crc-podified-edpm-baremetal&skip=0), there are few success. Can we move this job to non-voting instead of github-manual? By making it non-voting, it will give feedback to developer. But moving it to github-manual, it will be completely ignored.

SeanMooney:
My experience is most developers ignore non voting job unless they are working on the thing it's testing and most people on the edpm ansible repo are not working on ironic. That is why I didn't make it non voting. I think it would be better to keep it in manual until the job is stabilised to save for time.

raukadah (approved):
Thank you @SeanMooney , make sense. Added an issue in jira backlog to move the job back to voting once crc standalone work finishes.

raukadah:
recheck

rebtoor (approved):

rabi (changes requested):
The issue should be fixed after https://github.com/openstack-k8s-operators/openstack-operator/pull/427

It's red as a result of us force merging PRs that break the job without understanding the failures.

rebtoor:
I've closed this PR because the job has already been removed, and it will be re-added after proper testing.

PR #272, opened 2023-08-07
Added `edpm_download_cache` meta role and refactored `edpm_bootstrap` role to split packages download from their installation

rebtoor:
We are trying to isolate download of the packages from their installation in order to improve the entire deployment process.

Also

- added callbacks to molecule configuration in order to obtain timing
- fixed `mode` definition in tasks as for module indication
- moved `become` directive at the end of each tasks where it's used
- added `roles/edpm_bootstrap/tasks/download_cache.yml` file to download all packages required by the role
- use `edpm_bootstrap_dnf_use_cacheonly` to control if packages should be installed from cache (by `download_cache.yml` tasks file`) or not.
- added `edpm_download_cache` meta role
- manage failures for packages download/installation with retries


roles/edpm_bootstrap/tasks/bootstrap.yml
```diff
@@ -14,33 +14,32 @@
...
   register: cloud_init_enabled
+  become: true
 
-- debug:
+- name: Print cloud-init service status
+  ansible.builtin.debug:
     var: cloud_init_enabled
 
 - name: Check if cloud-init is disabled via kernel args
   ansible.builtin.shell: cat /proc/cmdline | grep -q cloud-init=disabled
```
jpodivin, line 38: I don't believe this task needs root. The `/proc/cmdline` is 444 by default.
You may also replace the shell with https://docs.ansible.com/ansible/latest/collections/ansible/builtin/lineinfile_module.html in check mode.

rebtoor:
recheck

slagle:
recheck

roles/edpm_bootstrap/tasks/packages.yml
```diff
@@ -21,46 +21,52 @@
...
-  failed_when: false
-
-- name: Deploy required packages to bootstrap EDPM
+- name: Install release version package
+  ansible.builtin.dnf:
+    name: "{{ edpm_bootstrap_release_version_package }}"
+    state: present
+    cacheonly: "{{ edpm_bootstrap_dnf_use_cacheonly | default(true) | bool }}"
+  when:
+    - (ansible_facts['distribution'] | lower) == 'redhat'
```
slagle, line 32: :+1: The package isn't available on CentOS, so good to check for redhat here.

roles/edpm_download_packages/meta/main.yml
```diff
@@ -0,0 +1,43 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+galaxy_info:
+  author: OpenStack
+  description: EDPM OpenStack Role -- edpm_download_packages
```
slagle, line 20: If the tasks files are going to be called download_cache.yml, then it might make sense to call this role edpm_download_cache.

jpodivin (changes requested):
Please make check that this doesn't inadvertly affect use of this role in other places. Specifically in the `configure_network.yaml` playbook. 

I think it may need inclusion of the new role as well. 

https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/playbooks/configure_network.yaml#L11

roles/edpm_download_packages/meta/main.yml
rebtoor: yeah, i guess you're right...

rebtoor:
> Please make check that this doesn't inadvertly affect use of this role in other places. Specifically in the `configure_network.yaml` playbook.
> 
> I think it may need inclusion of the new role as well.
> 
> https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/playbooks/configure_network.yaml#L11

I don't want to call the new role (`edpm_download_cache`, former `edpm_download_packages`) from other playbooks, i want that is called by its playbook when needed, specifically at the start of a fresh install, an update *or* if a user want to. If we include the role in each needed playbooks, we lose its purpose (since it will perform a large number of dnf tasks... now it include only bootstrap tasks, but it will be extended to other roles)

@slagle WDYT?

rebtoor:
recheck

fao89:
recheck

fao89:
oh! I thought it was because it was depending on the openstack-operator

roles/edpm_bootstrap/tasks/download_cache.yml
```diff
@@ -0,0 +1,24 @@
+---
+
+- name: Download needed packages
+  ansible.builtin.dnf:
```
fao89, line 4: should we add retries?

rebtoor: Mmmh, given that it could have sense and I'm not against it...we didn't have similar behavior for packages installation tasks, why we should need it for download? 

Do you want to prevent the playbook failure for temporary network issues or something like that?

fao89: Exactly! During the adoption hackathon I've seen people having network issues, and currently, when a job fails, it takes 10 minutes to retry: https://github.com/openstack-k8s-operators/edpm-ansible/pull/314

rebtoor: yeah,i like the idea, we could implement it both for download and install tasks (if download_cache service isn't triggered, downloads are performed during installation tasks to avoid errors).

rebtoor: @fao89 take a look now :)

roles/edpm_bootstrap/defaults/main.yml
```diff
@@ -17,6 +17,13 @@
 
 # All variables intended for modification should be placed in this file.
 
+
+# seconds between retries for download tasks
+edpm_bootstrap_download_delay: 5
+
+# number of retries for download tasks
+edpm_bootstrap_download_retries: 5
```
fao89, line 25: 👍🏽

PR #276, opened 2023-08-11
Revert "Removing MD5 from the list of approved algos"

ASBishop:
This reverts commit 6b75cd6c89e9661c039d141cd970eee1f4304d70.

It's true that MD5 is not compatible with FIPS, but MD5 is required in non-FIPS deployments with older storage hardware. Allowing MD5 by default is consistent with tripleo based deployments, and will give storage providers time to certify their product works in FIPS environments where MD5 is prohibited. In addition, removing MD5 by default will need to be coordinated with corresponding changes in the controlplane.


slagle (approved):

jpodivin:
recheck

fao89:
recheck

fao89:
recheck

PR #282, opened 2023-08-16
relax nova config patteren

SeanMooney:
This change relaxes the shell glob used by kolla_set_configs
to match which files shoudl be copied to "/etc/nova/nova.conf.d".
The new pattern will match "*nova*.conf" instead of "*nova.conf".

The molecule tests are expanded to assert tht the expected files
are copied to verify this change.

### DIFF
--- a/roles/edpm_nova/molecule/default/test-data/nova-cell1/01-nova.conf
+++ b/roles/edpm_nova/molecule/default/test-data/nova-cell1/01-nova.conf
@@ -2,7 +2,3 @@
 debug = True
 transport_url = fake:/
 compute_driver = libvirt.LibvirtDriver
-max_logfile_count=5
-max_logfile_size_mb=50
-log_rotation_type=size
-log_file = /var/log/containers/nova/nova-compute.log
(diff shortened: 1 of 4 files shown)


fultonj:
+2 
/lgtm

This should fix the following problem which I ran into:

I can see the resultant file has been copied to my compute node.
```
[root@edpm-compute-0 ~]# cat /var/lib/openstack/config/nova/02-nova-libvirt-qemu.conf
[libvirt]
virt_type = qemu
[root@edpm-compute-0 ~]# 
```
However, I do not see the file inside my nova_compute container.
```
[root@edpm-compute-0 ~]# podman exec -ti nova_compute ls -l /etc/nova/nova.conf.d/
total 4
-rw-------. 1 nova nova 3804 Aug 16 11:25 01-nova.conf
[root@edpm-compute-0 ~]# 
```
This is because the snippet file name needed to end in "nova.conf" or it won't be copied from `host:/var/lib/openstack/config/nova` to `container:/etc/nova/nova.conf.d/`. 

This PR changes the pattern from `*nova.conf` to `*nova*.conf`

bogdando (approved):

slagle (approved):

fao89:
recheck

fao89:
recheck

fao89:
recheck

PR #275, opened 2023-08-09
Refactoring telemetry playbooks to be compliant with spec.

jpodivin:

PR #284, opened 2023-08-17
Refactoring collection playbooks to use static imports of roles

jpodivin:
Telemetry playbooks are being handled by their own PR[0], owing to their specific lineage. 

[0] #275

### DIFF
--- a/playbooks/configure_network.yaml
+++ b/playbooks/configure_network.yaml
@@ -2,10 +2,8 @@
 - name: Deploy EDPM Network
   hosts: all
   strategy: linear
-  tasks:
-    - name: Grow volumes
-      import_role:
-        name: osp.edpm.edpm_growvols
+  roles:
+    - role: osp.edpm.edpm_growvols
       tags:
         - edpm_growvols
     - name: Install edpm_bootstrap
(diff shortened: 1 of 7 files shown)


jpodivin:
recheck

jpodivin:
recheck

jpodivin:
recheck

jpodivin:
recheck

jpodivin:
recheck

jpodivin:
recheck

playbooks/configure_network.yaml
```diff
@@ -30,8 +24,6 @@
         tasks_from: kernelargs.yml
```
slagle, line 24: Did you miss converting this one?

jpodivin: Actually no. Turns out there are certain special cases when role doesn't have a proper default behavior and can only be used by explicitly importing tasks from a file. Unfortunately, the role API doesn't support that, so we have to use `import_role` module.

More verbose, but still a same result.

slagle: If the "roles" section doesn't support "tasks_from", I'm not in favor of using it. Switching to use it just for the sake of it, and we end up doing things two different ways. I don't feel that's needed.

I don't understand any other point of the commit, given it's just "to use static..." as import_role is also static. https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#importing-roles-static-reuse

So it seems like the point of this change is merely for style. And given this style doesn't support all the cases we need, my take is not to use it and just stick to import_role.

slagle (changes requested):

playbooks/configure_network.yaml
jpodivin: I understand the sentiment. There is some difference in order of operations but other than that I have to agree that this change isn't worth pursuing.

PR #288, opened 2023-08-21
Apply settings from ovsdb-config file (mounted configmap) via ovs-vsctl

booxter:


jpodivin (approved):
Someone with in depth knowledge of ovs should check as well.

roles/edpm_ovn/tasks/configure.yml
```diff
@@ -14,6 +14,33 @@
...
+  - name: read configmap file contents
+    ansible.builtin.slurp:
+      src: "{{ item.path }}"
+    loop: "{{ edpm_ovn_configmaps.files }}"
+    register: ovn_configmap_contents
+    delegate_to: localhost
+    when: edpm_ovn_configmaps.matched > 0
+
+  - name: apply configuration to Open_vSwitch:external_ids
+    ansible.builtin.shell: |
```
karelyatin, line 38: shouldn't this be done after line 70 else defaults would override the settings from config map?

roles/edpm_ovn/tasks/configure.yml
```diff
@@ -14,6 +14,33 @@
...
+      {% for line in item.content | b64decode | split('\n') %}
+      ovs-vsctl set open . external-ids:{{ line }}
```
karelyatin, line 40: As we are in early phase thinking more on this wouldn't be better to write contents in dictionary format in config map and include_vars: <file> and do set these like line 70?

booxter: I don't want ovn-operator to assume ansible format and variable names. The fact the ConfigMap is consumed by EDPM-Ansible seems to me incidental and is not part of the ovn-operator definition. And most other operators use some form of "property" (INI) file for ConfigMaps, so key=value seemed fitting. (Of course, the final format is not too important because either way it's not native to OVS - since it doesn't have a concept of configuration files.)

That said, we could still make it a "property" (yaml) file that is loaded as `include_vars`. (As long as names in ConfigMaps are not prepended with `edpm` prefixes.) What does it buy us?

roles/edpm_ovn/tasks/configure.yml
booxter: I wasn't sure what's better. It may be said that it's wise to allow to override the values passed from CP layer in EDP NodeTemplate if needed. Otherwise you can't e.g. tweak timeouts for a node or the set of EDP nodes without affecting CP nodes / other EDP nodes.

karelyatin: Yeah very good point and we should figure this out which one should be higher priority if config is allowed via both options and see what should be set by OVN Controller CRs and what with dataplane CRs. Likely node specific settings(timeouts etc) should be not set with Control Plane CRs? And things like "ovn-remote" to be set with data generated by Control Plane CRs.

booxter: I actually wasn't sure if timeout proxy from CP to EDP is the right thing to do (asked about this before). The remote and tunnel type are definitely to be in sync, as for the rest - not sure. But maybe as "defaults" open for override - yes?

roles/edpm_ovn/tasks/configure.yml
karelyatin: No, just to be clear i didn't meant to use edpm variable names in ConfigMap, what i meant was to have a generated file with contents something like below in a file like ovsdb-config.yaml, load it with include_vars and then merge the map "ovs_external_ids_override" with "edpm_ovn_ovs_external_ids" and let line 70 apply it:- 
ovsdb_external_ids_override:
  ovn-remote: "<SB DB endpoint>"

roles/edpm_ovn/tasks/configure.yml
karelyatin: yes agree ovn-remote and ovn-tunnel-type to be in sync and set with OVNController CR. For other node specific bits feels like it would be good to handle from dataplane CR side as per role/node. But let's see what others say.
I  not quite understood the point re. "defaults" and override as what defaults and override you mean.

booxter: @karelyatin re defaults: we may look at "timeout" values set in OVNController as "defaults" that are passed to EDP through ConfigMap. Then the admin that defines nodeTempate may override the timeout values using one of `edpm_` variables for the OVN playbook.

Since the order of the actions in the current PR version is:
1. apply configMap,
2. apply external_ids from `edpm_*` variable,

this "override" should work.

If you don't need to override the "default" values from configMap, you don't specify them in nodeTemplate.

roles/edpm_ovn/tasks/configure.yml
booxter: @karelyatin ah so the goal would be to replace the step that parses the key=value file with (perhaps somewhat simpler) step that loads the yaml file as ansible variables + merges it into the edpm_ovn_ovs_external_ids?

I don't know if it saves much of complexity (I have little experience with ansible), but I can try. To confirm, you see the lines 39-41 as too complex?

booxter: On second thought, it's not too complex, and I don't think that we should adjust API (ConfigMap) to a tool in question (ansible), so the fact that a different format could be *marginally* simpler for ansible to load shouldn't drive the decision either way. Considering that respinning will take some of dev time and CI cycles, unless there's a good reason to go there, I prefer to leave it as-is.

roles/edpm_ovn/tasks/configure.yml
karelyatin: @booxter Thanks now got it what you meant. Actually as pointed initially with current implementation defaults from edpm will always override what's set with Configmap so current version needs to be fixed.

roles/edpm_ovn/tasks/configure.yml
karelyatin: yeap right it was not about complexity.  If edpm cores are ok with current implementation i can live with it :)

booxter: I switched to facts, so it's now yaml.

roles/edpm_ovn/tasks/configure.yml
booxter: Great catch. I fixed it. Also included molecule to test this.

booxter:
Changes:
- switch format of configmap file to yaml
- load as facts
- apply defaults to external_ids edpm variable
- added molecule verify step

roles/edpm_ovn/tasks/configure.yml
```diff
@@ -14,6 +14,18 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Apply ovsdb configuration from configMaps
+  block:
+  - name: Load ovsdb configuration from file
+    ansible.builtin.include_vars:
+      file: "{{ edpm_ovn_config_src }}/ovsdb-config"
+      name: edpm_ovn_configmap_facts
+    delegate_to: localhost
```
karelyatin, line 23: as a fail safe could also add so it works without config map
when: "'{{ edpm_ovn_config_src }}/ovsdb-config' is file"

roles/edpm_ovn/tasks/configure.yml
```diff
@@ -14,6 +14,18 @@
...
+
+  - name: Append ovsdb configMap options to external_ids
+    ansible.builtin.set_fact:
+      edpm_ovn_ovs_external_ids: "{{ edpm_ovn_ovs_external_ids | combine(edpm_ovn_configmap_facts) }}"
```
karelyatin, line 27: if above task is changed then can adjust this to combine(edpm_ovn_configmap_facts) -> combine(edpm_ovn_configmap_facts|default({}))

karelyatin (commented):
/lgtm

roles/edpm_ovn/tasks/configure.yml
booxter: Great. Ansible discourage jinja in conditionals, so I will handle it a bit differently, but overall in the direction you suggested.

booxter:
recheck etcdserver: request timed out

booxter:
recheck RETRY_LIMIT

fao89:
recheck

jpodivin:
recheck

booxter:
recheck

booxter:
The failure is legit

```
2023-08-29T16:36:54.475110792+00:00 stdout F TASK [osp.edpm.edpm_ovn : Load ovsdb configuration from file] ******************
2023-08-29T16:36:54.476636615+00:00 stdout F Tuesday 29 August 2023  16:36:54 +0000 (0:00:00.466)       0:00:09.481 ******** 
2023-08-29T16:36:54.538751534+00:00 stdout F [0;31mfatal: [edpm-compute-0 -> localhost]: FAILED! => {"ansible_facts": {"edpm_ovn_configmap_facts": {}}, "ansible_included_var_files": [], "changed": false, "message": "Could not find or access '/var/lib/openstack/configs/ovsdb-config' on the Ansible Controller.\nIf you are using a module and expect the file to exist on the remote, see the remote_src option"}[0m
```

roles/edpm_ovn/tasks/configure.yml
```diff
@@ -14,6 +14,25 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Apply ovsdb configuration from configMaps
+  block:
+  - name: Check that config file exists
+    stat:
+      path: "{{ edpm_ovn_config_src }}/ovsdb-config"
+    register: ovsdb_config_exists
```
booxter, line 22: this should delegate to localhost too I think

roles/edpm_ovn/tasks/configure.yml
```diff
@@ -14,6 +14,25 @@
...
+
+  - name: Load ovsdb configuration from file
+    ansible.builtin.include_vars:
+      file: "{{ edpm_ovn_config_src }}/ovsdb-config"
+      name: edpm_ovn_configmap_facts
+    delegate_to: localhost
+    when:
+    - ovsdb_config_exists
```
booxter, line 30: this check is invalid, I should check the "exists": true here. That's why this task is executed in CI even when the previous one couldn't find the file.

booxter: (I should add a scenario for missing file)

roles/edpm_ovn/defaults/main.yml
```diff
@@ -1,5 +1,6 @@
 ---
 # defaults file for edpm_ovn
+edpm_ovn_config_src: /var/lib/openstack/configs
```
booxter, line 3: the actual file path includes `ovn`, e.g. `/var/lib/openstack/configs/ovn/ovsdb-config`

slagle: correct, the "ovn" comes from the service name on the dataplane-operator side

roles/edpm_ovn/tasks/configure.yml
booxter: Added new scenario for missing config file.

roles/edpm_ovn/tasks/configure.yml
booxter: Moved delegation to localhost.

booxter:
recheck

booxter:
The failure seems legit, though I haven't seen it before, maybe something broke in tree in flight?

```
[pod/dataplane-deployment-ovn-edpm-compute-gtrpw/openstackansibleee] TASK [osp.edpm.edpm_ovn : Create persistent directories] ***********************
[pod/dataplane-deployment-ovn-edpm-compute-gtrpw/openstackansibleee] [1;30mtask path: /usr/share/ansible/collections/ansible_collections/osp/edpm/roles/edpm_ovn/tasks/install.yml:17[0m
[pod/dataplane-deployment-ovn-edpm-compute-gtrpw/openstackansibleee] Tuesday 29 August 2023  23:41:39 +0000 (0:00:00.987)       0:00:03.639 ******** 
[pod/dataplane-deployment-ovn-edpm-compute-gtrpw/openstackansibleee] [0;31mfailed: [edpm-compute-0] (item={'path': '/var/log/containers/openvswitch', 'setype': 'container_file_t', 'mode': '0750'}) => {"ansible_loop_var": "item", "changed": false, "item": {"mode": "0750", "path": "/var/log/containers/openvswitch", "setype": "container_file_t"}, "msg": "There was an issue creating /var/log/containers/openvswitch as requested: [Errno 13] Permission denied: b'/var/log/containers/openvswitch'", "path": "/var/log/containers/openvswitch"}[0m
[pod/dataplane-deployment-ovn-edpm-compute-gtrpw/openstackansibleee] [0;31mfailed: [edpm-compute-0] (item={'path': '/var/log/containers/neutron', 'setype': 'container_file_t', 'mode': '0750'}) => {"ansible_loop_var": "item", "changed": false, "item": {"mode": "0750", "path": "/var/log/containers/neutron", "setype": "container_file_t"}, "msg": "There was an issue creating /var/log/containers/neutron as requested: [Errno 13] Permission denied: b'/var/log/containers/neutron'", "path": "/var/log/containers/neutron"}[0m
[pod/dataplane-deployment-ovn-edpm-compute-gtrpw/openstackansibleee] [0;31mfailed: [edpm-compute-0] (item={'path': '/var/lib/openvswitch/ovn', 'setype': 'container_file_t'}) => {"ansible_loop_var": "item", "changed": false, "item": {"path": "/var/lib/openvswitch/ovn", "setype": "container_file_t"}, "msg": "There was an issue creating /var/lib/openvswitch/ovn as requested: [Errno 13] Permission denied: b'/var/lib/openvswitch/ovn'", "path": "/var/lib/openvswitch/ovn"}[0m
```

playbooks/ovn.yaml
```diff
@@ -2,7 +2,6 @@
 - name: Deploy EDPM OVN
   hosts: all
   strategy: linear
-  become: true
```
booxter, line 5: maybe this should not be removed... since we are getting permission denied in baremetal job.

slagle: Either that or you need to add it to all the tasks where it's needed

booxter:
OK, looks like the environment used for molecules is not the same as used for actual ansibleee images, hence failures. How do we make molecule env identical to ansibleee?.. Until we sync the environments somehow, my attempts to fix the CI failures are blind because I can't effectively reproduce the issue.

I can of course build a full cluster and then override edpm-ansible through NFS or otherwise, but that's not very effective, comparing to molecule...

karelyatin:
> OK, looks like the environment used for molecules is not the same as used for actual ansibleee images, hence failures. How do we make molecule env identical to ansibleee?.. Until we sync the environments somehow, my attempts to fix the CI failures are blind because I can't effectively reproduce the issue.
> 

This looks is happening as different image is used in molecule which have sudo installed https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/molecule/common/Containerfile.j2#L10 vs ansiblee which don't have sudo(https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/openstack_ansibleee/Dockerfile). And with global become: and delegate: localhost tasks it doesn't work
> I can of course build a full cluster and then override edpm-ansible through NFS or otherwise, but that's not very effective, comparing to molecule...
Seems same issue could be reproduced by not installing sudo as part of Containerfile.j2? Below reproduces the current issue as seen in CI as sudo is missing in the image.
apiVersion: ansibleee.openstack.org/v1alpha1
kind: OpenStackAnsibleEE
metadata:
  name: ansibleee-playbook-local
  namespace: openstack
spec:
  play: |
    - name: Print hello world
      hosts: all
      become: true
      tasks:
      - name: Using debug statement
        become: false
        ansible.builtin.debug:
          msg: "Hello, world! This is openstack-ansibleee-play.yaml"
  inventory: |
    all:
      vars:
        ansible_connection: local
      hosts:
        localhost
  env:
    - name: ANSIBLE_FORCE_COLOR
      value: "True"
    - name: ANSIBLE_SSH_ARGS
      value: "-C -o ControlMaster=auto -o ControlPersist=80s"
    - name: ANSIBLE_ENABLE_TASK_DEBUGGER
      value: "True"

oc logs job/ansibleee-playbook-local -f

booxter:
recheck

booxter:
recheck

booxter:
recheck

booxter:
recheck

booxter:
recheck

PR #290, opened 2023-08-22
Do not require a source address for ceph_hci_pre firewall

fultonj:
If variables matching edpm_ceph_hci_pre_*_ranges are set to an empty list, then do not specify a source address in the NFT rule. For example the rule will look like this:
```
  tcp dport { 3300, 6789 } ...
```
If the source address is specified, then the rule will look like this:
```
  ip saddr 172.18.0.0/24 tcp dport { 3300, 6789 } ...
```
Prior to this patch, if someone wanted to not pass a source address they would need to add 0.0.0.0/0 to the list which is unintuitive. It is better to handle when there is no source address by not creating a rule with a source address.

TripleO does not specify a source IP address when defining Ceph rules in iptables. A line from /etc/sysconfig/iptables in Wallaby TripleO CI is below.
```
-A INPUT -p tcp -m tcp --dport 3300 -m conntrack \
  --ctstate NEW -m comment --comment "110 ceph_mon ipv4" \
  -j ACCEPT
```
The default firewall rules in the edpm_ceph_hci_pre role are being updated in this patch to have the functional equivalent of the above in nftables. The option remains to override the defaults and pass a list of source addresses. It is assumed the storage and storage management networks will be isolated.

### DIFF
--- a/roles/edpm_ceph_hci_pre/defaults/main.yml
+++ b/roles/edpm_ceph_hci_pre/defaults/main.yml
@@ -22,21 +22,23 @@
 # Firewall configuration
 edpm_ceph_hci_pre_configure_firewall: true
 
-# Only allow default storage network to access Ceph RBD/Mon
-edpm_ceph_hci_pre_storage_ranges:
-  - 172.18.0.0/24
+# Name of file to be created on EDPM nodes with ceph firewall rules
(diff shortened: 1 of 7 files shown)


fmount (approved):
Looks good, thanks John

fultonj:
recheck

PR #291, opened 2023-08-23
Replace Ubuntu with Fedora

bshephar:
I noticed a failure to pull the ubuntu image in CI. I'm not sure why we need Ubuntu and especially such an old release. Someone feel free to advise why it's necessary. This change is replacing the Ubuntu image with Fedora.

### DIFF
--- a/roles/edpm_podman/molecule/login/prepare.yml
+++ b/roles/edpm_podman/molecule/login/prepare.yml
@@ -54,10 +54,10 @@
         name: registry
         state: absent
 
-    - name: Pull ubuntu image
+    - name: Pull Fedora image
       containers.podman.podman_image:
-        name: ubuntu
-        tag: 16.04
+        name: fedora
+        tag: latest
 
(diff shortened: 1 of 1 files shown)


jpodivin (commented):
You are right. Ubuntu was probably just inherited from OOO test infrastructure and we haven't changed it since. I'm just wondering if it is good idea to use rawhide here. Maybe we could pick a version and pin the test to that for better stability?

bshephar:
> You are right. Ubuntu was probably just inherited from OOO test infrastructure and we haven't changed it since. I'm just wondering if it is good idea to use rawhide here. Maybe we could pick a version and pin the test to that for better stability?

I'm not sure it really matters too much since we're not doing anything with it. I just didn't want to have to update it from 37 > 38 > 39 periodically. But I pushed a new revision changing the tag to latest which will get the latest stable.

slagle (approved):

PR #292, opened 2023-08-23
edpm_libivirt should reload NFT ruleset after configuring it

fultonj:
The post-install tasks file was calling the configure tasks in the edpm_nftables role to update the rules file in the /etc/nftables/ directory, but was not calling the run tasks to make the firewall reload the ruleset. This caused the output of `nft list ruleset` to not include the VNC rule which the edpm_libivirt role adds. Fix this by calling the run tasks file.

Jira: [OSP-27573](https://issues.redhat.com//browse/OSP-27573)

### DIFF
--- a/roles/edpm_libvirt/tasks/post-install.yml
+++ b/roles/edpm_libvirt/tasks/post-install.yml
@@ -44,3 +44,10 @@
   ansible.builtin.include_role:
       name: osp.edpm.edpm_nftables
       tasks_from: "configure.yml"
+- name: Reload firewall for new vnc rule
+  tags:
+    - install
+    - post-libvirt
+  ansible.builtin.include_role:
+      name: osp.edpm.edpm_nftables
+      tasks_from: "run.yml"
--- a/roles/edpm_libvirt/molecule/default/prepare.yml
+++ b/roles/edpm_libvirt/molecule/default/prepare.yml
@@ -36,3 +36,20 @@
     - name: install podman
       import_role:
         name: "edpm_podman"
+    - name: Create firewall directory
+      become: true
+      ansible.builtin.file:
+        path: '/var/lib/edpm-config/firewall'
+        state: directory
+        owner: root
+        group: root
+        mode: 0750
+    - name: open port 22 (edpm_nftables will active this later)
+      become: true
+      copy:
+        dest: /var/lib/edpm-config/firewall/sshd-networks.yaml
+        content: |
+          - rule_name: 003 Allow SSH
+            rule:
+              proto: tcp
+              dport: 22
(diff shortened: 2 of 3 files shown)


roles/edpm_libvirt/tasks/post-install.yml
```diff
@@ -43,3 +43,10 @@
   ansible.builtin.include_role:
       name: osp.edpm.edpm_nftables
       tasks_from: "configure.yml"
+- name: Reload firewall for new vnc rule
+  tags:
+    - install
+    - post-libvirt
+  ansible.builtin.include_role:
+      name: osp.edpm.edpm_nftables
+      tasks_from: "run.yml"
```
SeanMooney, line 52: ok yes this is a bug intoduced in the porting.

hum ok I'm not quite sure how this was working before but yes this is required.

SeanMooney (approved):

roles/edpm_libvirt/tasks/post-install.yml
SeanMooney: by the way it would be nice to add a molecule test to assert that the rule is actually reloaded. i.e. to exentd https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_libvirt/molecule/default/test-helpers/verify_firewall.yaml
to also grep the output of the nft rules.

fultonj:
recheck

fultonj:
recheck

SeanMooney:
the job failues in the kuttle jobs are because of errors in Play: Run ci/playbooks/collect-logs.yml in the ci framework.

SeanMooney:
https://github.com/openstack-k8s-operators/ci-framework/blob/1df1cf115de7288e6a558ba4b8958528538aa5ce/ci/playbooks/collect-logs.yml#L82-L87

this is what is failing

fultonj:
recheck

fultonj:
I wonder if molecule is failing because it's actually applying a firewall rule which doesn't include SSH and this is now the first time it's actually applying the firewall rule. I'm going to try a molecule change next.

slagle (approved):

slagle:
recheck

fultonj:
recheck

roles/edpm_libvirt/tasks/post-install.yml
fultonj: Here's that molecule test https://github.com/openstack-k8s-operators/edpm-ansible/pull/306

fultonj:
> im going to add a test in a follow up since i have a dev env where i can do this so approving as is.

Here's the follow up test https://github.com/openstack-k8s-operators/edpm-ansible/pull/306

PR #294, opened 2023-08-23
Test ovn role in zuul

booxter:


jpodivin (approved):

jpodivin (changes requested):
Actually, one thing. You need to include the new job in the `projects.yaml`, otherwise it won't run.

karelyatin:
The failure looks legit msg: 'Could not find the requested service openvswitch: host' https://logserver.rdoproject.org/94/294/7cee4f3ee7f5346a3e21cd373579e4fde0c737aa/github-check/edpm-ansible-molecule-edpm_ovn/2af11f0/job-output.txt

booxter:
Looks like our role / molecule doesn't install ovs on the compute node, so it fails to start if the service is not present, yet. We should make the role install the package on the EDP node before trying to enable the service.

slagle:
> Looks like our role / molecule doesn't install ovs on the compute node, so it fails to start if the service is not present, yet. We should make the role install the package on the EDP node before trying to enable the service.

It's taken care of by edpm_bootstrap as openvswitch is likely needed for the network configuration, which happens before ovn. You could install it in the molecule setup though.

booxter:
Installing ovs in bootstrap for the role just in case bootstrap is not executed (and to guarantee the package / service in molecule env). Also added the molecule in github workflow config.

booxter:
recheck SSL timeout when fetching kubectl from github

booxter:
Looks like pulling `podman` play in molecule scenario has its own issues?

```
TASK [edpm_podman : Delete legacy cni0 interface (podman < 1.6)] ***************
fatal: [instance]: FAILED! =>
  msg: |-
    The conditional check '"cni0" in ansible_facts["interfaces"]' failed. The error was: error while evaluating conditional ("cni0" in ansible_facts["interfaces"]): 'dict object' has no attribute 'interfaces'. 'dict object' has no attribute 'interfaces'

    The error appears to be in '/home/zuul/src/github.com/openstack-k8s-operators/edpm-ansible/roles/edpm_podman/tasks/install.yml': line 39, column 7, but may
    be elsewhere in the file depending on the exact syntax problem.

    The offending line appears to be:


        - name: Delete legacy cni0 interface (podman < 1.6)
          ^ here
```

karelyatin:
With this patch it's tested twice as in:-
- @github-actions Run Molecule tests / run_molecule (edpm_ovn) (pull_request) Successful in 7m 
- [edpm-ansible-molecule-edpm_ovn ](https://review.rdoproject.org/zuul/build/0d8499cd3cab441a9b35ceed9f95850c)FAILURE in 7m 12s

I think just 1 of them should be enough, github-action job is already green here and in my other PR https://github.com/openstack-k8s-operators/edpm-ansible/pull/309 with some newer tests.

I also tried to make zuul job green as part of https://github.com/openstack-k8s-operators/edpm-ansible/pull/315, that required couple of changes to make it work. So if there is no real reason to enable zuul job then we can keep it for later when required.

booxter:
Superseded by https://github.com/openstack-k8s-operators/edpm-ansible/pull/309

PR #296, opened 2023-08-25
Fix run_ovn_bgp_agent playbook definition

luis5tb:
Ensure the tasks is defined and imports the run role, instead of defining a role as this fails with:
    Validating arguments against arg spec 'main' - The main entry point for
    the edpm_ovn_bgp_agent role.] ***


jpodivin (approved):

fao89:
recheck

fao89:
recheck

PR #301, opened 2023-08-28
edpm_kernel: Fix Reboot Task

vkhitrin:
Reboot task may fail due to the remote user being an unprivileged user
(non root) which is not able to perform a reboot.
By escalating the permissions in the task, the reboot task should pass.


vkhitrin:
Planning to test it on the baremetal EDPM node, which encountered this issue initially.

jpodivin (approved):

jpodivin:
recheck

PR #302, opened 2023-08-28
Store Ceph client files in /var/lib/openstack/config/ceph

fultonj:
On EDPM nodes Ceph client files were being stored in /etc/ceph. In an HCI scenario /etc/ceph could also host Ceph server files. To prevent any possible conflict store the client files in /var/lib/openstack/config/ceph. Inside the continers the path will remain /etc/ceph even though it changed on the host.

Jira: [OSP-27777](https://issues.redhat.com//browse/OSP-27777)

### DIFF
--- a/roles/edpm_ceph_client_files/defaults/main.yml
+++ b/roles/edpm_ceph_client_files/defaults/main.yml
@@ -18,5 +18,9 @@
 # All variables intended for modification should be placed in this file.
 
 # All variables within this role should have a prefix of "edpm_ceph_client_files"
+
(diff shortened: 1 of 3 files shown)


fmount (approved):
This change updates the path from `/etc/ceph` to `/var/lib/openstack/config/ceph`, which is fine and something we already agreed in previous conversations.
I understand you're going to update the `post-install.yaml` script in a follow up change, hence I'm approving this one as it looks good for the purpose of moving the path.

PR #305, opened 2023-08-30
New role parameters to correctly handle templates

jpodivin:


slagle (changes requested):
I don't believe all these should be required:true. At the same time, I don't have recommendations for sane defaults, other than what may be gathered from config/samples in dataplane-operator or install_yamls.

I also think the nic template examples for this role can go away when we merge the following:

https://github.com/openstack-k8s-operators/edpm-ansible/pull/243
https://github.com/openstack-k8s-operators/dataplane-operator/pull/284/

jpodivin:
> I don't believe all these should be required:true. At the same time, I don't have recommendations for sane defaults, other than what may be gathered from config/samples in dataplane-operator or install_yamls.
> 
> I also think the nic template examples for this role can go away when we merge the following:
> 
> #243 [openstack-k8s-operators/dataplane-operator#284](https://github.com/openstack-k8s-operators/dataplane-operator/pull/284)

There may be a way around it. We could wrap the parameters in a dictionary and look for them there. That would give us necessary flexibility.

jpodivin:
I'm going to put this on hold until [#284](https://github.com/openstack-k8s-operators/dataplane-operator/pull/284) merges.

PR #307, opened 2023-08-30
Document building an ephemeral VM for local molecule testing

fultonj:
Update the testing roles documentation with complete steps to build a new CentOS9 stream VM which can run local molecule tests as resquired by the edpm_libvirt and edpm_nova roles.

### DIFF
--- a/docs/source/testing_roles.rst
+++ b/docs/source/testing_roles.rst
@@ -35,6 +35,62 @@ be sufficient for the `molecule test` command to work.
 
 Alternatively you can test roles in with :ref:`AnsibleEE<testing with ansibleee>`.
 
+Creating a VM to run molecule
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The
+[edpm_libvirt](https://github.com/openstack-k8s-operators/edpm-ansible/tree/main/roles/edpm_libvirt)
+molecule senario uses the delegated driver to run against localhost.
+Thus, it's better to run it inside an ephemeral virtual machine as
+it cannot be run insider of a container.
+
+Boot a VM based on `CentOS-Stream-GenericCloud-9-latest.x86_64.qcow2`
+from https://cloud.centos.org/centos/9-stream/x86_64/images.
+
+Use [repo-setup](https://github.com/openstack-k8s-operators/repo-setup)
+to configure the DNF repositories of the VM.
+
+.. code-block:: console
+
+    pushd /tmp
+    curl -sL https://github.com/openstack-k8s-operators/repo-setup/archive/refs/heads/main.tar.gz | tar -xz
+    pushd repo-setup-main
+    python3 -m venv ./venv
+    PBR_VERSION=0.0.0 ./venv/bin/pip install ./
+    sudo ./venv/bin/repo-setup current-podified-dev
+    popd
+    popd
+
+Clone the git repository on the VM.
+
+.. code-block:: console
+
+    cd ~
+    sudo dnf install -y git
+    git clone git@github.com:openstack-k8s-operators/edpm-ansible.git
+
+Prepare molecule.
+
+.. code-block:: console
+
+    pushd ~/edpm-ansible
+    python3 -m venv molecule-venv
+    source molecule-venv/bin/activate
+    pip install --upgrade pip
+    pip install bindep
+    bindep -b | xargs sudo dnf -y --setopt=install_weak_deps=False install
+    pip install -r molecule-requirements.txt
+    popd
+
+Consider taking a snapshot of the VM before running a test like the
+following.
+
+.. code-block:: console
+
+    cd ~/edpm-ansible/roles/edpm_libvirt/
+    molecule test --all
+
+
 Writing molecule tests
 ~~~~~~~~~~~~~~~~~~~~~~
 


docs/source/testing_roles.rst
```diff
@@ -35,6 +35,58 @@ be sufficient for the `molecule test` command to work.
...
+
+    pushd /tmp
+    curl -sL https://github.com/openstack-k8s-operators/repo-setup/archive/refs/heads/main.tar.gz | tar -xz
+    pushd repo-setup-main
+    python3 -m venv ./venv
+    PBR_VERSION=0.0.0 ./venv/bin/pip install ./
+    ./venv/bin/repo-setup current-podified-dev
+    popd
+
+Close the git repository on the VM.
```
SeanMooney, line 63: s/Close/clone/

docs/source/testing_roles.rst
```diff
@@ -35,6 +35,58 @@ be sufficient for the `molecule test` command to work.
...
+    pip install -r molecule-requirements.txt
+    popd
+
+Consider taking a snapshot of the VM before running a test like the
+following.
+
+.. code-block:: console
+
+    cd roles/edpm_libvirt/
+    molecule test --all
```
SeanMooney, line 87: technially --all is only needed if you have multiple scenarios but ya in general that would be the portable way to do that.

docs/source/testing_roles.rst
```diff
@@ -35,6 +35,58 @@ be sufficient for the `molecule test` command to work.
...
+Prepare molecule.
+
+.. code-block:: console
+
+    pushd edpm-ansible
+    python3 -m venv molecule-venv
+    source molecule-venv/bin/activate
+    pip install --upgrade pip
+    pip install -r molecule-requirements.txt
+    popd
```
SeanMooney, line 79: so you may also need to install packages listed in bindep depending on the image you use.
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/bindep.txt

i.e. 

pip install bindep
bindep -b | xargs sudo dnf -y --setopt=install_weak_deps=False install

this may not be required for the nova/libvirt roles but it likely is for other roles

docs/source/testing_roles.rst
fultonj: fixed

docs/source/testing_roles.rst
fultonj: OK, I'll add that. Thanks Sean.

SeanMooney:
rdo-check

im  going to quickly follow this in a clean vm but overall it looks fine to me

docs/source/testing_roles.rst
```diff
@@ -35,6 +35,60 @@ be sufficient for the `molecule test` command to work.
...
+    PBR_VERSION=0.0.0 ./venv/bin/pip install ./
+    ./venv/bin/repo-setup current-podified-dev
+    popd
+
+Clone the git repository on the VM.
+
+.. code-block:: console
+
+    dnf install -y git
+    git clone git@github.com:openstack-k8s-operators/edpm-ansible.git
```
SeanMooney, line 68: nit if you follow this as written we are still in 
/tmp/repo-setup-main
so maybe we should add a cd ~ or cd .. but that not really that imporant

docs/source/testing_roles.rst
SeanMooney: this block works as written.
althogh it would be better to do the bindep install after sorucing the venv on line
77 in case and of the deps in molecule-requiremets.txt need to compile c moudules

docs/source/testing_roles.rst
```diff
@@ -35,6 +35,60 @@ be sufficient for the `molecule test` command to work.
...
+    pip install bindep
+    bindep -b | xargs sudo dnf -y --setopt=install_weak_deps=False install
+    popd
+
+Consider taking a snapshot of the VM before running a test like the
+following.
+
+.. code-block:: console
+
+    cd roles/edpm_libvirt/
```
SeanMooney, line 88: nit your actully in the dir which has the git repo because of the popd
at the end fo the previous block

so this should be cd edpm-ansible/roles/edpm_libvirt/

SeanMooney: ingnoring that i was able to run the test in my VM at this point.

SeanMooney (commented):
so if you ran this as root and accounted for the minor nits inline then this does indeed work on a clean centos 9 strema VM

it will probably also work fine on fedora or rhel9 but i have not
tested either.

would you mind one more quick respin for the nit
other then that i think we can merge this.

docs/source/testing_roles.rst
```diff
@@ -35,6 +35,60 @@ be sufficient for the `molecule test` command to work.
...
+to configure the DNF repositories of the VM.
+
+.. code-block:: console
+
+    pushd /tmp
+    curl -sL https://github.com/openstack-k8s-operators/repo-setup/archive/refs/heads/main.tar.gz | tar -xz
+    pushd repo-setup-main
+    python3 -m venv ./venv
+    PBR_VERSION=0.0.0 ./venv/bin/pip install ./
+    ./venv/bin/repo-setup current-podified-dev
```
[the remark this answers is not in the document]
fultonj: done

docs/source/testing_roles.rst
```diff
@@ -35,6 +35,60 @@ be sufficient for the `molecule test` command to work.
...
+    popd
+
+Clone the git repository on the VM.
+
+.. code-block:: console
+
+    dnf install -y git
```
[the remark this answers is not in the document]
fultonj: done

docs/source/testing_roles.rst
fultonj: OK, i added `cd ~`

docs/source/testing_roles.rst
fultonj: fixed the directory locations

fultonj: I moved the bindep command to before the pip install of the molecule-requirements

SeanMooney (approved):

fultonj:
recheck

PR #308, opened 2023-08-31
Remove ci-framework network template

pablintino:
Since we are now injecting the networking using the networkConfig.template this is not needed anymore.


jpodivin (approved):
Looking for the usage of the template I really couldn't find any. I see this as a clear improvement.

PR #309, opened 2023-09-05
[ovn] Add missing firewall rules

karelyatin:
Without these genev/vxlan traffic across
edpm nodes is dropped.


karelyatin:
recheck known edpm issues fixed

slagle (approved):

PR #311, opened 2023-09-06
Removing json_query calls

jpodivin:
We aren't going to be shipping required package, so we need to work around this. In most cases the query wasn't necessary in the first place. The exception being the jump targets template, in which case it was replaced by a specialized filter.

Note that the filter isn't general, it can not be, otherwise we would have to reimplement json query, or get another dependency. Neither of which is something we need or should pursue. Therefore, if there comes a time when the template is adjusted, the filter will need some changes as well.

### DIFF
--- a/plugins/filter/helpers.py
+++ b/plugins/filter/helpers.py
@@ -25,6 +25,7 @@ class FilterModule:
             'needs_delete': self.needs_delete,
             'haskey': self.haskey,
             'dict_to_list': self.dict_to_list,
+            'jump_chain_targets': self.jump_chain_targets
         }
 
     def needs_delete(self, container_infos, config, config_id,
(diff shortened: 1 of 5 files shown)


fao89:
recheck

fao89 (approved):

rabi:
Failure looks related to the change.

2023-09-11T09:41:54.980108687+00:00 stdout F 
2023-09-11T09:41:54.980108687+00:00 stdout F TASK [osp.edpm.edpm_nftables : Generate chain jumps] ***************************
2023-09-11T09:41:54.983465046+00:00 stdout F Monday 11 September 2023  09:41:54 +0000 (0:00:00.850)       0:00:59.713 ****** 
2023-09-11T09:41:56.189074710+00:00 stdout F [0;32mok: [edpm-compute-0] => {"changed": false, "checksum": "4c6f036d2d5808f109acc0880c19aa74ca48c961", "dest": "/etc/nftables/edpm-jumps.nft", "gid": 0, "group": "root", "mode": "0600", "owner": "root", "path": "/etc/nftables/edpm-jumps.nft", "secontext": "system_u:object_r:etc_t:s0", "size": 98, "state": "file", "uid": 0}[0m
2023-09-11T09:41:56.208915176+00:00 stdout F 
2023-09-11T09:41:56.208915176+00:00 stdout F TASK [osp.edpm.edpm_nftables : Generate chain jumps] ***************************
2023-09-11T09:41:56.209820980+00:00 stdout F Monday 11 September 2023  09:41:56 +0000 (0:00:01.227)       0:01:00.941 ****** 
2023-09-11T09:41:56.537975060+00:00 stdout F [0;31mAn exception occurred during task execution. To see the full traceback, use -vvv. The error was: AttributeError: 'list' object has no attribute 'get'[0m
2023-09-11T09:41:56.537975060+00:00 stdout F [0;31mfatal: [edpm-compute-0]: FAILED! => {"changed": false, "msg": "AttributeError: 'list' object has no attribute 'get'"}[0m

jpodivin:
@rabi Where is that log from?

fao89:
> @rabi Where is that log from?

https://logserver.rdoproject.org/11/311/62d890ad214f550b7ed40e755ffe0f6420e3f302/github-check/podified-multinode-edpm-deployment-crc/48f563d/controller/ci-framework-data/logs/crc/pods/openstack_dataplane-deployment-libvirt-openstack-edpm-dkgw9_ffa85b4b-b8d1-4a2a-8130-4228119d3798/openstackansibleee/0.log

fao89 (approved):

PR #314, opened 2023-09-07
Reduce the idle timeout

fao89:
When the job fails, it waits 10 minutes to kill the job. Reducing it to 5 minutes

### DIFF
--- a/openstack_ansibleee/settings
+++ b/openstack_ansibleee/settings
@@ -1,6 +1,6 @@
 # If no output is detected from ansible in this number of seconds the execution will
 # be terminated.
-idle_timeout: ${RUNNER_IDLE_TIMEOUT:-600}
+idle_timeout: ${RUNNER_IDLE_TIMEOUT:-300}
 # The maximum amount of time to allow the job to run for, exceeding this and the
 # execution will be terminated.
 job_timeout: ${RUNNER_JOB_TIMEOUT:-3600}


fao89:
recheck

jpodivin (commented):
Would it make more sense to adjust the variable in the cases when it idles for too long?

fao89:
I think 10 minutes is too long, do we have any task that might idle for that long?
In any case, how can we know beforehand a task might idle for 5+ minutes?

openstack_ansibleee/settings
```diff
@@ -1,6 +1,6 @@
 # If no output is detected from ansible in this number of seconds the execution will
 # be terminated.
-idle_timeout: ${RUNNER_IDLE_TIMEOUT:-600}
```
rabi, line 3: I think this is the ansible runner default https://ansible.readthedocs.io/projects/runner/en/stable/intro/#env-settings-settings-for-runner-itself. Not sure if we should reduce it.

We probably could find the better way to know that the job has failed?

fao89: I tried to find something in the ansible-runner to understand why it hangs when it fails, but I didn't find anything

fao89:
I'll close as it is not the real issue, the real issue is to avoid ansible hanging when it fails

fao89:
> I'll close as it is not the real issue, the real issue is to avoid ansible hanging when it fails

this was the root cause: https://github.com/ansible/ansible-runner/issues/1327

PR #315, opened 2023-09-08
Ovn test

karelyatin:


booxter:
@karelyatin should it be discarded? I believe we enabled molecule tests for the ovn already.

karelyatin:
@booxter yes this was just test to run molecule tests differently in zuul job, since molecule tests running fine as part of github workflows will abandon this PR. If needed in future for some cases it can be restored.

PR #316, opened 2023-09-08
Add edpm role to deploy Neutron SRIOV agent

slawqo:
This new roles deploys, configures and runs neutron-sriov-agent on the EDPM node if needed.

### DIFF
--- a/playbooks/neutron_sriov.yaml
+++ b/playbooks/neutron_sriov.yaml
@@ -1,0 +1,11 @@
+---
+- name: Deploy EDPM Neutron SR-IOV agent
+  hosts: all
+  strategy: linear
+  become: true
+  tasks:
+    - name: Neutron SR-IOV agent
+      import_role:
+        name: osp.edpm.edpm_neutron_sriov
+      tags:
+        - edpm_neutron_sriov
--- a/roles/edpm_neutron_sriov/defaults/main.yml
+++ b/roles/edpm_neutron_sriov/defaults/main.yml
@@ -1,0 +1,71 @@
+---
+# Copyright 2023 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+# All variables intended for modification should be placed in this file.
+
+# All variables within this role should have a prefix of "edpm_neutron_sriov_agent"
+edpm_neutron_sriov_agent_config_dir: "/var/lib/config-data/ansible-generated/neutron-sriov-agent"
+edpm_neutron_sriov_image: "quay.io/podified-antelope-centos9/openstack-neutron-sriov-agent:current-podified"
+
+edpm_neutron_sriov_common_volumes:
+  - /lib/modules:/lib/modules:ro
+  - /dev:/dev
+  - "{{ edpm_neutron_sriov_agent_config_dir }}:/etc/neutron:z"
+  - /var/lib/neutron:/var/lib/neutron:shared,z
+  - /var/lib/kolla/config_files/neutron_sriov_agent.json:/var/lib/kolla/config_files/config.json:ro
+  - /var/log/containers/neutron:/var/log/neutron:z
+
+# neutron.conf
+# DEFAULT
+edpm_neutron_sriov_DEFAULT_debug: false
+edpm_neutron_sriov_DEFAULT_rpc_response_timeout: 60
+edpm_neutron_sriov_DEFAULT_transport_url: ''
+edpm_neutron_sriov_DEFAULT_host: '{{ ansible_facts["nodename"] }}'
+# oslo_concurrency
+edpm_neutron_sriov_oslo_concurrency_lock_patch: '$state_path/lock'
+# oslo_messaging_rabbit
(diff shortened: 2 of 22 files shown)


roles/edpm_neutron_sriov/defaults/main.yml
```diff
@@ -0,0 +1,71 @@
...
+
+# All variables intended for modification should be placed in this file.
+
+# All variables within this role should have a prefix of "edpm_neutron_sriov_agent"
+edpm_neutron_sriov_config_dir: "/var/lib/config-data/ansible-generated/neutron/sriov_agent"
+edpm_neutron_sriov_log_dir: "/var/log/containers/neutron"
+edpm_neutron_sriov_image: "quay.io/podified-antelope-centos9/openstack-neutron-sriov-agent:current-podified"
+
+edpm_neutron_sriov_common_volumes:
+  - /lib/modules:/lib/modules:ro
```
booxter, line 26: (Wondering) Why does it need /lib/modules? Does it load kernel modules?

booxter:
recheck

booxter:
btw you have some issues in arg_specs

```
ERROR:antsibull:func=create_plugin_rst:mod=antsibull_docs.write_docs.plugins:nonfatal_errors=["Error while processing role: We were unable to read either as JSON nor YAML, these are the errors we got from each:\nJSON: Expecting value: line 1 column 1 (char 0)\n\nSyntax Error while loading YAML.\n  mapping values are not allowed in this context\n\nThe error appears to be in '/home/runner/.ansible/collections/ansible_collections/osp/edpm/roles/edpm_neutron_sriov/meta/argument_specs.yml': line 34, column 16, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n      edpm_neutron_sriov_DEFAULT_debug: false\n        default: false\n               ^ here\n"]:plugin_name=osp.edpm.edpm_neutron_sriov:plugin_type=role|osp.edpm.edpm_neutron_sriov did not return correct DOCUMENTATION.  An error page will be generated.
```

roles/edpm_neutron_sriov/defaults/main.yml
```diff
@@ -0,0 +1,71 @@
...
+  - /dev:/dev
+  - "{{ edpm_neutron_sriov_config_dir }}/etc/neutron:/etc/neutron:z"
+  - /run/netns:/run/netns:shared
```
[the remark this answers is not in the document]
slawqo: Good question. I will need to check what volumes such container have in 17.1 and make it similar

roles/edpm_neutron_sriov/defaults/main.yml
slawqo: same here

roles/edpm_neutron_sriov/defaults/main.yml
```diff
@@ -0,0 +1,71 @@
...
+  - "{{ edpm_neutron_sriov_log_dir }}:/var/log/neutron"
+  - /var/lib/neutron:/var/lib/neutron:shared,z
+  - /var/lib/kolla/config_files/neutron_sriov_agent.json:/var/lib/kolla/config_files/config.json:ro
+
+# neutron.conf
+# DEFAULT
+edpm_neutron_sriov_DEFAULT_debug: false
+edpm_neutron_sriov_DEFAULT_rcp_response_timeout: 60
+edpm_neutron_sriov_DEFAULT_transport_url: ''
```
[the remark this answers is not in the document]
slawqo: We may also want to derive from CP value of the "edpm_neutron_sriov_agent_AGENT_report_interval" as this is related to the "agent_down_time" which is configured on the neutron-server side

roles/edpm_neutron_sriov/defaults/main.yml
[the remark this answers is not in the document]
slawqo: I removed it

roles/edpm_neutron_sriov/defaults/main.yml
slawqo: this is already also used in container in 17.1 so I'm not sure I want/should change it now

booxter: To clarify, I haven't requested to remove, just wondering. We may revisit the list later.

booxter:
recheck

slawqo:
recheck

booxter:
recheck

playbooks/neutron_sriov.yaml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Deploy EDPM Neutron SR-IOV agent
+  hosts: all
+  strategy: linear
+  become: true
+  tasks:
+    - name: ovn
```
karelyatin, line 7: s/ovn/Neutron SR-IOV agent

karelyatin (commented):
a nit, jobs failing so can push an update to trigger jobs :)

playbooks/neutron_sriov.yaml
slawqo: Done :)

roles/edpm_neutron_sriov/tasks/install.yml
```diff
@@ -0,0 +1,44 @@
...
+    - install
+    - neutron
+
+- name: render neutron-sriov-agent container
+  become: true
+  ansible.builtin.template:
+    src: "neutron_sriov_agent.yaml.j2"
+    dest: "/var/lib/openstack/config/containers/neutron_sriov_agent.yaml"
+    setype: "container_file_t"
+    mode: 0644
```
booxter, line 39: I was told by someone we should put these modes in double-quotes for ansible, just in case. But I think it works either way. You may want to follow up with a tiny patch to adjust this after this PR merges.

booxter:
recheck

booxter:
The latest failure doesn't seem relevant:

```
http: TLS handshake timeout
```

when calling to k8s api.

Also, the patch doesn't even enable the sriov role, so it's not engaged in deployment.

booxter:
recheck

`error: You must be logged in to the server (Unauthorized)`

booxter:
recheck

PR #319, opened 2023-09-11
add support for libvirt_guests to edpm_libvirt

SeanMooney:
This change ports the libvirt_guests code from edpm_nova_libvirt
to edpm_libvirt and adapts it to follow the conventions of the
new role including molecule testing.

libvirt guests is a script that ensure that vms are gracefully shutdown
when the host is requested to reboot or poweroff.
This service is not strictly required however it can reduce the risk
of guest filesystem curruption. Wihtout this service systemd
will still attempt to terminate the qemu instance with a much
shorter timeout before sending sig_kill.

As there is little reason not to always use libvirt_guests
to protect the guest filesystem this change makes its use
mandataory and does not allow it to be disabled unlike
the previous opt in behavior.

As part of this refactoring the verify_systemd_service.yaml
test helper has been renamed to verify_systemd_unit.yaml and
extened to make asserting the enabled and runnign state
configurable. the edpm_nova role has also been adapted to
that change.

Closes: [OSPRH-156](https://issues.redhat.com//browse/OSPRH-156)


roles/edpm_libvirt/molecule/default/verify.yml
```diff
@@ -42,3 +42,9 @@
         - "libvirt_virtproxyd"
         - "libvirt_virtqemud"
         - "libvirt_virtsecretd"
+    - name: ensure  libvirt.target exist and are running
```
rebtoor, line 45: just a minor typo, a double space between "ensure" and "libvirt"

roles/edpm_libvirt/tasks/main.yml
```diff
@@ -25,3 +25,6 @@
 
 - name: post-install libvirt
   include_tasks: post-install.yml
```
rebtoor, line 27: `ansible.builtin.include_tasks`

roles/edpm_libvirt/tasks/main.yml
```diff
@@ -25,3 +25,6 @@
...
+
+- name: Flush post-install handlers
+  meta: flush_handlers
```
rebtoor, line 30: `ansible.builtin.meta`

rebtoor (changes requested):
it's fine for me, just fix a couple of minor changes.

rebtoor (approved):

rebtoor (approved):

PR #324, opened 2023-09-12
ovn: fix metadata agent configuration

booxter:
- Mount config dir into /etc/neutron.conf.d, allowing it to load all config files automatically through oslo.config default config dir mechanism. (This will be needed later when we add a config file with SB connection string to the directory.)

- Rely on implicit CLI arguments for the agent.

- Rename edpm_ovn_neutron_config_dir into edpm_ovn_neutron_metadata_agent_config_dir because every agent will have its own set of config files. (And there will be more agents in the base `ovn` role - e.g. the `neutron-ovn-agent` for qos/sriov.)


slagle (approved):

booxter:
recheck

slawqo (commented):
/lgtm

jpodivin (commented):
Isn't this going to clash with https://github.com/openstack-k8s-operators/edpm-ansible/pull/325 ?

roles/edpm_ovn/templates/kolla_ovn_metadata_agent.yaml.j2
```diff
@@ -1,4 +1,4 @@
-command: "/usr/bin/networking-ovn-metadata-agent --config-file {{ edpm_ovn_metadata_agent_config_dir }}/neutron.conf --config-file {{ edpm_ovn_metadata_agent_config_dir }}/plugins/networking-ovn/networking-ovn-metadata-agent.ini --log-file={{ edpm_ovn_metadata_agent_log_dir }}/ovn-metadata-agent.log"
+command: "networking-ovn-metadata-agent"
```
karelyatin, line 1: would be good to change the binary also to "neutron-ovn-metadata-agent" as it's provided by neutron sub rpms, may be as followup.

roles/edpm_ovn/meta/argument_specs.yml
```diff
@@ -249,8 +241,8 @@ argument_specs:
       edpm_ovn_metadata_common_volumes:
         default:
         - /lib/modules:/lib/modules:ro
-        - /run/openvswitch:/run/openvswitch:z
-        - '{{ edpm_ovn_neutron_config_dir }}/etc/neutron:/etc/neutron:z'
+        - /run/openvswitch:/usr/local/var/run/openvswitch:z
```
karelyatin, line 244: couldn't get how change to /usr/local/var/run/openvswitch is related here with this change.

karelyatin:
recheck infra

karelyatin (changes requested):

roles/edpm_ovn/meta/argument_specs.yml
booxter: I noticed that ovs library cannot connect to ovs, see where it connects in logs

```
2023-09-12 14:07:17.947 2 INFO ovsdbapp.backend.ovs_idl.vlog [-] unix:/usr/local/var/run/openvswitch/db.sock: connecting...                                                             
2023-09-12 14:07:17.948 2 INFO ovsdbapp.backend.ovs_idl.vlog [-] unix:/usr/local/var/run/openvswitch/db.sock: connected                                                                 
```

But yes, it's not directly related to the PR goal (which is primarily - load config-dir).

booxter: That said, I just noticed this:

```
2023-09-12 19:13:59.883 2 DEBUG neutron.agent.ovn.metadata_agent [-] ovn.vhost_sock_dir             = /var/run/openvswitch log_opt_values /usr/lib/python3.9/site-packages/oslo_config/cfg.py:2609
```

So I wonder if we should set up ovsdbapp - or the vhost-sock-dir setting? - to sync them?

booxter: AFAIU this comes from neutron/conf/agent/ovn/metadata/config.py

```
    cfg.StrOpt('ovsdb_connection',
               default='unix:/usr/local/var/run/openvswitch/db.sock',
               help=_('The connection string for the native OVSDB backend. '
                      'Use tcp:IP:PORT for TCP connections. '
                      'Use unix:FILE for unix domain socket connections.')),

```

booxter: Nevermind, I think the errors I saw was due to the final config file not loaded, which lists `ovsdb_connection = tcp:127.0.0.1:6640`. So this change is not needed. I will remove it.

roles/edpm_ovn/templates/kolla_ovn_metadata_agent.yaml.j2
```diff
@@ -1,4 +1,4 @@
-command: "/usr/bin/networking-ovn-metadata-agent --config-file {{ edpm_ovn_metadata_agent_config_dir }}/neutron.conf --config-file {{ edpm_ovn_metadata_agent_config_dir }}/plugins/networking-ovn/networking-ovn-metadata-agent.ini --log-file={{ edpm_ovn_metadata_agent_log_dir }}/ovn-metadata-agent.log"
```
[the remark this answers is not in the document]
booxter: We already collect log messages into stdouts/service.log files. Why do we need to also collect the same here? (Maybe there's a reason. What is it?)

Regardless, I will revert back for now until we have a chance to discuss it.

booxter:
In the latest version
- reverted back to passing log-file= argument for now (though it's unclear to me why we want to log everything twice - in stdouts and in /var/log/containers/neutron);
- switched to neutron-ovn-metadata-agent executable;
- added descriptions to args I touched;
- reverted back the change in mount point for /run/openvswitch - not needed now that conf.d is working;
- addressed other comments here and in the dependent PR that belong to this PR.

booxter:
In the latest version
- reverted back to passing log-file= argument for now (though it's unclear to me why we want to log everything twice - in stdouts and in /var/log/containers/neutron);
- switched to neutron-ovn-metadata-agent executable;
- added descriptions to args I touched;
- reverted back the change in mount point for /run/openvswitch - not needed now that conf.d is working;
- addressed other comments here and in the dependent PR that belong to this PR.

slagle:
/lgtm

looks like all feedback has been addressed

PR #328, opened 2023-09-13
NFV OvS DPDK Configuration Changes

Jaganathancse:
This change is used to have custom ansible edpm
variables for OvS DPDK configurations.

### DIFF
--- a/docs/source/roles/role-edpm_ovn_dpdk.rst
+++ b/docs/source/roles/role-edpm_ovn_dpdk.rst
@@ -1,0 +1,29 @@
+========================
+Role - edpm_ovs_dpdk
+========================
+
+Usage
+~~~~~
+
+This Ansible role allows to do the following tasks:
+
+* Configure the required OvS DPDK configuration
+   based on the OvS DPDK edpm ansible variables.
+
+* Remove any existing OvS DPDK configuration based
+  on the OvS DPDK edpm ansible variables.
+
+  Here is an example of a playbook:
+
+.. code-block:: YAML
+
+    - name: "Configure OvS DPDK Configs"
+      include_role:
+        name: "osp.edpm.edpm_ovs_dpdk"
+      vars:
+         edpm_ovs_dpdk_pmd_core_list: "1,13,3,15"
+         edpm_ovs_dpdk_lcore_list: "0,12"
+         edpm_ovs_dpdk_socket_memory: "4096"
+         edpm_ovs_dpdk_memory_channels: 4
+         edpm_ovs_dpdk_vhost_postcopy_support: tru
+         edpm_nova_libvirt_vhostuser_socket_group: 'hugetlbfs'
--- a/playbooks/configure_ovs_dpdk.yml
+++ b/playbooks/configure_ovs_dpdk.yml
@@ -1,0 +1,12 @@
+---
+
+- name: Configure OvS DPDK
+  hosts: all
+  strategy: linear
+  become: true
+  tasks:
+    - name: Configure OvS DPDK configs
+      ansible.builtin.import_role:
+        name: osp.edpm.edpm_ovs_dpdk
+      tags:
+        - edpm_ovs_dpdk
--- a/plugins/filter/cpu_mask.yml
+++ b/plugins/filter/cpu_mask.yml
@@ -1,0 +1,16 @@
+---
+
+DOCUMENTATION:
+  name: cpu_mask
+  author: "EDPM team"
+  version_added: 2.9
+  short_description: Calculates cpu mask for the list of CPus.
+  description: |
+    This filter will take the list of CPUs and will convert
+    that to a cpu mask value.
+EXAMPLES: |
+  "{{ cpu_list | osp.edpm.cpu_mask }}"
+RETURN:
+  _value:
+    description:
+    type: string
--- a/plugins/filter/helpers.py
+++ b/plugins/filter/helpers.py
@@ -15,6 +15,7 @@
 # under the License.
 
 import ast
+import binascii
 import json
 import re
 
@@ -25,7 +26,8 @@ class FilterModule:
             'needs_delete': self.needs_delete,
             'haskey': self.haskey,
             'dict_to_list': self.dict_to_list,
-            'jump_chain_targets': self.jump_chain_targets
+            'jump_chain_targets': self.jump_chain_targets,
+            'cpu_mask': self.cpu_mask
         }
 
     def needs_delete(self, container_infos, config, config_id,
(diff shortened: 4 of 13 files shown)


roles/edpm_ovn_dpdk/meta/argument_specs.yml
```diff
@@ -0,0 +1,66 @@
...
+        description: OVN DPDK TSO enable or not
+      edpm_ovn_dpdk_pmd_auto_lb:
+        type: bool
+        default: false
+        description: OVN DPDK PMD auto load balance enable or not
+      edpm_ovn_dpdk_pmd_load_threshold:
+        type: str
+        default: ""
+        description: OVN DPDK PMD load threshold
+      edpm_ovn_dpdk_pmd_improvement_threshold:
```
slawqo, line 51: podified-multinode-edpm-deployment-crc job is reporting issue somewhere here. See https://logserver.rdoproject.org/28/328/3d51725ccb2b3cfcc85f0e37d773d8552e69d7b9/github-check/podified-multinode-edpm-deployment-crc/19348b1/controller/ci-framework-data/artifacts/must-gather.local.1599254522418717781/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-070a98749455b097c66a1a25ff827a3c0ee7385f1f0c8bf54f5b15438501f2bd/namespaces/openstack/pods/dataplane-deployment-configure-network-openstack-edpm-84q24/logs/dataplane-deployment-configure-network-openstack-edpm-84q24.log

rebtoor:
recheck

Jaganathancse:
> @Jaganathancse the jobs are failing on this task:
> 
> ```
> 2023-09-18T10:51:29.730392083+00:00 stdout F TASK [osp.edpm.edpm_ovn_dpdk : Check valid input for edpm_ovn_dpdk_pmd_core_list] ***
> 2023-09-18T10:51:29.730843763+00:00 stdout F Monday 18 September 2023  10:51:29 +0000 (0:00:00.075)       0:00:31.974 ****** 
> 2023-09-18T10:51:29.844025305+00:00 stdout F �[0;31mfatal: [edpm-compute-0]: FAILED! => {"changed": false, "msg": "List of PMD cores cannot be empty - edpm_ovn_dpdk_pmd_core_list"}�[0m
> ```
> 
> https://logserver.rdoproject.org/28/328/f74f9b83e263b5f0a7adca3541a53f9759bbcb9d/github-check/podified-multinode-edpm-deployment-crc/67d862a/controller/ci-framework-data/logs/crc/pods/openstack_dataplane-deployment-configure-network-openstack-edpm-bv7qc_fd63926c-cfd9-40ad-8291-3f14f7a0db85/openstackansibleee/0.log
@rebtoor Ack, Thanks Will fix it.

plugins/modules/openvswitch_db.py
rebtoor, line 1: I didn't review this PR yet, but this 3rd party module (which it seems also quite old) worries me a bit.

@slagle @bshephar @rabi @jpodivin any thoughts on including a 3rd party module in the collection?

jpodivin: Well I'm not 100% how Apache-2.0 license meshes with GPLv3.0. I believe it should be fine. But I'm not a lawyer.

Apart from that there is the stated dependency on `ovs-vsctl`, which I don't see being imported or included in our dependencies. We need to determine if it's really necessary and if how we can ensure it's present. 

There is a minor issue of the argument specification, we are usually parsing the `DOCUMENTATION` var to get them rather than hard coding them. But that should be easy to fix.

jpodivin (changes requested):

jpodivin (changes requested):
Molecule tests have to be run either in a github action or in as a RDO zuul job.

Jaganathancse:
> Molecule tests have to be run either in a github action or in as a RDO zuul job.

Ack, will work on molecule tests as separate patch once this changes merged.

Jaganathancse:
getting timeout error and looks like error is not due to this changes 
'cp: cannot stat '/home/zuul/*.log': No such file or directory'

Jaganathancse:
recheck

plugins/modules/openvswitch_db.py
slagle, line 1: It looks like this was just copied from https://github.com/ansible-collections/openvswitch.openvswitch/blob/main/plugins/modules/openvswitch_db.py
That's not really a sustainable path forward. Who intends to maintain this and keep it in sync with the upstream?

slagle (changes requested):

plugins/modules/openvswitch_db.py
vcandapp: @slagle , the file openvswitch_db.py is not from the ainsible-collection repo as you have mentioned. It is from our earlier releases: https://review.opendev.org/c/openstack/tripleo-ansible/+/677712 (@krsacme).
NFV team as been maintaing all these time
@Jaganathancse @karthiksundaravel

plugins/modules/openvswitch_db.py
```diff
@@ -0,0 +1,238 @@
+#!/usr/bin/python
+# coding: utf-8 -*-
+
+#
+# (c) 2015, Mark Hamilton <mhamilton@vmware.com>
+# Portions copyright @ 2015 VMware, Inc.
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
booxter, line 7: This repo is Apache licensed. How does it play with mixing in GPL code, 3.0 at that?

plugins/modules/openvswitch_db.py
slagle: And before that it came from the ansible collection. That's an issue we need to be aware of. Do we really need the whole module to enable dpdk? Can it be done a different way so that we need to copy it here?

plugins/modules/openvswitch_db.py
jpodivin: This is what Apache has to say on the matter[0]:

> Apache 2 software can therefore be included in GPLv3 projects, because the GPLv3 license accepts our software into GPLv3 works. However, GPLv3 software cannot be included in Apache projects. The licenses are incompatible in one direction only, and it is a result of ASF's licensing philosophy and the GPLv3 authors' interpretation of copyright law.

Seems like this would be one of the cases when it's not possible.

[0] https://www.apache.org/licenses/GPL-compatibility.html

Jaganathancse:
> Molecule tests have to be run either in a github action or in as a RDO zuul job.

Will do a seperate patch for molecule tests once this patch looks fine and merged.

Jaganathancse:
recheck

PR #330, opened 2023-09-14
Add `.vscode` to `.gitignore`

rebtoor:


rebtoor:
github-check is useless since only .gitignore is changed.

fao89 (approved):

PR #332, opened 2023-09-14
Role edpm_ceph_hci_pre needs become to update firewall

fultonj:
Add `become: true` to tasks in firewall.yml from the role edpm_ceph_hci_pre since the move to using cloud-admin [1] exposed that we were relying on the role being run as root. When run as cloud-admin without become these tasks failed: [Errno 13] Permission denied: b'/var/lib/edpm-config/firewall'

[1] https://github.com/openstack-k8s-operators/install_yamls/pull/531


fultonj:
recheck

PR #333, opened 2023-09-14
ovn: remove rootwrap.d from the list of filters_path

booxter:
The directory doesn't exist, and we move the filter files under /usr/share/neutron/rootwrap in rpms. See:

https://github.com/rdo-packages/neutron-distgit/blob/antelope-rdo/openstack-neutron.spec#L499

### DIFF
--- a/roles/edpm_neutron_dhcp/defaults/main.yml
+++ b/roles/edpm_neutron_dhcp/defaults/main.yml
@@ -60,7 +60,7 @@ edpm_neutron_dhcp_oslo_middleware_enable_proxy_headers_parsing: 60
 
 # rootwrap.conf
 # DEFAULT
-edpm_neutron_dhcp_rootwrap_DEFAULT_filters_path: '/etc/neutron/rootwrap.d,/usr/share/neutron/rootwrap'
+edpm_neutron_dhcp_rootwrap_DEFAULT_filters_path: '/usr/share/neutron/rootwrap'
 edpm_neutron_dhcp_rootwrap_DEFAULT_exec_dirs: '/sbin,/usr/sbin,/bin,/usr/bin,/usr/local/bin,/usr/local/sbin,/etc/neutron/kill_scripts'
 edpm_neutron_dhcp_rootwrap_DEFAULT_use_syslog: false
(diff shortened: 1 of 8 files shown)


booxter:
recheck

roles/edpm_ovn/defaults/main.yml
```diff
@@ -101,7 +101,7 @@ edpm_ovn_metadata_agent_oslo_concurrency_lock_patch: '$state_path/lock'
 edpm_ovn_metadata_agent_agent_report_interval: '300'
 
 # rootwrap.conf
-edpm_ovn_metadata_agent_rootwrap_DEFAULT_filters_path: '/etc/neutron/rootwrap.d,/usr/share/neutron/rootwrap'
```
karelyatin, line 104: Seems the use case was operator/users/deployment tools can put files at /etc/neutron/rootwrap.d, while rpm ships defaults at /usr/share/neutron/rootwrap? I see other services(like glance,cinder,nova etc) also following this pattern. So would be good to check more before dropping.

fao89:
@karelyatin @booxter I think you can add a group for you here: https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/OWNERS_ALIASES  and  https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/OWNERS

roles/edpm_ovn/defaults/main.yml
booxter: Let's have a holistic discussion about how we expect 3rd parties to extend EDP, if at all. I will mark the PR draft for now.

booxter: `etc/neutron/rootwrap.d` comes from upstream repository, and it looks like it is a historical remnant from the time when everything was kept in-tree, including 3rd party plugins, that could install its own filters in global directory with other filters.

This is no longer applicable here. First, because each agent runs with its own file system. Second, because rootwrap filter ad-hoc extension is not encouraged. And if a partner would like to ship their customizations to any of the agents, they will do it as a separate agent image (incl. runner image with a customized edpm-ansible).

I'm still leaving the variable available for override just in case we ever want to, but reducing the default list to just the one directory that actually contains filters.

slawqo (approved):

PR #335, opened 2023-09-14
Add firewall rules for ceph-nfs

fmount:
When `ceph-nfs` is deployed as part of `Ceph` with an associated `ingress` daemon, we should provide firewall rules to avoid unexpected behavior when the cluster is created and clients try to reach the `VIP` endpoint.
Required for https://github.com/openstack-k8s-operators/ci-framework/pull/557


fultonj (approved):
+2

fmount:
recheck

gouthampacha (approved):
ty @fmount

rebtoor (approved):

rebtoor:
recheck

PR #345, opened 2023-09-19
Enable `download-cache` support in `edpm_iscsid` role

rebtoor:

PR #352, opened 2023-09-20
Revert "Enable `download-cache` support in `edpm_iscsid` role"

rebtoor:
Reverts openstack-k8s-operators/edpm-ansible#345

despite tests are green on ci, podman isn't installed by default on a "fresh" EL9 nodes and the `edpm_podman` role is executed later (in both install_os and configure_os playbooks) so local installs are now broken.


gibizer (approved):

rabi (approved):
/lgtm

PR #358, opened 2023-09-21
Enable `download-cache` support in `edpm_libvirt` role

rebtoor:


fao89 (approved):

fao89 (approved):

rabi:
Phew, this has broken the baremetal job as there is not enough space to download the images. We removed the baremetal job and then merged this. That's why I'm not for removing that job. We should not be running download-cache before running growvol. https://github.com/openstack-k8s-operators/dataplane-operator/pull/429

rebtoor:
Thx for finding this @rabi ! 

BTW i don't think that download-cache should be executed after config-network and validate-network playbooks. I'd rather call growvol role inside download-cache playbook, just like we did for the podman role.

The download-cache playbook is meant to be run before all other playbooks to let users the to download everything before the setup start, otherwise is almost useless.

WDYT @slagle ?

rabi:
> BTW i don't think that download-cache should be executed after config-network and validate-network playbooks

I don't see how these container images would be required in configure-network/validate-network services. Don't see any point downloading them unnecessarily.

PR #360, opened 2023-09-21
SR-IOV is not configured when the provider is not set

Jaganathancse:
The default provider still remains as 'ifcfg' and when its not specified, the SR-IOV PF configuration is skipped. Adding the changes to support the same.

### DIFF
--- a/plugins/modules/edpm_os_net_config.py
+++ b/plugins/modules/edpm_os_net_config.py
@@ -115,6 +115,8 @@ def _run_os_net_config(config_file, cleanup=False, debug=False,
         argv.append('--noop')
     if use_nmstate:
         argv.append('--provider nmstate')
+    else:
+        argv.append('--provider ifcfg')
     cmd = " ".join(argv)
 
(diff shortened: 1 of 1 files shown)


karthiksundaravel (commented):
LGTM ! Thanks

rebtoor:
recheck

plugins/modules/edpm_os_net_config.py
```diff
@@ -115,6 +115,8 @@ def _run_os_net_config(config_file, cleanup=False, debug=False,
         argv.append('--noop')
     if use_nmstate:
         argv.append('--provider nmstate')
+    else
```
rebtoor, line 118: there's a missing colon

rebtoor (changes requested):

rebtoor (approved):

rebtoor:
recheck

there was a 401 error on the cluster (dataplane stuff has been executed so it should not be caused by this PR)

PR #364, opened 2023-09-22
Do not attempt to manage collocated Ceph containers

fultonj:
If an HCI EDPM node has Ceph containers, then they should not be be manged by the edpm_container_manage role. Add a new task to exclude Ceph containers from the list of containers built by the call to
containers.podman.podman_container_info.

Without this new task, the edpm_container_manage role delete_orphan.yml tasks file attempts to process Ceph containers but it cannot handle them so it fails with an AttributeError: 'NoneType' object has no attribute 'get'.

### DIFF
--- a/roles/edpm_container_manage/tasks/delete_orphan.yml
+++ b/roles/edpm_container_manage/tasks/delete_orphan.yml
@@ -20,6 +20,13 @@
   when:
     - edpm_container_manage_cli == 'podman'
 
+- name: Exclude Ceph containers from podman container list
+  ansible.builtin.set_fact:
+    filtered_containers: "{{ filtered_containers | default([]) + [item] }}"
+  when:
+    - not (item['Name'] | regex_search('^ceph.*'))
+  loop: "{{ podman_containers.containers }}"
+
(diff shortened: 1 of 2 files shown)


roles/edpm_container_manage/tasks/delete_orphan.yml
```diff
@@ -20,6 +20,13 @@
   when:
     - edpm_container_manage_cli == 'podman'
 
+- name: Exclude Ceph containers from podman container list
```
jpodivin, line 23: On one hand I'm wondering if this could be done as filter in the followup task. Otoh this is way more readable.

jpodivin (changes requested):
Great work. The only addition I would recommend is a note in role documentation, explicitly stating that this role skips Ceph containers. Otherwise people may use it for that and wonder why nothing is happening.

fultonj:
> Great work. The only addition I would recommend is a note in role documentation, explicitly stating that this role skips Ceph containers. Otherwise people may use it for that and wonder why nothing is happening.

Done.

abays (commented):
+1

This change works for me in my CephHCI environment

jpodivin (approved):

PR #369, opened 2023-09-25
Refactored `edpm_download_cache` role and added call to `edpm_growvols` role as requirement

rebtoor:
* Added `requirements.yml` tasks file for role dependencies
* Added `edpm_download_role_requirements` boolean variable to enable/disable role dependencies installation
* Called `edpm_growvols` role in dependencies to avoid free disk space issues while downloading packages and container images


roles/edpm_download_cache/tasks/requirements.yml
slagle, line 1: requirements.yml feels like an outlier compared to the other role task files. What about calling this install.yml?

roles/edpm_download_cache/tasks/container_images.yml
```diff
@@ -1,21 +1,5 @@
...
-    - edpm_podman
-    - download_cache
-
-- name: Login to container registries if needed
-  ansible.builtin.include_role:
-    name: osp.edpm.edpm_podman
-    tasks_from: login.yml
-  tags:
-    - edpm_podman
-    - download_cache
```
slagle, line 17: I feel like we may still want the login task here. Would the login credentials ever expire?

roles/edpm_download_cache/tasks/requirements.yml
rebtoor: Ok, fair enough.

roles/edpm_download_cache/tasks/container_images.yml
rebtoor: I hadn't tought about it, but it could be a problem, since we are going to perform the installation of the rpms between the login task and the download of the containers. Let me move this task back in this file.

slagle (approved):

PR #370, opened 2023-09-26
Create systemd files requires root

jlarriba:


rebtoor (changes requested):
I'll approve the change, but I'll propose a followup PR for this role, there are few things that should be fixed (usage of fqcn, best practices and so on)

rebtoor (approved):

PR #371, opened 2023-09-26
Enable download-cache support in edpm_neutron_sriov role

slawqo:


roles/edpm_download_cache/tasks/container_images.yml
```diff
@@ -55,3 +55,11 @@
   tags:
     - edpm_nova
     - download_cache
+
+- name: Download images for edpm_neutron_sriov role
+  ansible.builtin.include_role:
+    name: osp.edpm.edpm_neutron_sriov
+    tasks_from: download_cache.yml
```
karelyatin, line 62: seems you also need to add tasks/download_cache.yml in the edpm_neutron_sriov role

karelyatin (changes requested):

rebtoor (approved):

PR #373, opened 2023-09-27
Replacing edpm_chrony with linux-systemroles.timesync

jpodivin:


playbooks/install_os.yml
```diff
@@ -4,6 +4,21 @@
   hosts: all
   strategy: linear
   become: true
+  vars:
```
fultonj, line 7: Is the plan to document that users of the playbook should override `timesync_ntp_servers` as well as have install_yamls generate EDPM ansible vars which can override these values?

playbooks/install_os.yml
```diff
@@ -17,10 +32,9 @@
         tasks_from: install.yml
       tags:
         - edpm_sshd
-    - name: Install edpm_chrony
+    - name: Install and configure edpm_chrony
       ansible.builtin.import_role:
-        name: osp.edpm.edpm_chrony
-        tasks_from: install.yml
+        name: fedora.linux_system_roles.timesync
```
fultonj, line 37: Do you know if the downstream FCQN will be the same? 

I have no idea, I just didn't want to have to also have a version of this with something like `rhel.linux_system_roles.timesync`. 

Also, what installs this module usptream and downstream?

rebtoor: nope, they are a bit different. I already faced this problem calling `network` role in `edpm_network_config`. I build the collection FQCN based on distro -> https://github.com/openstack-k8s-operators/edpm-ansible/blob/26fe00b33b84a05ee7107a704d9dcd2bc1692612/roles/edpm_network_config/tasks/main.yml#L22-L26

fultonj: OK, I guess this patch should borrow from @rebtoor's example

rebtoor: > Also, what installs this module usptream and downstream?

https://access.redhat.com/articles/3050101

TL;DR

`dnf -y install rhel-system-roles`

jpodivin: In upstream it's ansible galaxy, in downstream it depends. Ansible doesn't support downstream collection releases, or any other delivery mechanism than galaxy. Unless we are building package somewhere, this is the way it will be in downstream.

Technically it's not such a problem, we are sending this over as an image with dependencies installed.

fultonj: I just wanted to know who that image will be built for either case. Thanks to @rebtoor for the answer

playbooks/install_os.yml
jpodivin: Overrides in install_yamls only. These are reasonable defaults and I don't think we should have users overriding them unless absolutely necessary.

playbooks/install_os.yml
```diff
@@ -17,10 +32,17 @@
         tasks_from: install.yml
       tags:
         - edpm_sshd
-    - name: Install edpm_chrony
-      ansible.builtin.import_role:
-        name: osp.edpm.edpm_chrony
-        tasks_from: install.yml
+    - name: Set FQN for the timesync role
```
rebtoor, line 35: FQCN?

playbooks/install_os.yml
fultonj: I agree these are reasonable defaults. I know for sure though there will be Red Hatters and customers not using install_yamls who want to override them.

rebtoor: Nope, I disagree. I'm putting my sysadmin hat here. :D

The vast majority of our customers environments have internal NTP servers (and most of the time servers can't even connect to internet) so I'm 100% sure they need to customize them.

I'd leave the `timesync_ntp_servers` var here as "sane default" (and self-explaining example) but we need to document how to override it, even linking the role documentation.

rebtoor (changes requested):
As stated by @fultonj , you need to "build

playbooks/install_os.yml
jpodivin: Well, we don't really have documentation for individual playbooks yet. I suppose it's time to start one. In the transitionary period, before install_yamls is patched, we'll need to support three way resolution for defaults, at least for the ntp servers list.

jpodivin:
I've updated the docs with some notes on the playbook. The defaults situation is not exactly clean, but it's unavoidable until the install_yamls is patched to use new var names.

fultonj (commented):
This is coming along nicely. I think we need a 3-way depends-on.

https://github.com/openstack-k8s-operators/dataplane-operator/pull/463
depends-on
https://github.com/openstack-k8s-operators/edpm-ansible/pull/373
depends-on
"new install_yamls patch to genereate OpenStackDataPlaneNodeSets with timesync_ntp_servers and respect DATAPLANE_CHRONY_NTP_SERVER"

That final patch would need to touch areas affected by this older patch:

https://github.com/openstack-k8s-operators/install_yamls/commit/c0f7f44e7ca981662b8c583077608b91a88dfad1

playbooks/install_os.yml
```diff
@@ -17,10 +28,25 @@
...
+      delegate_to: localhost
+    - name: Compile list of ntp servers with their attributes
+      ansible.builtin.set_fact:
+        timesync_ntp_servers: "{{ timesync_ntp_servers | default([]) + [{ 'hostname' : item }]}}"
+      loop: "{{ timesync_ntp_servers }}"
+      when: item is string
+      failed_when: >
+        ( item is hash ) and
+        ( 'hostname' not in item )
+    - name: Install and configure edpm_chrony
```
slagle, line 47: Does edpm_chrony need to be changed here?

jpodivin: Eh. I'm not really sure what should we call that step. Yes, it isn't the `edpm_chrony` role anymore. OTOH, the service is still for edpm and it's still chrony being deployed ... How about "Install and configure chrony"?

rebtoor: What about "Install and configure chrony via linux-system-roles" ?

It could be helpful to declare that we are going to use a role from another collection. WDYT?

jpodivin: I don't know about that, after all if they read the source they can see what we are using, and if something goes wrong they will see the role name in logs.

lewisdenny:
recheck
Due to a CI Blocker

playbooks/install_os.yml
jpodivin: I've changed the name.

jpodivin:
recheck - error: unable to default to a user name: the server is currently unable to handle the request (get users.user.openshift.io ~)

jpodivin:
CI is passing now so I'm opening this for a proper review.
Handful of note:

- Dependent PRs in dataplane-operator and install-yamls will be merged after this one.
- Some of the code is only intended as a temporary bridging solution, until aforementioned PRs are merged.
- Once the dependent PRs are merged, a new, capping, PR will be sent against this repo, removing the newly superfluous logic.

playbooks/install_os.yml
```diff
@@ -17,10 +28,25 @@
...
+      delegate_to: localhost
+    - name: Compile list of ntp servers with their attributes
+      ansible.builtin.set_fact:
+        timesync_ntp_servers: "{{ timesync_ntp_servers | default([]) + [{ 'hostname' : item }]}}"
+      loop: "{{ timesync_ntp_servers }}"
+      when: item is string
+      failed_when: >
+        ( item is mapping ) and
+        ( 'hostname' not in item )
+    - name: Install and configure time service using timesync system role
```
fultonj, line 47: So you're building this list in lines 40-46 so that the patches which depend on this one can keep passing a list of NTP servers instead of passing a dictionary which maps hostname to NTP server as pointed out here.

  https://github.com/openstack-k8s-operators/dataplane-operator/pull/463/files#r1355074751
  https://github.com/openstack-k8s-operators/install_yamls/pull/575/files#r1356693584

fultonj (commented):
1. What is order you want these patches merged in?

- https://github.com/openstack-k8s-operators/install_yamls/pull/575
- https://github.com/openstack-k8s-operators/dataplane-operator/pull/463
- https://github.com/openstack-k8s-operators/edpm-ansible/pull/373

I see you have the following depends-on relationships:
```
https://github.com/openstack-k8s-operators/dataplane-operator/pull/463
  https://github.com/openstack-k8s-operators/edpm-ansible/pull/373

https://github.com/openstack-k8s-operators/install_yamls/pull/575
  https://github.com/openstack-k8s-operators/edpm-ansible/pull/373
```
Shouldn't dataplane-operator go before install_yamls? If so shouldn't the depends-on order be like this?
```
https://github.com/openstack-k8s-operators/install_yamls/pull/575
  https://github.com/openstack-k8s-operators/dataplane-operator/pull/463
    https://github.com/openstack-k8s-operators/edpm-ansible/pull/373
```
Then when the above is done would you patch edpm-ansible to remove the old tripleo chrony role?

I mentioned this before, why change the name of `edpm_ntp_servers` but have it keep the same structure as `edpm_chrony_ntp_servers`?

playbooks/install_os.yml
```diff
@@ -4,6 +4,17 @@
   hosts: all
   strategy: linear
   become: true
+  vars:
+    default_timesync_ntp_servers:
+      - '0.pool.ntp.org'
+      - '1.pool.ntp.org'
+      - '2.pool.ntp.org'
+      - '3.pool.ntp.org'
```
slagle, line 12: I don't think EDPM should maintain a list of default NTP servers. We don't want to repeat what we did with tripleo-heat-templates, where we had to maintain defaults for components on our own. In this case, the timesync role does not maintain defaults, so EDPM shouldn't either. Instead, we can rely on the defaults set by the distro.

slagle (changes requested):
I'm also not following how this will eventually evolve to be used.

I would instead add the new functionality to use the timesync system role to edpm-ansible, and guard that with a "when" on if timesync_ntp_servers is defined and set.

Then update dataplane-operator and install_yamls to use the system role by defining timesync_ntp_servers. I don't see a reason for edpm_ntp_servers, as we don't want that abstraction over the system role variables. Just use the system role variables directly.

Then followup with an edpm-ansible PR to remove the chrony role, both the role iteself and from the playbooks. You could also drop the "when" at that point guarding the usage of the timesync role, assuming that role does the right thing when timesync_ntp_servers is not defined.

playbooks/install_os.yml
jpodivin: Pretty much yes. It does minimize the number of things that can break. The existing usage stays almost the same, all important changes will be contained here.

playbooks/install_os.yml
jpodivin: Ok. That does make sense.

jpodivin:
> 1. What is order you want these patches merged in?
> 
> 
>     * [Replacing edpm_chrony_ntp_servers with edpm_ntp_servers install_yamls#575](https://github.com/openstack-k8s-operators/install_yamls/pull/575)
> 
>     * [Replacing edpm_chrony_ntp_servers with edpm_ntp_servers dataplane-operator#463](https://github.com/openstack-k8s-operators/dataplane-operator/pull/463)
> 
>     * [Replacing edpm_chrony with linux-systemroles.timesync #373](https://github.com/openstack-k8s-operators/edpm-ansible/pull/373)
> 
> 
> I see you have the following depends-on relationships:
> 
> ```
> https://github.com/openstack-k8s-operators/dataplane-operator/pull/463
>   https://github.com/openstack-k8s-operators/edpm-ansible/pull/373
> 
> https://github.com/openstack-k8s-operators/install_yamls/pull/575
>   https://github.com/openstack-k8s-operators/edpm-ansible/pull/373
> ```
> 
> Shouldn't dataplane-operator go before install_yamls? If so shouldn't the depends-on order be like this?
> 
> ```
> https://github.com/openstack-k8s-operators/install_yamls/pull/575
>   https://github.com/openstack-k8s-operators/dataplane-operator/pull/463
>     https://github.com/openstack-k8s-operators/edpm-ansible/pull/373
> ```
> 
> Then when the above is done would you patch edpm-ansible to remove the old tripleo chrony role?
> 
> I mentioned this before, why change the name of `edpm_ntp_servers` but have it keep the same structure as `edpm_chrony_ntp_servers`?

The dependent patches should be fine to merge in any order. At least it appears to be the case from the CI. There may be a short period of time when something doesn't quite work. But that would be the case anyway when the backwards compatibility is not an issue. The variable now actually holds a dictionary.

playbooks/install_os.yml
jpodivin: List compilation and checks removed.

fultonj:
The changes that I have requested have been addressed, so thanks for that. 

As per @slagle's comments:

https://github.com/openstack-k8s-operators/edpm-ansible/pull/373#pullrequestreview-1694636108

You've added `timesync_ntp_servers` to the two patches which depend on this one (and that variable's structure looks good). However, he also suggested you "guard that with a `when` on if `timesync_ntp_servers` is defined and set". That `when` is not here. Also, this patch could be split into two so that you don't remove the calling of the old role until later.

Would you please reply to these comments with an ack or reasons for a nack instead of simply requesting re-review?

jpodivin:
> I'm also not following how this will eventually evolve to be used.
> 
> I would instead add the new functionality to use the timesync system role to edpm-ansible, and guard that with a "when" on if timesync_ntp_servers is defined and set.
> 
> Then update dataplane-operator and install_yamls to use the system role by defining timesync_ntp_servers. I don't see a reason for edpm_ntp_servers, as we don't want that abstraction over the system role variables. Just use the system role variables directly.
> 
> Then followup with an edpm-ansible PR to remove the chrony role, both the role iteself and from the playbooks. You could also drop the "when" at that point guarding the usage of the timesync role, assuming that role does the right thing when timesync_ntp_servers is not defined.

Original proposal was backwards compatible, in that the incoming variable was checked and turned into format the timesync role expected, no matter what was used with the playbook. This one is not backwards compatible. But it has following advantages:

1. Minimizes number of patches we need to merge.
2. Keeps the code clean.

Wrapping original role invocation in conditional is possible, but imho inferior approach. The conditional would have to be implemented not just in one, but in all three playbooks using the `edpm_chrony` role. And then subsequently removed. 

This way we would just need to remove `edpm_chrony` which would no longer be used anywhere at that point. 

If you want backwards compatibility, with no chance of failure, I can restore the original argument checking task, with some adjustment to names of variables.

playbooks/install_os.yml
```diff
@@ -17,12 +17,29 @@
...
+              "{%- if ansible_facts['distribution'] == 'RedHat' -%}
+                redhat.rhel{%- else -%}
+                fedora.linux{%- endif -%}
+              _system_roles.timesync"
+          delegate_to: localhost
+        - name: Install and configure time service using timesync system role
+          ansible.builtin.include_role:
+            name: "{{timesyncfqcn}}"
+          tags:
+            - edpm_chrony
```
rebtoor, line 35: I'd remove `edpm_chrony` tag with something like `timesync` or `chrony`. I don't think that `edpm_` prefix is mandatory for the tags.

rebtoor (changes requested):
I left a single comment for the code, just a minor issue on tag.

BTW, I'm fine with this approach, as long as it followed pretty soon by a followup PR which removes the edpm_chrony role and its invocation within the playbooks. 

Keeping 2 ways to configure ntp it could only generate noise during the CRs generation.

playbooks/install_os.yml
jpodivin: `dataplane_chrony` @slagle @fultonj wdyt?

fultonj: I'm fine with the tag being called `edpm_chrony` since it involes calling the edpm_chrony role. I expect the calls to that role and the tag will be removed soon so I don't think this needs further revision on that level.

jpodivin: Ok, then `edpm_chrony` when it calls the `edpm_chrony`, otherwise the new tag.

rebtoor (approved):

PR #376, opened 2023-09-27
Set correct default for SRIOV agent report_interval

slawqo:
In Neutron default value for the 'report_interval' option is set to 30 seconds and this works well for the typical deployments, together with the 'agent_down_time' option set on the neutron server side to 75 seconds by default.

In the neutron_sriov role default for the report_interval was previously set to 300 seconds which was causing agent flapping UP/DOWN.

This patch sets default value for this config option to 30 seconds which is the same as is used in Neutron.


karelyatin (commented):
/lgtm

I see it's similarly set for ovn metadata agent, is that config used and require reset too?

PR #381, opened 2023-09-28
Upgrade collection requirements during ansibleee runner build

rebtoor:
During the build of the AnsibleEE runner image, if the collections are already installed (as it happens on github runners) they are never upgraded.

If a specific version of a collection is needed, it should be pinned in the `requirements.yml` file in the root of the repository.


slagle (approved):

rebtoor:
Standard build log: https://github.com/openstack-k8s-operators/edpm-ansible/actions/runs/6342331627/job/17227832268
Updated build log (on my branch): https://github.com/rebtoor/edpm-ansible/actions/runs/6343436726/job/17231305130

PR #386, opened 2023-10-03
Speed up Ceph exclude task

fao89:
Context: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/171#issuecomment-1745184297


fultonj (approved):
I tested this in my HCI environment and it does the functional equivalent of what the old code did but faster.

It doesn't break it and speeds it up. Thank you @fao89 !

PR #387, opened 2023-10-03
Revert "Do not attempt to manage collocated Ceph containers"

fultonj:
This reverts commit 0e6058cc47c5f147a2338b41f07d6f8dd8bac5de.

This breaks HCI but HCI is not yet in CI so a follow up patch will fix what this patch breaks for people deploying HCI.

The commit being reverted introduced a task that we still the functionality of, however, its implementation was not efficient and takes too long to run in the CI as documented:

https://github.com/openstack-k8s-operators/data-plane-adoption/pull/171

We still need to exclude the ceph containers but will need to do it with less computational resources (e.g. maybe pure jinja will be faster or we might need a custom filter).

Conflicts:
	roles/edpm_container_manage/tasks/delete_orphan.yml

### DIFF
--- a/docs/source/roles/role-edpm_container_manage.rst
+++ b/docs/source/roles/role-edpm_container_manage.rst
@@ -29,9 +29,6 @@ This Ansible role allows to do the following tasks:
 
   * The container config_id doesn't match with the one in input.
 
-  * The container's name starts with "ceph". Ceph containers are not
-    managed by this role.
-
   Once the previous conditions checked, then these reasons will make the
   containers deleted:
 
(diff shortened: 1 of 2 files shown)


fultonj:
I put this on hold so we can decide on priority.

fultonj:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/386 looks promising. I'll test and review that  patch

fultonj:
No revert needed as @fao89's other patch fixes the problem https://github.com/openstack-k8s-operators/edpm-ansible/pull/386

PR #389, opened 2023-10-04
Ensure logrotate image is also pre-downloaded

luis5tb:


jpodivin (approved):

luis5tb:
still testing this as it is not working as expected

luis5tb:
now it is working

rebtoor (changes requested):
edpm_logrotate_crond is already in the download_cache: https://github.com/openstack-k8s-operators/edpm-ansible/pull/338

luis5tb:
> edpm_logrotate_crond is already in the download_cache: #338

but that is only for the packages, not for the container image

rebtoor:
> > edpm_logrotate_crond is already in the download_cache: #338
> 
> but that is only for the packages, not for the container image

Oh, sorry, you're right. I'll leave you a review just to change a bit in the download task. (I saw that you already solved the merge issue, thanks!)

roles/edpm_download_cache/tasks/container_images.yml
```diff
@@ -76,3 +76,11 @@
   ansible.builtin.include_role:
     name: osp.edpm.edpm_neutron_dhcp
     tasks_from: download_cache.yml
+
+- name: Download images for edpm_logrotate_crond role
```
rebtoor, line 80: We already call download_cache.yml during the download of the packages so this task, as it stands, would be redundant. Do you mind splitting the download_cache in two different task files so we can call them independently?

roles/edpm_logrotate_crond/tasks/download_cache.yml
```diff
@@ -9,3 +9,14 @@
   until: edpm_logrotate_crond_cronie_package_download.rc == 0
   retries: "{{ edpm_logrotate_crond_download_retries }}"
   delay: "{{ edpm_logrotate_crond_download_delay }}"
+
+- name: Download needed container images
+  containers.podman.podman_image:
+    name: "{{ item }}"
+  loop:
```
rebtoor, line 16: I think we can avoid to use a loop here since a single image is downloaded.

rebtoor (changes requested):

rebtoor (approved):

PR #394, opened 2023-10-05
Sidecar containers in neutron agents

slawqo:
Add support for sidecar containers in neutron_dhcp and neutron_ovn roles


roles/edpm_neutron_dhcp/molecule/default/verify.yml
```diff
@@ -51,3 +51,31 @@
...
+          become: true
+          ansible.builtin.stat:
+            path: "/var/lib/neutron/dhcp_agent_haproxy_wrapper"
+          register: haproxy_wrapper
+
+        - name: assert that haproxy wrapper exists
+          ansible.builtin.assert:
+            that:
+              - haproxy_wrapper.stat.exists
+            fail_msg: "haproxy wrapper script does not exist"
```
jpodivin, line 67: I would have gone with fail, but this is also valid.

roles/edpm_neutron_dhcp/templates/wrappers/dnsmasq.j2
```diff
@@ -0,0 +1,54 @@
...
+
+CLI="nsenter --net=/run/netns/${NETNS} --preserve-credentials -m -t 1 podman"
+LOGGING="--log-driver k8s-file --log-opt path=/var/log/containers/stdouts/${NAME}.log"
+CMD='/usr/sbin/dnsmasq -k'
+
+LIST=$($CLI ps -a --filter name=neutron-dnsmasq- --format '{{.ID}}:{{.Names}}:{{.Status}}' | awk '{print $1}')
+
+# Find orphaned containers left for dead after its main process terminated by neutron parent process
+# FIXME(cjeanner): https://github.com/containers/libpod/issues/1703
+ORPHANTS=$(printf "%s\n" "${LIST}" | grep -E ":(Exited|Created)")
```
jpodivin, line 29: ORPHANS?

roles/edpm_neutron_dhcp/templates/wrappers/dnsmasq.j2
```diff
@@ -0,0 +1,54 @@
...
+if [ -n "${ORPHANTS}" ]; then
+    for orphant in $(printf "%s\n" "${ORPHANTS}" | awk -F':' '{print $1}'); do
+        echo "Removing orphaned container ${orphant}"
+        $CLI stop ${orphant} || true
```
jpodivin, line 33: These will pass no matter what, even if the runtime fails for some reason. It may be a good idea to at least report that the container wasn't removed.

jpodivin (changes requested):

roles/edpm_neutron_dhcp/templates/wrappers/dnsmasq.j2
slawqo: I don't want to change this logic now. It's like it was in Tripleo until now. I added TODO with Your comment about logging and I will want to get back to it later in the follow up PR. Would that be ok for You?

jpodivin: Leaving it for later is fine. As long as we have a note about the deficiency.

roles/edpm_neutron_metadata/molecule/default/converge.yml
```diff
@@ -22,6 +22,7 @@
     - ansible.builtin.include_role:
         name: "osp.edpm.edpm_neutron_metadata"
       vars:
+        ansible_user: root
```
karelyatin, line 25: what's triggering this change now? which user was used before?

karelyatin (changes requested):
some comments inline

roles/edpm_neutron_metadata/molecule/default/converge.yml
slawqo: I think that previously we didn't use this variable in this role and that's why it wasn't needed.

roles/edpm_neutron_dhcp/templates/wrappers/dnsmasq.j2
```diff
@@ -0,0 +1,58 @@
...
+# Extract the network namespace UUID from the command line args provided by
+# neutron. Typically of the form (with dnsmasq as an example):
+#
+# dnsmasq --no-hosts --no-resolv --except-interface=lo \
+#   --pid-file=/var/lib/neutron/dhcp/317716b8-919a-4a6f-8db1-78128ec3b100/pid  \
+#   --dhcp-hostsfile=/var/lib/neutron/dhcp/317716b8-919a-4a6f-8db1-78128ec3b100/host ...
+NETNS=$(ip netns identify)
+NAME=neutron-dnsmasq-${NETNS}
+
+CLI="nsenter --net=/run/netns/${NETNS} --preserve-credentials -m -t 1 podman"
```
booxter, line 23: I think --net= may not be necessary here since we are already inside the netns (as can be seen from line 20 and inspection of the logic here).

roles/edpm_neutron_dhcp/templates/wrappers/haproxy.j2
```diff
@@ -0,0 +1,59 @@
...
+{% raw -%}
+# Extract the network namespace UUID from the command line args provided by
+# neutron. Typically of the form (with dnsmasq as an example):
+#
+# dnsmasq --no-hosts --no-resolv --except-interface=lo \
+#   --pid-file=/var/lib/neutron/dhcp/317716b8-919a-4a6f-8db1-78128ec3b100/pid  \
+#   --dhcp-hostsfile=/var/lib/neutron/dhcp/317716b8-919a-4a6f-8db1-78128ec3b100/host ...
+NETNS=$(ip netns identify)
+NAME=neutron-haproxy-${NETNS}
+HAPROXY_CMD='$(if [ -f /usr/sbin/haproxy-systemd-wrapper ]; then echo "/usr/sbin/haproxy -Ds"; else echo "/usr/sbin/haproxy -Ws"; fi)'
```
booxter, line 22: I don't know details but I suspect that the wrapper is always present in our images and so we could simplify this command a bit. This would also allow to remove `bash` from the final CLI below.

Also, isn't it brittle to detect a "mode of operation" by the contents of the (common?) container image? I don't think we run haproxy under systemd, no? What if our image gets the wrapper installed in it? Will we change the behavior?

(AFAIU the wrapper is removed in RPMs. From changelog in Fedora:

```
* Fri Jan 05 2018 Ryan O'Hara <rohara@redhat.com> - 1.8.3-2
 - Remove haproxy-systemd-wrapper
```

But still, this seems brittle.

roles/edpm_neutron_dhcp/templates/wrappers/haproxy.j2
```diff
@@ -0,0 +1,59 @@
...
+{%- endif -%}
+
+ARGS="$@"
+
+IMAGE_NAME={{ edpm_neutron_dhcp_sidecar_haproxy_image_name }}
+NEUTRON_CONFIG_DIR={{ edpm_neutron_dhcp_agent_config_dir }}
+NEUTRON_LIB_DIR={{ edpm_neutron_dhcp_agent_lib_dir }}
+
+{% raw -%}
+# Extract the network namespace UUID from the command line args provided by
```
booxter, line 14: The logic of wrappers is largely repetitive. I wonder if we could consolidate it somehow.

roles/edpm_neutron_dhcp/templates/wrappers/dnsmasq.j2
```diff
@@ -0,0 +1,58 @@
...
+        $CLI rm -f ${orphant} || true
+    done
+fi
+
+# If the NAME is already taken by a container, give it an unique name
+printf "%s\n" "${LIST}" | grep -q "${NAME}$" && NAME="${NAME}-$(date +%Y-%m-%d-%H%M%S-%N)"
+echo "Starting a new child container ${NAME}"
+$CLI run --detach ${LOGGING} \
+     -v ${NEUTRON_CONFIG_DIR}:/etc/neutron:ro \
+     -v /run/netns:/run/netns:shared \
```
booxter, line 47: I wonder... why does dnsmasq need access to namespaces volume here? Same for /etc/neutron. Same for the other two mounts.

roles/edpm_neutron_dhcp/molecule/default/verify.yml
```diff
@@ -51,3 +51,31 @@
...
+
+    - name: Ensure that dnsmasq wrapper script was created
+      block:
+        - name: dnsmasq wrapper exists
```
booxter, line 71: I believe names of tasks must start from upper-case (after lint-production enforcement lands)

.github/workflows/molecule.yaml
```diff
@@ -27,6 +27,7 @@ jobs:
         - edpm_logrotate_crond
         - edpm_multipathd
         - edpm_network_config
+        - edpm_neutron_dhcp
```
booxter, line 30: Thanks for enabling the molecule.

As discussed elsewhere, I think we should add a check in molecule scenario that would validate sidecar are not cleaned up when agent(s) are down.

For this, you would 

- make sure the agent container is up (already done in test)
- create qdhcp- (or metadata equivalent) namespace from inside the agent container
- start dnsmasq (or metadata equivalent) via wrapper (calling /usr/local/bin/<service> from the agent container)
- stop the agent container
- check that qdhcp- (or metadata equivalent) namespace is up and the service process is up

roles/edpm_neutron_metadata/templates/wrappers/haproxy.j2
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
booxter, line 1: Looks like this file is not just similar but identical to what is used in the other dhcp role. I wonder if we could use a symlink here.

booxter (changes requested):
My main concern here is that this is not actually tested (I mean - the fact that sidecars are NOT cleaned up on agent down). I think molecule scenario should be expanded to validate it.

I think core folks are also forceful about new arg specs having their descriptions filled in.

The rest of my comments are more inspirational, but I would appreciate if you could address some of them, or at least capture some of the questions as TODOs or elsewhere e.g. in Jira.

booxter:
And also to capture the context of our 1:1 discussion that is relevant to my comments here for other reviewers, there is a concern that the existing tripleo integration for sidecar survival may not do what it's supposed to do, at least if I interpret the relevant tobiko test case correctly:

https://opendev.org/x/tobiko/src/commit/5d04e37ab05c214fbfa36d1d299b2cb8d0c0025c/tobiko/tests/faults/neutron/test_agents.py#L792

booxter:
/hold

As per Slawek, this is WIP.

jpodivin (commented):
Good work. But I'm going to wait until others questions are answered before giving it ++

roles/edpm_neutron_metadata/templates/wrappers/haproxy.j2
slawqo: It's now done in more "unified" way. I hope it's fine for You

roles/edpm_neutron_dhcp/tasks/install.yml
```diff
@@ -19,15 +19,19 @@
     path: "{{ item.path }}"
     setype: "container_file_t"
     state: directory
-    owner: "{{ item.owner | default(lookup('env', 'USER')) }}"
```
jpodivin, line 22: Please revert this. The `ansible_user` var isn't safe, we need something that works all the time. You can change the user it's executed under instead.

jpodivin (changes requested):

roles/edpm_neutron_dhcp/templates/wrappers/dnsmasq.j2
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+{% if edpm_neutron_dhcp_sidecar_debug -%}
```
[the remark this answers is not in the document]
cmsj: @booxter The guidance we've received is that if an individual file we're copying (or copying non-trivial amounts of content from) contains a copyright/attribution notice, we should retain that in our new file. Additionally, if the source repository's README contains a global notice of copyright/attribution, we should retain that in our README.

If, on the other hand, neither of those things are true (which I believe is the case with at least this particular file), then we do not need to reconstruct a copyright/attribution listing from the source repo's history (which is good because there's like 160 committers to puppet-tripleo!)

roles/edpm_neutron_dhcp/tasks/install.yml
slawqo: Done

roles/edpm_neutron_metadata/molecule/default/verify.yml
```diff
@@ -15,3 +15,17 @@
             that:
               - metadata_config.stat.exists
             fail_msg: "metadata agent config file does not exist"
+
+    - name: Ensure that haproxy wrapper script was created
```
booxter, line 19: haven't you switched to testinfra? Should you not kill this script as you did for the other role?

roles/edpm_neutron_metadata/templates/wrappers/common_part.j2
```diff
@@ -0,0 +1,37 @@
+{% raw -%}
+
+CLI="nsenter --net=/run/netns/${NETNS} --preserve-credentials -m -t 1 podman"
+LOGGING="--log-driver k8s-file --log-opt path=/var/log/containers/stdouts/${NAME}.log"
+
+LIST=$($CLI ps -a --filter name=neutron-dnsmasq- --format '{{.ID}}:{{.Names}}:{{.Status}}' | awk '{print $1}')
```
booxter, line 6: shouldn't the name be different depending on the actual process?

roles/edpm_neutron_metadata/templates/wrappers/haproxy.j2
```diff
@@ -0,0 +1,25 @@
...
+
+IMAGE_NAME={{ edpm_neutron_metadata_agent_sidecar_haproxy_image_name }}
+NEUTRON_CONFIG_DIR={{ edpm_neutron_metadata_agent_config_dir }}
+NEUTRON_LIB_DIR={{ edpm_neutron_metadata_agent_lib_dir }}
+
+{% raw -%}
+# Extract the network namespace UUID from the command line args provided by
+# neutron. Typically of the form (with dnsmasq as an example):
+#
+# dnsmasq --no-hosts --no-resolv --except-interface=lo \
```
booxter, line 17: the comment is not valid for reasons I listed elsewhere (no CLI args are checked to extract namespace name), but here also because you list dnsmasq in haproxy wrapper.

booxter:
/lgtm

This is great!

PR #406, opened 2023-10-05
Remove neutron-cleanup service

booxter:
It is not relevant to ml2/ovn which is the driver used here. And for what I understand, it will kill the dataplane for all ports if run.


booxter:
/hold

Let's first confirm I don't miss some use case for the script.

@karelyatin @slawqo any idea why we have the service?

booxter:
/unhold

confirmed with Jakub Libosvar that this is not ml2/ovn material, but a remnant from ml2/ovs.

karelyatin (commented):
/lgtm

PR #407, opened 2023-10-05
Make `edpm_iscsid` role compliant with ansible-lint `production` profile

rebtoor:
- Fixed a couple of checks against boolean using a more "ansible" way
- Added "mode" in `file` and `copy` tasks to hardening security
- Replaced the use of `podman run` inside `command` tasks with `podman_container` module from `containers.podman` collection
- Implemented an handler task instead of a standard task with a `when: $foo.changed` condition as for ansible best practices
- Replaced `systemctl` usage inside a `command` module with `ansible.builtin.service_facts` module and adapted `when` conditions for systemd task
- Fixed the indentation of the tasks under `block`

### DIFF
--- a/roles/edpm_iscsid/handlers/main.yml
+++ b/roles/edpm_iscsid/handlers/main.yml
@@ -1,0 +1,8 @@
+---
+
+- name: Record the iscsid container restart is required
+  become: true
+  ansible.builtin.file:
+    path: /etc/iscsi/.iscsid_restart_required
+    state: touch
+    mode: "0600"
--- a/roles/edpm_iscsid/tasks/configure.yml
+++ b/roles/edpm_iscsid/tasks/configure.yml
@@ -26,57 +26,52 @@
 # .initiator_reset sentinel file.
 
 - name: Ensure the system has a unique IQN
-  when: initiator_reset_state.stat.exists == False
+  when: not initiator_reset_state.stat.exists
   become: true
   block:
 
-  - name: Generate a unique IQN
-    ansible.builtin.command: podman run -ti --rm --name iscsid_config {{ edpm_iscsid_image }} /usr/sbin/iscsi-iname
-    register: iscsi_iname
+    - name: Generate a unique IQN
+      containers.podman.podman_container:
+        name: iscsid_config
+        image: "{{ edpm_iscsid_image }}"
+        rm: true
+        tty: true
+        detach: false
+        command: /usr/sbin/iscsi-iname
+      register: iscsi_iname
 
-  - name: Save the new IQN
-    ansible.builtin.copy:
-      dest: /etc/iscsi/initiatorname.iscsi
-      content: "InitiatorName={{ iscsi_iname.stdout }}"
+    - name: Save the new IQN
+      ansible.builtin.copy:
+        dest: /etc/iscsi/initiatorname.iscsi
+        mode: "0644"
+        content: "InitiatorName={{ iscsi_iname.stdout | trim }}"
 
-  - name: Record the IQN has been reset
-    ansible.builtin.file:
-      path: /etc/iscsi/.initiator_reset
-      state: touch
+    - name: Record the IQN has been reset
+      ansible.builtin.file:
+        path: /etc/iscsi/.initiator_reset
+        mode: "0600"
+        state: touch
 
-- name: Check if /etc/iscsi/iscsid.conf exists
+- name: Check if /etc/iscsi/iscsid.conf exists on the host
   ansible.builtin.stat:
     path: /etc/iscsi/iscsid.conf
   register: result
 
-- name: Create /etc/iscsi/iscsid.conf if necessary
-  when: result.stat.exists == False
-  become: true
-  block:
-
-  - name: Fetch iscsid.conf from the iscsid container
-    ansible.builtin.command: podman run -ti --rm --name iscsid_config {{ edpm_iscsid_image }} cat /etc/iscsi/iscsid.conf
-    register: iscsid_conf
-
-  - name: Create a local copy of iscsid.conf
-    ansible.builtin.copy:
-      dest: /etc/iscsi/iscsid.conf
-      content: "{{ iscsid_conf.stdout }}"
-      mode: 0600
-    check_mode: false
+- name: Copy iscsid.conf from the iscsid container to the host
+  when: not result.stat.exists
+  containers.podman.podman_container:
+    name: iscsid_config
+    image: "{{ edpm_iscsid_image }}"
+    rm: true
+    command: cp /etc/iscsi/iscsid.conf /host/etc/iscsi/
+    volume:
+      - /etc/iscsi:/host/etc/iscsi
 
 - name: Write CHAP algorithms
+  become: true
   ansible.builtin.lineinfile:
     path: "/etc/iscsi/iscsid.conf"
     line: "node.session.auth.chap_algs = {{ edpm_iscsid_chap_algs }}"
     regexp: "^node.session.auth.chap_algs"
     insertafter: "^#node.session.auth.chap.algs"
-  become: true
-  register: modify_stat
-
-- name: Record the iscsid container restart is required
-  when : modify_stat.changed
-  become: true
-  ansible.builtin.file:
-   path: /etc/iscsi/.iscsid_restart_required
-   state: touch
+  notify: Record the iscsid container restart is required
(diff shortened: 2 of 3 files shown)


rebtoor:
> This is a significant rewrite, and will require extensive testing. Please don't merge until I can do that.

Sure thing, I'll leave you as the only reviewer of this PR, feel free to review it when you can.

> I also would like to see a more informative commit message. The motivation for the change may be motivated by the desire mentioned in the subject line, but the impact of what that involves should be described as well.

You're absolutely right, I'll resubmit the commit message with more info about changes.

roles/edpm_iscsid/tasks/configure.yml
```diff
@@ -26,57 +26,60 @@
...
+      register: iscsi_iname
 
-  - name: Save the new IQN
-    ansible.builtin.copy:
-      dest: /etc/iscsi/initiatorname.iscsi
-      content: "InitiatorName={{ iscsi_iname.stdout }}"
+    - name: Save the new IQN
+      ansible.builtin.copy:
+        dest: /etc/iscsi/initiatorname.iscsi
+        mode: "0600"
```
ASBishop, line 44: "0644" is more standard (it's not a security critical file)

roles/edpm_iscsid/tasks/configure.yml
```diff
@@ -26,57 +26,60 @@
...
-  when: result.stat.exists == False
+  when: not result.stat.exists
   become: true
   block:
 
-  - name: Fetch iscsid.conf from the iscsid container
-    ansible.builtin.command: podman run -ti --rm --name iscsid_config {{ edpm_iscsid_image }} cat /etc/iscsi/iscsid.conf
-    register: iscsid_conf
+    - name: Fetch iscsid.conf from the iscsid container
+      containers.podman.podman_container:
```
ASBishop, line 64: Same here, this doesn't work because all you get is the container ID.

roles/edpm_iscsid/tasks/configure.yml
```diff
@@ -26,57 +26,60 @@
...
-    ansible.builtin.copy:
-      dest: /etc/iscsi/iscsid.conf
-      content: "{{ iscsid_conf.stdout }}"
-      mode: 0600
-    check_mode: false
+    - name: Create a local copy of iscsid.conf
+      ansible.builtin.copy:
+        dest: /etc/iscsi/iscsid.conf
+        content: "{{ iscsid_conf.stdout }}"
+        mode: "0600"
```
ASBishop, line 75: This mode is correct.

roles/edpm_iscsid/tasks/install.yml
```diff
@@ -43,16 +43,14 @@
         enabled: false
       when: stat_iscsid_socket.stat.exists
 
-    - name: Check if iscsi.service is enabled
-      ansible.builtin.command: systemctl is-enabled --quiet iscsi.service
-      failed_when: false
-      register: iscsi_service_enabled_result
+    - name: Gather services facts
+      ansible.builtin.service_facts:
```
ASBishop, line 47: I'm OK with this, but bear in mind tripleo deployments struggled with performance and memory consumption issues, and one thing they did to mitigate the problem(s) was to avoid global fact gathering when the requirement was more specific. See commits like https://review.opendev.org/c/openstack/tripleo-ansible/+/775708.

In this instance, I don't care about ALL services, I just want to know about the iscsi service.

roles/edpm_iscsid/handlers/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+
+- name: Record the iscsid container restart is required
```
ASBishop, line 3: Ah, good! The first time I looked at this PR, the handler was empty and I was wondering how it was all supposed to work! This makes sense now.

ASBishop (changes requested):

roles/edpm_iscsid/tasks/configure.yml
```diff
@@ -26,57 +26,60 @@
...
-  when: initiator_reset_state.stat.exists == False
+  when: not initiator_reset_state.stat.exists
   become: true
   block:
 
-  - name: Generate a unique IQN
-    ansible.builtin.command: podman run -ti --rm --name iscsid_config {{ edpm_iscsid_image }} /usr/sbin/iscsi-iname
-    register: iscsi_iname
+    - name: Generate a unique IQN
+      containers.podman.podman_container:
```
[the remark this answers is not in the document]
rebtoor: Ah! Nice catch. By default it prints the container id. The missing piece, other than `tty: true` (`-t`), is `detach: false` (it's `true` for default). From the podman-run manual:

```
    --detach, -d
       Detached mode: run the container in the background and print the new container ID. The default is false.
```

and here's the output i had locally:

```
TASK [Debug] *********************************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "container_output": {
        "actions": [
            "started container"
        ],
        "changed": true,
        "container": {},
        "failed": false,
        "podman_actions": [
            "podman run --name container --tty=True --detach=False --rm quay.io/podified-antelope-centos9/openstack-iscsid:current-podified /usr/sbin/iscsi-iname"
        ],
        "podman_systemd": {},
        "stderr": "",
        "stderr_lines": [],
        "stdout": "iqn.1994-05.com.redhat:e5fb951692c\r\n",
        "stdout_lines": [
            "iqn.1994-05.com.redhat:e5fb951692c"
        ]
    }
}
```

roles/edpm_iscsid/tasks/configure.yml
rebtoor: ack

roles/edpm_iscsid/tasks/configure.yml
rebtoor: solved, as above

roles/edpm_iscsid/tasks/install.yml
rebtoor: It's a fair point and I totally understand it.

We'd like to avoid the use of "command" and "shell" modules when a native ansible module is available but obviously not at the price of deployment performance. 

I'd go with this "format" for now (since analog tasks are already been merged for all the other roles) but we'll keep this aspect monitored especially during scale tests. Let me know WDYT.

roles/edpm_iscsid/tasks/configure.yml
```diff
@@ -26,57 +26,64 @@
...
-  - name: Create a local copy of iscsid.conf
-    ansible.builtin.copy:
-      dest: /etc/iscsi/iscsid.conf
-      content: "{{ iscsid_conf.stdout }}"
-      mode: 0600
-    check_mode: false
+    - name: Create a local copy of iscsid.conf
+      ansible.builtin.copy:
+        dest: /etc/iscsi/iscsid.conf
+        content: "{{ iscsid_conf.stdout }}"
```
ASBishop, line 78: For some reason the file it creates contains DOS line endings. I have no idea why, and couldn't find a way to fix the file other than adding another play to this block. Then I hit upon a solution that works quite well, and also eliminates the need for a block.

Try replacing L55..L80 with these. Notice I slightly changed the "name" of the first one. The main change is the second.

```
- name: Check if /etc/iscsi/iscsid.conf exists on the host
  ansible.builtin.stat:
    path: /etc/iscsi/iscsid.conf
  register: result

- name: Copy iscsid.conf from the iscsid container to the host
  when: not result.stat.exists
  containers.podman.podman_container:
    name: iscsid_config
    image: "{{ edpm_iscsid_image }}"
    rm: true
    command: cp /etc/iscsi/iscsid.conf /host/etc/iscsi/
    volume:
      - /etc/iscsi:/host/etc/iscsi

```

ASBishop (changes requested):
This is looking good, but I'm requesting another change.

roles/edpm_iscsid/tasks/install.yml
ASBishop: I understand the tradeoffs, and am fine with using service_facts. We can revert to using the prior method if performance/scaling becomes a problem.

roles/edpm_iscsid/tasks/configure.yml
rebtoor: you're right, there are `\r` for each EOL.

I'll push your suggested fix.

ASBishop (approved):
/lgtm

Thanks, Roberto!

PR #413, opened 2023-10-06
Make `edpm_nova` role compliant with ansible-lint `production` profile

rebtoor:
- Rename all tasks capitalizing the first letter
- Define "mode" as string in all impacted tasks
- Trim extra spaces in `arguments_specs.yml` file
- Fix spacing issues
- Fixed tasks to use FQCN


roles/edpm_nova/tasks/configure.yml
```diff
@@ -12,65 +13,65 @@
...
+    - { "path": "/var/lib/nova" }
+    - { "path": "/var/lib/_nova_secontext" }
+    - { "path": "/var/lib/nova/instances" }
+    - { "path": "/var/log/containers/nova", "mode": "0750" }
+    - { "path": "/var/log/containers/stdouts" }
+    - { "path": "/etc/ceph", "mode": "0750", "owner": "root", "group": "root" }
+    - { "path": "/etc/multipath" }
+    - { "path": "/etc/iscsi" }
+    - { "path": "/var/lib/iscsi" }
+    - { "path": "/run/openvswitch" }
```
rebtoor, line 40: @SeanMooney @gibizer can we define "mode" for these dirs? For iscsi related dirs i can take the values from `edpm_iscsi` role but I'll need your help for other directories. :)

roles/edpm_nova/tasks/configure.yml
```diff
@@ -12,65 +13,65 @@
     group: "{{ item.group | default(ansible_user) }}"
     mode: "{{ item.mode | default(omit) }}"
   with_items:
-  - { "path": "{{edpm_nova_config_dest}}" }
-  - { "path": "/var/lib/openstack/config/containers"}
-- name: create persistent directories
+    - { "path": "{{ edpm_nova_config_dest }}" }
+    - { "path": "/var/lib/openstack/config/containers" }
```
rebtoor, line 17: @SeanMooney @gibizer Same as below... which "mode" can we define here?

SeanMooney: mode is intentionally not defined here but if we were to set it needs to be "0755" to ensure the traversability of the directories. "0750" will break things as will dropping the executable bit and just have read access

roles/edpm_nova/tasks/configure.yml
SeanMooney: so we are not defiening mode so that we don't change it if its already crated by another role.

The nova role needs these directories to exist because we bind mount them in to the nova container.  the other way to do this would be to not define the 
```
   - { "path": "/var/log/containers/stdouts" }
  - { "path": "/etc/ceph", "mode": "0750", "owner": "root", "group": "root"}
    - { "path": "/etc/ceph", "mode": "0750", "owner": "root", "group": "root" }
  - { "path": "/etc/multipath" }
    - { "path": "/etc/multipath" }
  - { "path": "/etc/iscsi" }
    - { "path": "/etc/iscsi" }
  - { "path": "/var/lib/iscsi" }
    - { "path": "/var/lib/iscsi" }
  - { "path": "/run/openvswitch" }
    - { "path": "/run/openvswitch" }
 ```
and assume they are created by other roles

but update the modeclue prepare playbook to create them.

perhaps that is the better approach

by default for directories our modes should be 0755 if they may be share between different services or where traversiablity to sub dirs is a factor.

Comment (rebtoor) [reply roles/edpm_nova/tasks/configure.yml:40]: yeah, even "0755" is fine if needed. 

I agree with the path that you are suggesting and i'll fix the molecule prepare playbook as well.

Comment (SeanMooney) [reply roles/edpm_nova/tasks/configure.yml:40]: just lookign at them /etc/ceph is nolonger used by the container we now use a spereate directory in /var/lib/openstack/config/ceph i belive so you can entirely remove the reference to /etc/cpeh i belive but i would check with @fultonj to confirm

Comment (fultonj) [reply roles/edpm_nova/tasks/configure.yml:40]: I believe Sean is right. If you include that change here the CI should catch it if we're mistaken. It could also be addressed in a follow up patch if we want to limit the scope of this change to satisfying the linter.

Comment (rebtoor) [reply roles/edpm_nova/tasks/configure.yml:40]: I'll leave ceph directory as special case since you're right that we can't test it by now  and I'm ensuring that other folders are created by other roles before removing them from here.

Comment (rebtoor) [reply roles/edpm_nova/tasks/configure.yml]: I added "mode: 0755" only for nova folders, left the other dirs untouched. It shouldn't be an issue since they are created by their own roles.

I'll propose a followup patch in the future.

**Comment by SeanMooney** [comment]
ill take a look at this again on Monday but at a quick glance it looks ok to me
I'm not sure if you are planning to factor out the other directories in this pr or not so let me know  what your preference is.

**Comment by rebtoor** [comment]
recheck

**Comment by rebtoor** [comment]
recheck

File: roles/edpm_nova/tasks/configure.yml [hunk 3]
```diff
@@ -12,65 +13,65 @@
     group: "{{ item.group | default(ansible_user) }}"
     mode: "{{ item.mode | default(omit) }}"
   with_items:
```
bshephar, line 15: Since this PR is about bringing the role up to the standard of the repo, it might be nice to use loop instead of with_items.

bshephar (commented):
Looks good, just a comment about the use of `with_items`. Not worth blocking over though, we can fix it in a later PR if you prefer.

roles/edpm_nova/tasks/configure.yml
rebtoor: Thanks @bshephar !

rebtoor:
> ill take a look at this again on Monday but at a quick glance it looks ok to me I'm not sure if you are planning to factor out the other directories in this pr or not so let me know what your preference is.

I'll leave those entries as-is for now since they don't trigger any errors in the linter check. I need to test the impact of the change before removing them, so I'll propose a followup PR later.

fultonj (approved):
/lgtm

roles/edpm_nova/tasks/configure.yml
SeanMooney: why woud we prefer loop over with items?

SeanMooney: ya i just found that.
i general dislike the loop syntax but i can try an d use it.
form an understanding point of view not haveing typed loops i feel is more error prone but
if that is the new way i guess i can try an unlern my old ansibel preferences for using with_*

rebtoor: > why woud we prefer loop over with items?

Just for coherence with the rest of the repo. We started using `loop` while importing roles so we're using as a standard (where possible, obviously).  BTW, just to be fair, it isn't deprecated (as stated in the notes of the doc linked by @fultonj) nor a linting issue.

PR #422, opened 2023-10-09
[edpm_ssh_known_hosts]Fix hostname pattern

gibizer:
Originally the role generated hostnames in the known_hosts file with the
following pattern:
[edpm-compute-0]*,...

This pattern does not work as it does not match with the edpm-compute-0
hostname:
```
[root@edpm-compute-0 ~]# ssh edpm-compute-0
The authenticity of host 'edpm-compute-0 (192.168.122.100)' can't be established.
ED25519 key fingerprint is SHA256:ba66xyWDSkO0qv/9EATAZWT4NHhgDnScka4U7FIreaY.
This host key is known by the following other names/addresses:
    /etc/ssh/ssh_known_hosts:4: [edpm-compute-0]*
Are you sure you want to continue connecting (yes/no/[fingerprint])?
```

According to the ssh docs * and ? can be used for globbing. However []
can only be used for defining hosts with non default ssh port in the form
of [hostname]:portnumber. Based on my testing on EDPM nodes. The globbing
does not work for port numbers even if I add the missing ':'. Our ssh
server in EDPM runs on the default port so the []: format is not at all
needed. Also we have separate hostkeys for each node so globbing on the
hostname does not needed either.

This PR removes the unnecessary complication from the known_hosts
generation and therefore fixing the hostname matching during host key
verification.

### DIFF
--- a/roles/edpm_ssh_known_hosts/tasks/main.yml
+++ b/roles/edpm_ssh_known_hosts/tasks/main.yml
@@ -63,16 +63,16 @@
           {%     set enabled_host_networks = hostdata['edpm_role_networks']|default([]) %}
           {%     for network in enabled_host_networks %}
           {%       if (network ~ '_ip') in hostdata %}
-          {%         set _ = entries.append('[' ~ hostdata[network ~ '_ip'] ~ ']*') %}
+          {%         set _ = entries.append(hostdata[network ~ '_ip']) %}
           {%       endif %}
           {%       if (network ~ '_hostname') in hostdata %}
-          {%         set _ = entries.append('[' ~ hostdata[network ~ '_hostname'] ~ ']*') %}
+          {%         set _ = entries.append(hostdata[network ~ '_hostname']) %}
           {%       endif %}
           {%     endfor %}
           {%     if 'canonical_hostname' in hostdata %}
-          {%       set _ = entries.append('[' ~ hostdata['canonical_hostname'] ~ ']*') %}
+          {%       set _ = entries.append(hostdata['canonical_hostname']) %}
           {%     endif %}
-          {%     set _ = entries.append('[' ~ host ~ ']*') %}
+          {%     set _ = entries.append(host) %}
           {%     set host = entries|unique|join(',') %}
           {%     for key, type in key_types.items() %}
           {%       if key in hostdata['ansible_facts'] %}
--- a/roles/edpm_ssh_known_hosts/molecule/default/tests/test_default.py
+++ b/roles/edpm_ssh_known_hosts/molecule/default/tests/test_default.py
@@ -25,9 +25,9 @@ testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
 
 def test_ssh_host_keys(host):
     expected = [
-        '[10.0.0.1]*,[centos.ctlplane.localdomain]*,[10.0.1.1]*,[centos.internalapi.localdomain]*,[centos.localdomain]*,[instance]* ssh-rsa AAAATESTRSA',
-        '[10.0.0.1]*,[centos.ctlplane.localdomain]*,[10.0.1.1]*,[centos.internalapi.localdomain]*,[centos.localdomain]*,[instance]* ssh-ed25519 AAAATESTED',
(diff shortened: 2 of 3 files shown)


rebtoor (approved):

rebtoor:
@gibizer the fix is correct but I'm wondering if we should handle the case where we have a non-standard ssh port (so we need to keep the square brackets).

gibizer:
> @gibizer the fix is correct but I'm wondering if we should handle the case where we have a non-standard ssh port (so we need to keep the square brackets).

Sure but then we need to know the port number of the nonstandard port to store something like  `[edpm-compute-0]:2222 ` in the known_hosts file

rebtoor:
> > @gibizer the fix is correct but I'm wondering if we should handle the case where we have a non-standard ssh port (so we need to keep the square brackets).
> 
> Sure but then we need to know the port number of the nonstandard port to store something like `[edpm-compute-0]:2222 ` in the known_hosts file

IIRC, we are passing the `ansible_port` variable while generating the inventory (if `ansible_port` is > 0) 

BTW it's not the scope of this PR, we can address this later in a followup patch.

gibizer:
recheck
retry limits

gibizer:
> > > @gibizer the fix is correct but I'm wondering if we should handle the case where we have a non-standard ssh port (so we need to keep the square brackets).
> > 
> > 
> > Sure but then we need to know the port number of the nonstandard port to store something like `[edpm-compute-0]:2222 ` in the known_hosts file
> 
> IIRC, we are passing the `ansible_port` variable while generating the inventory (if `ansible_port` is > 0)
> 
> BTW it's not the scope of this PR, we can address this later in a followup patch.

good points. I will take a look separately.

Meanwhile I checked that the following format is correct if the sshd is on a nonstandard port:
```
root@92c0c5f5d63c:~# cat .ssh/known_hosts 
[192.168.1.171]:2222 ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIELzaxdMurFChcjKxQjoSooobGWCaSNN/XZ33YHZafoN
```

roles/edpm_ssh_known_hosts/tasks/main.yml
```diff
@@ -63,16 +63,16 @@
...
           {%     for network in enabled_host_networks %}
           {%       if (network ~ '_ip') in hostdata %}
-          {%         set _ = entries.append('[' ~ hostdata[network ~ '_ip'] ~ ']*') %}
+          {%         set _ = entries.append(hostdata[network ~ '_ip']) %}
           {%       endif %}
           {%       if (network ~ '_hostname') in hostdata %}
-          {%         set _ = entries.append('[' ~ hostdata[network ~ '_hostname'] ~ ']*') %}
+          {%         set _ = entries.append(hostdata[network ~ '_hostname']) %}
           {%       endif %}
           {%     endfor %}
```
rabi, line 71: Why remove the `*`? That would allow us to ssh using fqdn when the entries are something like...

`edpm-compute-1* ssh-rsa .....`

gibizer: as far as I see we already generate hostname and canonica_hostname as an alternative name for the given hostkey. So I assumed that the name globbing is not needed as the fqdn will be added explicitly to the known_hosts file

rabi: We don't have `canonical_hostname` in host vars atm and that entry is not added. Though it could add that var in inventory, I would have left the name globbing.

gibizer: I can add back the globbing selectively to the hostname case but having it at the IP or the canonical_hostname is an error.

gibizer: https://github.com/openstack-k8s-operators/edpm-ansible/pull/429

gibizer:
> > > @gibizer the fix is correct but I'm wondering if we should handle the case where we have a non-standard ssh port (so we need to keep the square brackets).
> > 
> > 
> > Sure but then we need to know the port number of the nonstandard port to store something like `[edpm-compute-0]:2222 ` in the known_hosts file
> 
> IIRC, we are passing the `ansible_port` variable while generating the inventory (if `ansible_port` is > 0)
> 
> BTW it's not the scope of this PR, we can address this later in a followup patch.
@rebtoor here is an improvement that adds non standard port support: https://github.com/openstack-k8s-operators/edpm-ansible/pull/438

PR #424, opened 2023-10-09
ovn: actually apply hw-offload setting

booxter:
The code was constructing a new dict but never applying it.

### DIFF
--- a/roles/edpm_ovn/molecule/default/verify-tasks.yml
+++ b/roles/edpm_ovn/molecule/default/verify-tasks.yml
@@ -1,0 +1,96 @@
+# we have to use shell to extract the values because openvswitch ansible
+# module doesn't seem to support dashes in key names, see:
+# https://github.com/ansible-collections/openvswitch.openvswitch/issues/111
+- name: Verify ovn-remote
+  ansible.builtin.shell: >
+    /usr/bin/ovs-vsctl get open_vswitch . external_ids:ovn-remote
+  register: output
+  failed_when: output.stdout != '"tcp:1.2.3.4:5678"'
+
+- name: Verify ovn-encap-type
+  ansible.builtin.shell: >
+    /usr/bin/ovs-vsctl get open_vswitch . external_ids:ovn-encap-type
+  register: output
+  failed_when: output.stdout != 'geneve'
+
+- name: Verify ovn firewall rules are applied
+  block:
+    - name: Rule source file exists
+      become: true
+      ansible.builtin.stat:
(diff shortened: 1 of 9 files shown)


roles/edpm_ovn/tasks/configure.yml
```diff
@@ -66,7 +66,7 @@
 
 - name: Configure OVS other_config
   ansible.builtin.shell: >
-    ovs-vsctl set open . {% for key, value in edpm_ovn_ovs_other_config.items() %} other_config:{{ key }}={{ value }} {% endfor %}
+    ovs-vsctl set open . {% for key, value in other_config.items() %} other_config:{{ key }}={{ value }} {% endfor %}
```
karelyatin, line 69: need to change L61 instead so it works with edpm_enable_hw_offload=false:-
edpm_ovn_ovs_other_config: "{{ edpm_ovn_ovs_other_config | combine(hw_offload) }}"

booxter:
@jpodivin would you mind confirming this is the right approach to avoid duplicating code over multiple molecule scenarios? (using symlinks + common tasks yaml that is included in multiple places).

roles/edpm_ovn/molecule/default/verify-tasks.yml
```diff
@@ -0,0 +1,96 @@
...
+    - name: Check if nftables rule exists in /etc/nftables/edpm-rules.nft
+      become: true
+      ansible.builtin.shell: grep -q "{{ item }}" /etc/nftables/edpm-rules.nft
+      register: rule_exists
+      ignore_errors: true
+      loop:
+        - "118 neutron vxlan networks"
+        - "119 neutron geneve networks"
+        - "120 neutron geneve networks no conntrack"
+        - "121 neutron geneve networks no conntrack"
```
slawqo, line 39: I wonder if wouldn't it be better to move to testinfra as verifier. It has easy to used dedicated modules to e.g. test if file exists and if it contains some strings, see https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.file.File
I'm using it now in the edpm_neutron_{metadata,dhcp} roles to test sidecars

booxter: Perhaps though here I'm just moving code around.

booxter:
recheck

booxter:
Failed to download some files for the env, unrelated failure

booxter:
recheck

jpodivin (approved):
Great work!

PR #425, opened 2023-10-09
ovn: don't explicitly set vlan-limit=0

booxter:
ovn-controller does it automatically since 21.09.0 [1].

https://github.com/ovn-org/ovn/commit/7e2c892c380aeddc4bbc9ceb5e759ee9ac62176d

### DIFF
--- a/roles/edpm_ovn/defaults/main.yml
+++ b/roles/edpm_ovn/defaults/main.yml
@@ -85,5 +85,4 @@ edpm_ovn_ovs_external_ids:
   rundir: "/var/run/openvswitch"
 
 # Set openvswitch other_config.
-edpm_ovn_ovs_other_config:
-  vlan-limit: 0
+edpm_ovn_ovs_other_config: {}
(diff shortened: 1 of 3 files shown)


roles/edpm_ovn/defaults/main.yml
```diff
@@ -93,8 +93,7 @@ edpm_ovn_ovs_external_ids:
   rundir: "/var/run/openvswitch"
 
 # Set openvswitch other_config.
-edpm_ovn_ovs_other_config:
-  vlan-limit: 0
+edpm_ovn_ovs_other_config: {}
```
karelyatin, line 96: to handle empty dict need to add when condition at https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_ovn/tasks/configure.yml#L69C5-L69C131 
can also add that to external_ids as same applies there too.

slawqo:
Please rebase it to resolve conflicts and then it's LGTM for me :)

PR #429, opened 2023-10-10
[edpm_ssh_known_hosts] Add back hostname globbing

gibizer:
The #422 removed all the name globbing as it was in the wrong place. Now this PR adds back globbing for hostname to support the case when hostname is not and fqdn during deployment but later and fqdn is used to ssh between nodes.

### DIFF
--- a/roles/edpm_ssh_known_hosts/tasks/main.yml
+++ b/roles/edpm_ssh_known_hosts/tasks/main.yml
@@ -72,7 +72,7 @@
           {%     if 'canonical_hostname' in hostdata %}
           {%       set _ = entries.append(hostdata['canonical_hostname']) %}
           {%     endif %}
-          {%     set _ = entries.append(host) %}
+          {%     set _ = entries.append(host ~ '*') %}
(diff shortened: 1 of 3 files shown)


jpodivin (approved):

roles/edpm_ssh_known_hosts/tasks/main.yml
```diff
@@ -66,7 +66,7 @@
           {%         set _ = entries.append(hostdata[network ~ '_ip']) %}
           {%       endif %}
           {%       if (network ~ '_hostname') in hostdata %}
-          {%         set _ = entries.append(hostdata[network ~ '_hostname']) %}
+          {%         set _ = entries.append(hostdata[network ~ '_hostname'] ~ '*') %}
```
rabi, line 69: We don't have  `ctlplane_hostname` and other `{{network}}_hostname` host vars either. I guess you've to change L75. I've also proposed https://github.com/openstack-k8s-operators/dataplane-operator/pull/462

rabi (approved):
/lgtm

PR #432, opened 2023-10-10
Update testing_with_ansibleee example for OpenStackDataPlaneNodeSet

fultonj:
With the switch from `OpenStackDataPlaneNode` to `OpenStackDataPlaneNodeSet`, the `nodeTemplate` entry for `extraMounts` is no longer under `roles` and now under `spec` so update the example accordingly.


slawqo (approved):
Thx a lot. I was struggling with this today :)

PR #437, opened 2023-10-11
Make `edpm_ssh_known_hosts` role compliant with ansible-lint `production` profile

rebtoor:
- Fixed indentation under `tags`
- Added `changed_when`/`failed_when` in `shell` task

### DIFF
--- a/roles/edpm_ssh_known_hosts/tasks/main.yml
+++ b/roles/edpm_ssh_known_hosts/tasks/main.yml
@@ -19,7 +19,7 @@
   become: true
   check_mode: false
   tags:
-  - edpm_ssh_known_hosts
+    - edpm_ssh_known_hosts
   block:
     - name: Create temporary file for ssh_known_hosts
       ansible.builtin.tempfile:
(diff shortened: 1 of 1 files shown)


roles/edpm_ssh_known_hosts/tasks/main.yml
```diff
@@ -88,14 +88,17 @@
         path: "{{ ssh_known_hosts_tmp.path }}"
         block: "{{ ssh_known_hosts_lines }}"
         create: true
-        mode: 0644
+        mode: "0644"
 
-    # Workaround https://bugs.launchpad.net/edpm/+bug/1810932
```
jpodivin, line 93: That webpage doesn't exist. If the bug isn't relevant anymore we should remove the link altogether, if it is we should keep it unchanged, maybe with a note about why it's still with us since OOO.

jpodivin (changes requested):

roles/edpm_ssh_known_hosts/tasks/main.yml
rebtoor: i thought i changed this url, let me push a fix

rebtoor: The comment about why this "hack" is needed is still valid. Ansible, when using modules that involves files, create a temporary files and then overwrite the targets. If those files are used by containers, they aren't upgraded.

jpodivin (approved):

PR #441, opened 2023-10-11
Merge the frr and ovn-bgp-agent roles into one

fernandoroyosanchez:
There is no need to split the roles in install, configure and run. This task is about unifying them so that we just have a frr role and an ovn-bgp-agent role.

To avoid breaking other repositories that still use the split roles, we will decompose the merging of the roles in two steps. This patch represents the first step, creating the future role that will replace the other three (install, configure, run).

A future patch, once it is confirmed that the dependent repositories are already using the unified role, will eliminate the split roles.

### DIFF
--- a/playbooks/frr.yml
+++ b/playbooks/frr.yml
@@ -1,0 +1,12 @@
+---
+
+- name: Deploy EDPM FRR
+  hosts: all
+  strategy: linear
+  become: true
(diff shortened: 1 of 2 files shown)


luis5tb:
recheck unrelated tempest_run_output.failed

PR #447, opened 2023-10-13
Execute `Ensure legacy iptables services are off` task only if iptables and ip6tables services are present

rebtoor:
If the task is executed in environment without those services, it raise an error:

```
TASK [osp.edpm.edpm_nftables : Ensure legacy iptables services are off] ********
Friday 13 October 2023  08:23:44 +0000 (0:00:03.253)       0:00:15.801 ******** 
failed: [edpm-compute-0] (item=iptables.service) => {"ansible_loop_var": "item", "changed": false, "item": "iptables.service", "msg": "Could not find the requested service iptables.service: host"}
failed: [edpm-compute-1] (item=iptables.service) => {"ansible_loop_var": "item", "changed": false, "item": "iptables.service", "msg": "Could not find the requested service iptables.service: host"}
failed: [edpm-compute-1] (item=ip6tables.service) => {"ansible_loop_var": "item", "changed": false, "item": "ip6tables.service", "msg": "Could not find the requested service ip6tables.service: host"}
failed: [edpm-compute-0] (item=ip6tables.service) => {"ansible_loop_var": "item", "changed": false, "item": "ip6tables.service", "msg": "Could not find the requested service ip6tables.service: host"}
```

### DIFF
--- a/roles/edpm_nftables/tasks/service-bootstrap.yml
+++ b/roles/edpm_nftables/tasks/service-bootstrap.yml
@@ -31,6 +31,8 @@
       loop:
         - iptables.service
         - ip6tables.service
+      when:
+        - ansible_facts.services["{{ item }}"] is defined
 
     - name: Ensure nftables service is enabled and running
       ansible.builtin.systemd:


roles/edpm_nftables/tasks/service-bootstrap.yml
```diff
@@ -31,6 +31,9 @@
       loop:
         - iptables.service
         - ip6tables.service
+      when:
```
eduolivares, line 34: Not sure if having only iptables without IPv6 is possible.
If it is, this could be better:
```
when:
  - ansible_facts.services["{{ item }}"] is defined
```

roles/edpm_nftables/tasks/service-bootstrap.yml
```diff
@@ -31,6 +31,9 @@
...
+        - ansible_facts.services["iptables.service"] is defined
+        - ansible_facts.services["ip6tables.service"] is defined
```
slagle, line 36: This should probably be an OR to maintain the previous behavor where either service may or may not exist.

rebtoor: `iptables` and `ip6tables` service aren't installed together?

rebtoor: https://access.redhat.com/solutions/1411953

in EL9:

- `iptables-nft-services` provides both services (v4 and v6)
- isn't installed by default
- it's deprecated

I can change it the condition to be an or but if one of the services isn't present it's not the standard scenario.

roles/edpm_nftables/tasks/service-bootstrap.yml
rebtoor: Se the below reply :) 

BTW using item it's better, I'll fix it ASAP

eduolivares (approved):

slagle (approved):

PR #448, opened 2023-10-13
[edpm_ssh_known_hosts]Cleanup unused code paths

gibizer:
The logic that collected the name and IP of the host was referring to variables that are not defined in the EDPM inventory today. Those are probably old tripleo variables. So this patch removes the dead code.

### DIFF
--- a/roles/edpm_ssh_known_hosts/tasks/main.yml
+++ b/roles/edpm_ssh_known_hosts/tasks/main.yml
@@ -60,15 +60,6 @@
           {%   set hostdata = hostvars[host] %}
           {%   if hostdata['ansible_facts'].keys() | intersect(key_types.keys()) %}
           {%     set entries = [] %}
-          {%     set enabled_host_networks = hostdata['edpm_role_networks']|default([]) %}
-          {%     for network in enabled_host_networks %}
-          {%       if (network ~ '_ip') in hostdata %}
-          {%         set _ = entries.append(hostdata[network ~ '_ip']) %}
-          {%       endif %}
-          {%       if (network ~ '_hostname') in hostdata %}
-          {%         set _ = entries.append(hostdata[network ~ '_hostname']) %}
-          {%       endif %}
-          {%     endfor %}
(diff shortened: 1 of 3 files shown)


roles/edpm_ssh_known_hosts/tasks/main.yml
```diff
@@ -60,15 +60,6 @@
           {%   set hostdata = hostvars[host] %}
           {%   if hostdata['ansible_facts'].keys() | intersect(key_types.keys()) %}
           {%     set entries = [] %}
-          {%     set enabled_host_networks = hostdata['edpm_role_networks']|default([]) %}
-          {%     for network in enabled_host_networks %}
```
fultonj, line 64: I thought that `edpm_role_networks` become `role_networks` from the following change:

https://github.com/openstack-k8s-operators/dataplane-operator/pull/441/files

and thus, we could still populate these vars if the key name is changed. Do you see `role_networks` in your inventory?

fultonj: @rabi any thoughts on this?

rabi: There was never an ansible var called `edpm_role_networks`, it was always `role_networks`[1].  Also, `role_networks` never contained `ctlplane` from TripleO times.


[1] https://github.com/openstack-k8s-operators/dataplane-operator/blob/dev-preview/config/samples/dataplane_v1beta1_openstackdataplane.yaml#L94-L103

gibizer: Based no that this data was never generated in NG to the known_hosts file due to the mismatching var names I believe these logic is unused in NG. Do we really want to maintain something that is unused? Can I simply drop the logic as proposed here? If there is a clear need for the extra logic later then we can add it back and fix it.

gibizer:
recheck

fultonj (commented):
/lgtm

PR #449, opened 2023-10-13
[edpm_ssh_known_hosts]Ensure host IP is added

gibizer:
Unfortunately nova still uses the IP of the compute nodes to scp data during cold migration instead of the hypervisor_hostname which is the FQDN of the compute node. So we need to make sure the ctlplane_ip of the node is also in the known_hosts file.

### DIFF
--- a/roles/edpm_ssh_known_hosts/tasks/main.yml
+++ b/roles/edpm_ssh_known_hosts/tasks/main.yml
@@ -72,6 +72,9 @@
           {%     if 'canonical_hostname' in hostdata %}
           {%       set _ = entries.append(hostdata['canonical_hostname']) %}
           {%     endif %}
+          {%     if 'ctlplane_ip' in hostdata %}
+          {%       set _ = entries.append(hostdata['ctlplane_ip']) %}
+          {%     endif %}
           {%     set _ = entries.append(host ~ '*') %}
           {%     if 'ansible_port' in hostdata and hostdata['ansible_port']|int != 22 %}
           {%       set hosts = [] %}


roles/edpm_ssh_known_hosts/tasks/main.yml
```diff
@@ -72,6 +72,9 @@
           {%     if 'canonical_hostname' in hostdata %}
           {%       set _ = entries.append(hostdata['canonical_hostname']) %}
           {%     endif %}
+          {%     if 'ctlplane_ip' in hostdata %}
```
gibizer, line 75: Note to the reviews: there is no addition test case as the ctlplane_ip is in the test input already and the host IP is in the output already because other fields. Those other fields are not exists in EDPM so they will be cleaned up separately.   https://github.com/openstack-k8s-operators/edpm-ansible/pull/448

slagle: The loop over the role networks that is cleaned up in the other PR (https://github.com/openstack-k8s-operators/edpm-ansible/pull/448/) is what set this previously. While we don't set edpm_role_networks, we do set role_networks. Perhaps the right fix is just to change that other PR with s/edpm_role_networks/role_networks.

Are we sure that this network traffic always uses the ctlplane?

gibizer: Good point. Let me try to combine the two PR and provide a fix that reuses the loop

gibizer: In the inventory we have:
```
        role_networks:
            - InternalApi
            - Storage
            - Tenant
```
and
```
            ctlplane_ip: 192.168.122.100
            internal_api_ip: 172.17.0.101
            storage_ip: 172.18.0.101
            tenant_ip: 172.19.0.101
```

Does it make sense to encode a transformation in ansible between the format of the role names (InternalApi) an the format of the IP key names (internal_api_ip)? Or would it be better to agree on a common naming scheme. Either internal_api in the role_networks or InternalAPIIP as the IP key?

Moreover I don't see a network name added that would point to ctlplane_ip. (Right now the EDPM compute nodes uses 192.168.122.x network for migration)

slagle: The transformation should already be available  in the networks_lower var.

https://github.com/openstack-k8s-operators/dataplane-operator/blob/5bcb2d658bbe651baf8faa6461cb410a0c37f695/pkg/deployment/inventory.go#L184

The ctlpane vars would still need to be added statically since ctlplane is not in role_networks. Perhaps it should be added there though. @rabi WDYT?

rabi: I don't think we can add ctlplane without changing network configs and it would break all old network configs used. https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml#L61-L83

gibizer:
recheck

roles/edpm_ssh_known_hosts/tasks/main.yml
jpodivin: Can it be merged even without #448  ?

gibizer: The change of this PR is independent from #448 . #448 tries to remove unused and broken code, while this PR adds the control plane IP of the node in the known_hosts file.

jpodivin: Ok.

jpodivin (approved):

PR #450, opened 2023-10-16
Fixed systemd checks for non existent services

rebtoor:


roles/edpm_nftables/tasks/service-bootstrap.yml
```diff
@@ -32,7 +32,7 @@
         - iptables.service
         - ip6tables.service
       when:
-        - ansible_facts.services["{{ item }}"] is defined
```
karelyatin, line 35: need to keep "is defined" condition on top else it would have issues when service itself don't exist, same at other places

roles/edpm_nftables/tasks/service-bootstrap.yml
```diff
@@ -32,7 +33,8 @@
...
+        - ansible_facts.services[item] is defined
+        - ansible_facts["services"][item]["status"] != "not-found"
```
karelyatin, line 37: ansible_facts.services[item]["status"] != "not-found" for consistency

rebtoor: daaaaamn, thanks @karelyatin

karelyatin (commented):
/lgtm

PR #459, opened 2023-10-18
Add `galaxy[version-incorrect]` to ansible-lint skip-list

rebtoor:
Since we are on version 0.0.1, that check fails because it expects at least a `1.x.x` version despite in the official documentation [1] the only requirement is to be compliant with semantic versioning.

[1] https://docs.ansible.com/ansible/latest/dev_guide/collections_galaxy_meta.html


jpodivin (approved):

fultonj:
+2

PR #461, opened 2023-10-18
CI testing for edpm_ovn_bgp_agent role

jpodivin:
This one has a lot of dependencies, so it calls other roles we have to set things up. 
Not an exactly isolated test, but it's about as close as I could get to one.


jpodivin:
recheck - failed with "The connection to the server api.crc.testing:6443 was refused - did you specify the right host or port?"

fao89 (approved):

rebtoor (approved):

PR #466, opened 2023-10-19
[edpm_libvirt]Add libvirt user to the EDPM host

gibizer:
This is needed for two reasons:
1) to eventually add the nova user to the libvirt group on the host so
   that the incoming live migration connections can access the libvirt
   socket
2) to allow libvirt to be de-containerized in https://github.com/openstack-k8s-operators/edpm-ansible/pull/467 while keeping the ui and guid
   of the libvirt user unchanged


jpodivin (approved):
LGTM

PR #472, opened 2023-10-20
Adding molecule tests for edpm_neutrond_dhcp role in the zuul CI

jpodivin:
Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/475

### DIFF
--- a/roles/edpm_neutron_dhcp/molecule/default/molecule.yml
+++ b/roles/edpm_neutron_dhcp/molecule/default/molecule.yml
@@ -4,7 +4,7 @@ dependency:
   options:
     role-file: collections.yml
 driver:
-  name: podman
+  name: delegated
 platforms:
 - command: /sbin/init
   dockerfile: ../../../../molecule/common/Containerfile.j2
--- a/zuul.d/projects.yaml
+++ b/zuul.d/projects.yaml
@@ -13,3 +13,4 @@
         - edpm-ansible-molecule-edpm_frr
         - edpm-ansible-molecule-edpm_iscsid
         - edpm-ansible-molecule-edpm_ovn_bgp_agent
+        - edpm-ansible-molecule-edpm_neutron_dhcp
--- a/roles/edpm_neutron_dhcp/molecule/default/converge.yml
+++ b/roles/edpm_neutron_dhcp/molecule/default/converge.yml
@@ -16,6 +16,7 @@
 
 - name: Converge
   hosts: all
+  become: true
   roles:
     - role: "edpm_neutron_dhcp"
       vars:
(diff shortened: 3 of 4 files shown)


jpodivin:
It appears that the `ansible_user` variable, doesn't in fact have to be set at all, so it isn't a reasonable default for user name/id. I'll look into alternatives.
https://github.com/ansible/ansible/issues/23530

roles/edpm_neutron_dhcp/molecule/default/molecule.yml
```diff
@@ -4,7 +4,7 @@ dependency:
   options:
     role-file: collections.yml
 driver:
-  name: podman
+  name: delegated
```
booxter, line 7: (please bear with me since I am not knowledgeable about molecule) does it switch to a driver that won't allow to run the role locally in podman? rationale? can't we run the agent inside the podman container?

zuul.d/projects.yaml
```diff
@@ -13,3 +13,4 @@
         - edpm-ansible-molecule-edpm_frr
         - edpm-ansible-molecule-edpm_iscsid
         - edpm-ansible-molecule-edpm_ovn_bgp_agent
+        - edpm-ansible-molecule-neutron-dhcp
```
booxter, line 16: should the name be in line with the rest of jobs?

roles/edpm_neutron_dhcp/molecule/default/molecule.yml
jpodivin: Exactly. There may be a way to run this test in a containerized environment. But from experience I can say that tests that require installation of packages and kernel modules are rarely worth the effort, if it's even possible in the first place.

zuul.d/projects.yaml
jpodivin: Good catch. Fixed.

roles/edpm_neutron_dhcp/molecule/default/molecule.yml
booxter: I believe @slawqo is able to run these in podman, incl. creating network namespaces etc. I'd prefer this stays this way for easier development.

slawqo: yes, I was able to even run sidecar container tests in https://github.com/openstack-k8s-operators/edpm-ansible/pull/394 using podman driver. And with podman driver is much easier to run molecule tests locally also which helps a lot with development. So unless there is any really important reason to switch to the delegated driver, I would prefer to stay with podman driver here for as long as it's possible.

jpodivin: Yes, but running it locally with podman means you have to give it root permissions. Which in my opinion is a no go, as it can cause changes to a dev machine and potentially contaminate environment for further testing.

Not to mention the damage it can potentially cause.

slawqo: yes, we discussed that already on Slack. I opened Jira ticket for that and will try to fix it in next weeks.

PR #474, opened 2023-10-20
Adding edpm_ceph_hci_pre molecule tests to CI

jpodivin:
Tests were adjusted to respect new settings of the role.


fmount (approved):
/lgtm

jpodivin:
recheck - failed with "error: unable to default to a user name: the server is currently unable to handle the request (get users.user.openshift.io ~)"

PR #478, opened 2023-10-25
[edpm_nova]Configure my_ip to ctlplane_ip for nova-compute

gibizer:
This is needed as nova-compute, if my_ip is not configured, tries to figure out its own IP address in complicated way and in a system with multiple network interfaces there is no guarantee that it selects the IP of the designated control plane interface.

The my_ip configuration is EDPM node specific so it cannot be generated globally for a cell in nova-operator. So this is generated by ansible now.

There is an upstream nova blueprint to eventually get rid of the need of this configuration: https://blueprints.launchpad.net/nova/+spec/libvirt-migrate-with-hostname-instead-of-ip

### DIFF
--- a/roles/edpm_nova/molecule/default/verify.yml
+++ b/roles/edpm_nova/molecule/default/verify.yml
@@ -43,6 +43,12 @@
         - "Copying /var/lib/kolla/config_files/02-nova-log.conf to /etc/nova/nova.conf.d/02-nova-log.conf"
         - "Copying /var/lib/kolla/config_files/ssh-config to /var/lib/nova/.ssh/config"
         - "Copying /var/lib/kolla/config_files/ssh-privatekey to /var/lib/nova/.ssh/ssh-privatekey"
+        - "Copying /var/lib/kolla/config_files/02-nova-host-specific.conf to /etc/nova/nova.conf.d/02-nova-host-specific.conf"
+
+    - name: Assert that my_ip is rendered into the host specific config
+      ansible.builtin.assert:
+        that:
+          - "'10.0.0.3' in lookup('ansible.builtin.file', '/var/lib/openstack/config/nova/02-nova-host-specific.conf')"
 
     - name: Check if user exists
       ansible.builtin.getent:
--- a/roles/edpm_nova/tasks/configure.yml
+++ b/roles/edpm_nova/tasks/configure.yml
@@ -51,6 +51,18 @@
     - {"src": "config.json.j2", "dest": "config.json"}
     - {"src": "nova-blank.conf", "dest": "nova-blank.conf"}
     - {"src": "ssh-config", "dest": "ssh-config"}
+    # NOTE(gibi): This is unfortunate as we would like to avoid config
+    # generation in ansible. This config is only needed to specify the IP
+    # address of the node nova-compute should use. Right now this is hardcoded
+    # to use the ctlplane_ip. In the future we might want to support
+    # configuring which network nova-compute uses when connecting between
+    # compute nodes. However that needs careful syncing between IPs in the
+    # known_hosts file and IPs in the TLS certificates.
+    # NOTE(gibi): There is an upstream nova blueprint that will allow us to
+    # remove this host specific configuration in the future (not earlier than
+    # openstack Caracal)
+    # https://blueprints.launchpad.net/nova/+spec/libvirt-migrate-with-hostname-instead-of-ip
+    - {"src": "02-nova-host-specific.conf.j2", "dest": "02-nova-host-specific.conf"}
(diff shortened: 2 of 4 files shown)


roles/edpm_nova/molecule/default/verify.yml
```diff
@@ -43,6 +43,12 @@
         - "Copying /var/lib/kolla/config_files/02-nova-log.conf to /etc/nova/nova.conf.d/02-nova-log.conf"
         - "Copying /var/lib/kolla/config_files/ssh-config to /var/lib/nova/.ssh/config"
         - "Copying /var/lib/kolla/config_files/ssh-privatekey to /var/lib/nova/.ssh/ssh-privatekey"
+        - "Copying /var/lib/kolla/config_files/02-nova-host-specific.conf to /etc/nova/nova.conf.d/02-nova-host-specific.conf"
+
+    - name: Assert that my_ip is rendered into the host specific config
+      ansible.builtin.assert:
+        that:
+          - "'10.0.0.3' in lookup('ansible.builtin.file', '/var/lib/openstack/config/nova/02-nova-host-specific.conf')"
```
SeanMooney, line 51: ack so this was executed in ci correctly

https://review.rdoproject.org/zuul/build/b2cdfbf23ae648a6a902c7cd8f383c73/log/ansible-execution.log#1301-1303

roles/edpm_nova/templates/02-nova-host-specific.conf.j2
```diff
@@ -0,0 +1,2 @@
+[DEFAULT]
+my_ip = {{ ctlplane_ip }}
```
SeanMooney, line 2: so... the main openquestion i have is should we introduce an ansible var to select what interface this should be or hardcode the control plane.

for context this IP will need to be included in certs and or know hosts files so we likely need a way to align this to one of the ansibel managed interface rather then just the default route.

rebtoor:
> my_ip in nova is basically the ip of the interface with the default route. that is not in general garunetted to be the contolplane interface so i agree this is required if we want it to be the control plane interface

If we need to retrieve the ip of the interface with the default route, what about using `ansible_default_ipv4`? 

> i have one openquestion in line so not approving for now until we agree on if we need this to be configurable as an ansible var (which would default to the contolplane IP)

roles/edpm_nova/templates/02-nova-host-specific.conf.j2
gibizer: I will look into making the nova management network configurable in the role and default it to ctlplane

gibizer, line 2: Note to myself: check if nova accepts a hostname/FQDN here, just for curiosity. :)

gibizer:
> > my_ip in nova is basically the ip of the interface with the default route. that is not in general garunetted to be the contolplane interface so i agree this is required if we want it to be the control plane interface
> 
> If we need to retrieve the ip of the interface with the default route, what about using `ansible_default_ipv4`?
> 
> edit: obviously we need to be sure to have that fact first (an `ansible.builtin.setup` task retrieving only the network subset of facts, for example)

In the contrary, we don't want the nova default behavior (which is looking up the default route from the host) to be used as that is not guaranteed to be the IP we need to ssh between the compute nodes. So we want to set this explicitly. By default to the ctlplane_ip but maybe we want to make it configurable so the network can be defined to be used to sshe between computes.

roles/edpm_nova/templates/02-nova-host-specific.conf.j2
gibizer: For the sake of progress I suggest to merge this as is and get back to extend configuration later.

roles/edpm_nova/tasks/configure.yml
```diff
@@ -51,6 +51,18 @@
...
+    # address of the node nova-compute should use. Right now this is hardcoded
+    # to use the ctlplane_ip. In the future we might want to support
+    # configuring which network nova-compute uses when connecting between
+    # compute nodes. However that needs careful syncing between IPs in the
+    # known_hosts file and IPs in the TLS certificates.
+    # NOTE(gibi): There is an upstream nova blueprint that will allow us to
+    # remove this host specific configuration in the future (not earlier than
+    # openstack Caracal)
+    # https://blueprints.launchpad.net/nova/+spec/libvirt-migrate-with-hostname-instead-of-ip
+    - {"src": "02-nova-host-specific.conf.j2", "dest": "02-nova-host-specific.conf"}
```
SeanMooney, line 65: +1

SeanMooney (approved):

PR #483, opened 2023-11-01
Adding a non-voting adoption job

jpodivin:


slagle (changes requested):
See my comments on https://github.com/openstack-k8s-operators/dataplane-operator/pull/502

PR #486, opened 2023-11-07
cmdline reading issue in kernel args

Jaganathancse:


jpodivin (changes requested):
Could you please point to the issue this is supposed to solve? As it is I have no idea.

jpodivin:
Ok. Could you please put that in the commit msg?

Jaganathancse:
> Ok. Could you please put that in the commit msg?
@jpodivin  Done.

PR #489, opened 2023-11-09
Added downgrade of crun in molecule scenario where podman is needed

rebtoor:
There's a regression in crun 1.11.1 (see https://github.com/containers/crun/issues/1338) that prevents podman to being executed on podman containers in some molecule scenario.

We are downgrade crun to last known version (1.10.1) until the new package will arrive in Centos 9 Stream.

### DIFF
--- a/roles/edpm_frr/molecule/default/prepare.yml
+++ b/roles/edpm_frr/molecule/default/prepare.yml
@@ -26,6 +26,17 @@
           - python3-pip
           - os-net-config
   tasks:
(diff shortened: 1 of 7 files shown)


gibizer (approved):

karelyatin (commented):
/lgtm

https://github.com/containers/crun/commit/7c3393ce3781bfbab6d1884150cc3fccb4c8e797 triggered the issue

PR #491, opened 2023-11-09
Mount TLS certs

vakwetu:
The dataplane operator will provide certs and keys to the ansibleEE pod in /var/lib/openstack/certs as well as a cacert file in /var/lib/openstack/cacerts.  It will also pass an ansible variable tls_certs_enabled to indicate that the certs are present.

This PR copies these files to the appropriate place on the compute node and then mounts these files in the nova-compute container.

We still need to confirm that these files are in the right location in the container and add config to use these certs - as well as code to trust the cacert.  Will need help from compute DFG for that.

### DIFF
--- a/roles/edpm_nova/defaults/main.yml
+++ b/roles/edpm_nova/defaults/main.yml
@@ -28,3 +28,10 @@ edpm_nova_image_download_retries: 5
 edpm_nova_config_src: /var/lib/openstack/configs
 edpm_nova_config_dest: /var/lib/openstack/config/nova
 edpm_nova_compute_image: "quay.io/podified-antelope-centos9/openstack-nova-compute:current-podified"
+
+# certs
+edpm_nova_certs_src: /var/lib/openstack/certs
+edpm_nova_certs_dest: /var/lib/openstack/certs/nova
+edpm_nova_cacerts_src: /var/lib/openstack/cacerts
+edpm_nova_cacerts_dest: /var/lib/openstack/cacerts/nova
+edpm_nova_tls_certs_enabled: "{{ tls_certs_enabled | default(False) }}"
--- a/roles/edpm_nova/meta/argument_specs.yml
+++ b/roles/edpm_nova/meta/argument_specs.yml
@@ -25,6 +25,36 @@ argument_specs:
         description: |
           The path to the directory where the nova config files
           will be rendered on the compute node.
+      edpm_nova_certs_src:
+        type: str
+        default: /var/lib/openstack/certs
+        description: |
+          The path to the directory containing the nova cert and key files
+          in the ansibleEE container. This is the directory
+          where all TLS certs and keys for the nova service are mounted.
+      edpm_nova_certs_dest:
+        type: str
+        default: /var/lib/openstack/certs/nova
+        description: |
+          The path to the directory where the nova cert and key files
+          will be rendered on the compute node.
+      edpm_nova_cacerts_src:
+        type: str
+        default: /var/lib/openstack/cacerts
+        description: |
+          The path to the directory containing the cacert files
+          in the ansibleEE container. This is the directory
+          where all cacert PEM files for the nova service are mounted.
+      edpm_nova_cacerts_dest:
+        type: str
+        default: /var/lib/openstack/cacerts/nova
+        description: |
+          The path to the directory where the cacert PEM files
+          for the nova service will be rendered on the compute node.
+      edpm_nova_tls_certs_enabled:
+        type: bool
+        default: false
+        description: Boolean to specify whether TLS certs are enabled.
       edpm_nova_compute_image:
         type: str
         default: "quay.io/podified-antelope-centos9/openstack-nova-compute:current-podified"
--- a/roles/edpm_nova/tasks/configure.yml
+++ b/roles/edpm_nova/tasks/configure.yml
@@ -15,6 +15,8 @@
   loop:
     - {"path": "{{ edpm_nova_config_dest }}", "mode": "0755"}
     - {"path": "/var/lib/openstack/config/containers", "mode": "0755"}
+    - {"path": "{{ edpm_nova_certs_dest }}", "mode": "0755"}
+    - {"path": "{{ edpm_nova_cacerts_dest }}", "mode": "0755"}
 - name: Create persistent directories
   tags:
     - configure
@@ -118,3 +120,20 @@
     mode: '0600'
     owner: nova
     group: nova
+
+- name: Copy TLS files to the compute node
+  tags:
+    - configure
+    - nova
+  become: true
+  loop:
+    - {"src": "{{ edpm_nova_certs_src }}/{{ inventory_hostname }}-tls.crt", "dest": "{{ edpm_nova_certs_dest }}/tls.crt"}
+    - {"src": "{{ edpm_nova_certs_src }}/{{ inventory_hostname }}-tls.key", "dest": "{{ edpm_nova_certs_dest }}/tls.key"}
+    - {"src": "{{ edpm_nova_cacerts_src }}/TLSCABundleFile", "dest": "{{ edpm_nova_cacerts_dest }}/TLSCABundleFile"}
+  ansible.builtin.copy:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    mode: '0600'
+    owner: nova
+    group: nova
+  when: edpm_nova_tls_certs_enabled
--- a/roles/edpm_nova/templates/nova_compute.json.j2
+++ b/roles/edpm_nova/templates/nova_compute.json.j2
@@ -10,6 +10,11 @@
     },
     "volumes": [
         "/var/lib/openstack/config/nova:/var/lib/kolla/config_files:ro",
+{% if edpm_nova_tls_certs_enabled %}
+	"/var/lib/openstack/certs/nova/tls.crt:/etc/pki/nova/server-cert.pem:ro",
+	"/var/lib/openstack/certs/nova/tls.key:/etc/pki/nova/private/server-key.pem:ro",
+	"/var/lib/openstack/cacerts/nova/TLSCABundleFile:/etc/pki/CA/cacert.pem:ro",
+{% endif %}
         "/etc/localtime:/etc/localtime:ro",
         "/lib/modules:/lib/modules:ro",
         "/dev:/dev",


fao89:
recheck

jpodivin (commented):
I think this is good to go, but I would like to give others a chance to take a look. If no one objects in 48 hrs, I'm giving it my LGTM.

SeanMooney:
ill try an review this later but we likely should add molecule coverage

roles/edpm_nova/defaults/main.yml
```diff
@@ -28,3 +28,9 @@ edpm_nova_image_download_retries: 5
 edpm_nova_config_src: /var/lib/openstack/configs
 edpm_nova_config_dest: /var/lib/openstack/config/nova
 edpm_nova_compute_image: "quay.io/podified-antelope-centos9/openstack-nova-compute:current-podified"
+
+# certs
+edpm_nova_certs_src: /var/lib/openstack/certs
+edpm_nova_certs_dest: /var/lib/openstack/certs/nova
+edpm_nova_cacerts_src: /var/lib/openstack/cacerts
+edpm_nova_cacerts_dest: /var/lib/openstack/cacerts/nova
```
SeanMooney, line 36: jsut some notes:
both src dirs are paths in the AEE pod
both dest dirs are paths on the EDPM host.

we use a single AEE pod for all hosts in the OpenStack dataplane nodeset

we do not want the copy the certs for all hosts in the nodeset to each host 
so we are going to need to make sure we only copy the relevant certs for this host when we use these later

roles/edpm_nova/tasks/configure.yml
```diff
@@ -118,3 +120,33 @@
...
+  set_fact: tls_certs_enabled="{{ lookup('env','tls_certs_enabled') }}"
+
+- name: copy cert file to {{edpm_nova_certs_dest}}
+  tags:
+    - configure
+    - nova
+  ansible.builtin.copy:
+    src: "{{ edpm_nova_certs_src }}/{{ inventory_hostname }}-tls.crt"
+    dest: "{{edpm_nova_certs_dest}}/tls.crt"
+  when: tls_certs_enabled
```
SeanMooney, line 134: so here we are using inventory_hostname to select the
relevant cert for this hsot and copy it to {{edpm_nova_certs_dest}}/tls.crt


so on each host the cert will be named 
tls.crt and it will be the cert for the current host +1

roles/edpm_nova/tasks/configure.yml
```diff
@@ -118,3 +120,33 @@
...
+
+- name: copy cacert file to {{edpm_nova_cacerts_dest}}
+  tags:
+    - configure
+    - nova
+  ansible.builtin.copy:
+    src: "{{ edpm_nova_cacerts_src }}/TLSCABundleFile"
+    dest: "{{edpm_nova_cacerts_dest}}/TLSCABundleFile"
+  when: tls_certs_enabled
```
SeanMooney, line 152: since the when condition is the same between these 3 copies
i would consider refactoring this into a loop

the other thing to consider is these files likely need to be owned by the nova user

which is created above.

the nova user on the host ahs the same uid and gid as the nova user in the container.

you can presumable use the same permissions as we are doing for the ssh key above

i.e. add  the following to each copy 
```
mode: '0600'
owner: nova
group: nova
```

roles/edpm_nova/templates/nova_compute.json.j2
```diff
@@ -10,6 +10,9 @@
     },
     "volumes": [
         "/var/lib/openstack/config/nova:/var/lib/kolla/config_files:ro",
+	"/var/lib/openstack/certs/nova/tls.crt:/etc/pki/nova/server-cert.pem:ro",
+	"/var/lib/openstack/certs/nova/tls.key:/etc/pki/nova/private/server-key.pem:ro",
+	"/var/lib/openstack/cacerts/nova/TLSCABundleFile:/etc/pki/CA/cacert.pem:ro",
```
SeanMooney, line 15: because we are bindmounting the certs directly instead of having kolla copy them we will need to ensure the permission on the files are correct as noted above.

if we were using kolla_start to copy the file then we would not need to do this.

roles/edpm_nova/tasks/configure.yml
```diff
@@ -15,6 +15,8 @@
   loop:
     - {"path": "{{ edpm_nova_config_dest }}", "mode": "0755"}
     - {"path": "/var/lib/openstack/config/containers", "mode": "0755"}
+    - { "path": "{{edpm_nova_cacerts_dest}}" }
```
jpodivin, line 18: These are also gong to need permissions.

jpodivin (changes requested):
Looking at this again there are some things to improve. The permissions for directories created specifically. There is also the matter of linting errors in the pre-commit.

roles/edpm_nova/tasks/configure.yml
```diff
@@ -118,3 +120,33 @@
     mode: '0600'
     owner: nova
     group: nova
+
+- name: get fact for tls_certs_enabled
+  set_fact: tls_certs_enabled="{{ lookup('env','tls_certs_enabled') }}"
```
bogdando, line 125: should we use lookup ('env','tls_certs_enabled', true) so that empty (unset) env considered false?

roles/edpm_nova/tasks/configure.yml
bogdando, line 134: |to_bool?

roles/edpm_nova/tasks/configure.yml
vakwetu: ack will refactor and add ownership

roles/edpm_nova/tasks/configure.yml
vakwetu: ack -- presumably set to false if not set.  This will allow older systems to continue to work

vakwetu:
recheck

vakwetu:
ah - I think this is failing here because its trying to pull in the depends-on - for which I need to fix the tests still to ensure that the required internal cas are present.

roles/edpm_nova/defaults/main.yml
SeanMooney: you shoudl add

edpm_nova_tls_certs_enabled:"{{ tls_certs_enabled | default(False) }}"

roles/edpm_nova/tasks/configure.yml
```diff
@@ -118,3 +119,24 @@
     mode: '0600'
     owner: nova
     group: nova
+
+- name: Get fact for tls_certs_enabled
+  ansible.builtin.set_fact:
+    tls_certs_enabled: "{{ lookup('env', 'tls_certs_enabled') | default(false, true) }}"
```
SeanMooney, line 125: I dont think this is correct we should not be using environmental variable in the AEE container to enable this

this should be enabled via ansible_vars which will be set in the inventory.

roles/edpm_nova/tasks/configure.yml
```diff
@@ -118,3 +119,24 @@
...
+    - {"src": "{{ edpm_nova_certs_src }}/{{ inventory_hostname }}-tls.crt", "dest": "{{ edpm_nova_certs_dest }}/tls.crt"}
+    - {"src": "{{ edpm_nova_certs_src }}/{{ inventory_hostname }}-tls.key", "dest": "{{ edpm_nova_certs_dest }}/tls.key"}
+    - {"src": "{{ edpm_nova_cacerts_src }}/TLSCABundleFile", "dest": "{{ edpm_nova_cacerts_dest }}/TLSCABundleFile"}
+  ansible.builtin.copy:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    mode: '0600'
+    owner: nova
+    group: nova
+  when: tls_certs_enabled|bool
```
SeanMooney, line 142: this should be edpm_nova_tls_certs_enabled

SeanMooney (changes requested):

roles/edpm_nova/tasks/configure.yml
vakwetu: done

roles/edpm_nova/tasks/configure.yml
vakwetu: done

roles/edpm_nova/tasks/configure.yml
vakwetu: no longer applicable

roles/edpm_nova/tasks/configure.yml
vakwetu: no longer applicable

roles/edpm_nova/tasks/configure.yml
vakwetu: done

roles/edpm_nova/templates/nova_compute.json.j2
vakwetu: done

roles/edpm_nova/defaults/main.yml
vakwetu: done

vakwetu:
recheck

vakwetu:
recheck

vakwetu:
recheck

NODE_FAILURE

vakwetu:
recheck

vakwetu:
recheck

roles/edpm_nova/meta/argument_specs.yml
```diff
@@ -25,6 +25,32 @@ argument_specs:
...
+        description: |
+          The path to the directory containing the cacert files
+          in the ansibleEE container. This is the directory
+          where all cacert PEM files for the nova service are mounted.
+      edpm_nova_cacerts_dest:
+        type: str
+        default: /var/lib/openstack/cacerts/nova
+        description: |
+          The path to the directory where the cacert PEM files
+          for the nova service will be rendered on the compute node.
```
SeanMooney, line 53: i would have expected edpm_nova_tls_certs_enabled to be required here too.

SeanMooney (commented):
i think this is generally good as is.
i have one question regrading argument_specs but if ci is ok with it its not a blocker.

roles/edpm_nova/meta/argument_specs.yml
vakwetu: done

rebtoor (approved):

rebtoor:
recheck

PR #495, opened 2023-11-15
Delegate fact gathering to container

jpodivin:


playbooks/install_os.yml
```diff
@@ -25,6 +25,7 @@
             fedora.linux{%- endif -%}
           _system_roles.timesync"
       delegate_to: localhost
+      delegate_facts: true
     - name: Install and configure time service using timesync system role
       ansible.builtin.include_role:
         name: "{{timesyncfqcn}}"
```
odyssey4me, line 31: on this task, have

  vars:
    timesyncfqcn: "{{ (ansible_facts['distribution'] == 'RedHat') | ternary('redhat.rhel', 'fedora.linux' }}_system_roles.timesync"

Then remove the set_fact task.

set_fact is a very slow way of achieving this, and introduces complexity in the approach

A further improvement would be to set these constants at build time and just use the constants... this is not really something that should be resolved at run time.

PR #497, opened 2023-11-16
Improved the robustness of `molecule_venv.sh` script

rebtoor:
- used versioned `python3` command instead of unversioned `python`
- added a check for the presence of the python development libraries and their installation in case of rpm based distro

### DIFF
--- a/scripts/molecule_venv.sh
+++ b/scripts/molecule_venv.sh
@@ -5,11 +5,21 @@ LOCAL_MOLECULE_ENV=$(mktemp -d)/molecule
 if which python3.9 2>/dev/null; then
     python3.9 -m venv $LOCAL_MOLECULE_ENV
 else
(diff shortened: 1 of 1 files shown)


scripts/molecule_venv.sh
```diff
@@ -5,11 +5,21 @@ LOCAL_MOLECULE_ENV=$(mktemp -d)/molecule
...
+    python3 -m venv $LOCAL_MOLECULE_ENV
+fi
+
+# some python modules need to be compiled
+if ! command -v python3-config 2>/dev/null; then
+    if -e /etc/redhat-release; then
+        sudo dnf -y install python3-devel
+    else
+        echo "Please, install python3 development package otherwise installation of modules could fail"
+    fi
```
fao89, line 18: sweet!

fao89 (approved):

PR #498, opened 2023-11-16
[edpm_nova]Pre-generate compute_id

gibizer:
Use the nova stable compute uuid feature.
* In greenfield pre-generate compute_id before the first start of the nova-compute service
* During adoption of old deployment preserve the compute_id if exists but make it ready only

Depends-on: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/223
Implements: https://issues.redhat.com/browse/OSPRH-117

### DIFF
--- a/roles/edpm_nova/tasks/configure.yml
+++ b/roles/edpm_nova/tasks/configure.yml
@@ -118,3 +118,58 @@
     mode: '0600'
     owner: nova
     group: nova
+
+- name: Copy TLS files to the compute node
+  tags:
+    - configure
+    - nova
+  become: true
+  loop:
+    - {"src": "{{ edpm_nova_certs_src }}/{{ inventory_hostname }}-tls.crt", "dest": "{{ edpm_nova_certs_dest }}/tls.crt"}
+    - {"src": "{{ edpm_nova_certs_src }}/{{ inventory_hostname }}-tls.key", "dest": "{{ edpm_nova_certs_dest }}/tls.key"}
+    - {"src": "{{ edpm_nova_cacerts_src }}/TLSCABundleFile", "dest": "{{ edpm_nova_cacerts_dest }}/TLSCABundleFile"}
+  ansible.builtin.copy:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    mode: '0600'
+    owner: nova
+    group: nova
+  when: edpm_nova_tls_certs_enabled
+
+- name: Check if compute_id exists in state_path
+  tags:
+    - configure
+    - nova
+  become: true
+  ansible.builtin.stat:
+    path: /var/lib/nova/compute_id
+  register: compute_id
+
+- name: Pre-generate compute_id if not already exists
+  tags:
+    - configure
+    - nova
+  become: true
+  ansible.builtin.copy:
+    dest: "/var/lib/nova/compute_id"
+    owner: nova
+    group: nova
+    mode: "0400"
+    attributes: "+i"
+    content: |
+      {{ lookup('ansible.builtin.pipe', 'uuidgen') }}
+  when: not compute_id.stat.exists
+
+- name: Make compute_id ready-only
+  tags:
+    - configure
+    - nova
+  become: true
+  ansible.builtin.file:
+    path: "/var/lib/nova/compute_id"
+    state: file
+    owner: nova
+    group: nova
+    mode: "0400"
+    attributes: "+i"
+  when: compute_id.stat.exists
--- a/roles/edpm_nova/molecule/default/verify.yml
+++ b/roles/edpm_nova/molecule/default/verify.yml
@@ -91,3 +91,16 @@
       ansible.builtin.assert:
         that:
           - "nova_authorized_keys.stat.mode == '0600'"
+
+    - name: Stat /var/lib/nova/compute_id
+      become: true
+      ansible.builtin.stat:
+        path: /var/lib/nova/compute_id
+      register: compute_id
+
+    - name: Assert that compute_id exists, not empty, and read-only
+      ansible.builtin.assert:
+        that:
+          - "compute_id.stat.size == 37"
+          - "compute_id.stat.mode == '0400'"
+          - "compute_id.stat.attr_flags == 'i'"


roles/edpm_nova/tasks/configure.yml
```diff
@@ -118,3 +118,70 @@
...
+  tags:
+    - configure
+    - nova
+  become: true
+  ansible.builtin.shell:
+    cmd: uuidgen > "{{ edpm_nova_config_dest }}/compute_id"
+  changed_when: true
+  when: not compute_id.stat.exists
+
+- name: Make compute_id ready only
```
SeanMooney, line 177: read-only

roles/edpm_nova/tasks/configure.yml
```diff
@@ -118,3 +118,70 @@
...
+- name: Delete compute_id from state_path
+  tags:
+    - configure
+    - nova
+    - adoption
+  become: true
+  ansible.builtin.file:
+    state: absent
+    path: "/var/lib/nova/compute_id"
+  when: old_compute_id.stat.exists
```
SeanMooney, line 156: so the copy and delete can just be a call to "mv"

SeanMooney (changes requested):
this also need a change to the nova container to extend the kolla config json to copy this to the /etc/nova dir in the contaier

roles/edpm_nova/tasks/configure.yml
```diff
@@ -118,3 +118,70 @@
...
+  tags:
+    - configure
+    - nova
+  become: true
+  ansible.builtin.file:
+    path: "{{ edpm_nova_config_dest }}/compute_id"
+    state: file
+    owner: nova
+    group: nova
+    mode: "0400"
```
bogdando, line 187: I would also call ``chattr +i`` in shell

roles/edpm_nova/tasks/configure.yml
rebtoor: or something like this

```
- name: Move compute_id from state_path to the config dir if present
  ansible.posix.synchronize:
    src: "/var/lib/nova/compute_id"
    dest: "{{ edpm_nova_config_dest }}/compute_id"
    rsync_opts:
      - "--remove-source-files"
```

roles/edpm_nova/tasks/configure.yml
gibizer: Good suggestion. Done.

roles/edpm_nova/tasks/configure.yml
gibizer: Done.

roles/edpm_nova/tasks/configure.yml
gibizer: I changed direction. I prefer not to move the compute_id file from the state_path (/var/lib/nova). But keep (or generate) it there and make it read-only. This greatly simplifies the logic and no adoption specific logic is needed.

roles/edpm_nova/tasks/configure.yml
```diff
@@ -118,3 +118,35 @@
...
+    path: /var/lib/nova/compute_id
+  register: compute_id
+
+- name: Pre-generate compute_id if not already exists
+  tags:
+    - configure
+    - nova
+  become: true
+  ansible.builtin.shell:
+    cmd: uuidgen > "/var/lib/nova/compute_id"
```
rebtoor, line 137: I had a bias against shell module, sorry :D 

But i guess you can replace two tasks with this one. What about:
```
    - name: Pre-generate compute_id if not already exists
      ansible.builtin.copy:
        dest: "/var/lib/nova/compute_id"
        owner: nova
        group: nova
        mode: "0400"
        attributes: "+i"
        content: |
          "{{ lookup('community.general.random_string') | to_uuid }}"
```

SeanMooney: hum honestly i think uuidgen is a better approach then that.
this can be the command module if you don't like shell

but you have made the ansible massively more complicated 

we also still need to keep the "name: Make compute_id ready-only"
later so setting the permission here is kind of redundant.

gibizer: > we also still need to keep the "name: Make compute_id ready-only"
later so setting the permission here is kind of redundant.

Yes, we need to keep the file permission settings independently as the file might exists during adoption but we want to still change its permissions.

rebtoor: @SeanMooney i'm fine with `shell` module if it's unavoidable or makes the life easier.
(can't be `command` since you need to use redirection, which is unsupported on that module)

SeanMooney: `
    - name: Pre-generate compute_id if not already exists
      tags:
        - configure
        - nova
      become: true
      ansible.builtin.copy:
        dest: "/var/lib/nova/compute_id"
        owner: nova
        group: nova
        mode: "0400"
        attributes: "+i"
        content: |
          "{{ lookup('ansible.builtin.pipe', 'uuidgen') }}"
      when: not compute_id.stat.exists
`
how about this

bogdando: if only that doesn't introduce unexpected newlines and closing/wrapping lines

roles/edpm_nova/tasks/configure.yml
```diff
@@ -118,3 +118,40 @@
...
+  become: true
+  ansible.builtin.copy:
+    dest: "/var/lib/nova/compute_id"
+    owner: nova
+    group: nova
+    mode: "0400"
+    attributes: "+i"
+    content: |
+      {{ lookup('ansible.builtin.pipe', 'uuidgen') }}
+  when: not compute_id.stat.exists
```
SeanMooney, line 144: @gibi so one thing to keep in mind is this will break adoption with upsream nova, and downstream until we complete the backprot of the 17.1 change to create the file.

that should be in 17.1.2 i think or 17.1.3

for now with adoption we can simply add a step to create this file with the correct compute service id but i don't know what to do about rdo.

we either need to backport this in upstream nova or our downstream backport need to be put into rdo too

gibizer: /o\ you are right it will break adoption

gibizer: I think the way forward is to add an adoption step as from ansible we cannot get the existing UUID.

jpodivin (commented):
Compatibility with nova aside, this is some good work.

gibizer:
> this is unblocked now as we've merged adoption FFU (but we need to test this change in a DNM adoption patches as a depends-on)

I need to push an adoption specific change as the adoption from state (17.1) does not have the compute_id yet (we have a downstream backport for that feature)

gibizer:
> > this is unblocked now as we've merged adoption FFU (but we need to test this change in a DNM adoption patches as a depends-on)
> 
> I need to push an adoption specific change as the adoption from state (17.1) does not have the compute_id yet (we have a downstream backport for that feature)

The temporary adoption fix is https://github.com/openstack-k8s-operators/data-plane-adoption/pull/223

gibizer:
/unhold 
https://github.com/openstack-k8s-operators/data-plane-adoption/pull/223 has been merged

roles/edpm_nova/tasks/configure.yml
```diff
@@ -137,3 +137,40 @@
...
+
+- name: Make compute_id ready-only
```
slagle, line 165: is there a reason this task is needed given the previous task also sets all the same file properties?

bogdando: this extends the above lines for cases when this file is already exists and is not being created

slagle: oh ok...i would suggest adding that "when".

roles/edpm_nova/tasks/configure.yml
gibizer: Adoption step is merged

roles/edpm_nova/tasks/configure.yml
gibizer: Done.

gibizer:
rebased to resolve merge conflict and fixed up a nit

gibizer:
recheck 

```
TASK [tempest : Get keystone data _raw_params=oc get keystoneapi keystone -n openstack -o json] ***
Monday 15 January 2024  07:28:18 -0500 (0:00:00.057)       0:37:14.616 ******** 
fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["oc", "get", "keystoneapi", "keystone", "-n", "openstack", "-o", "json"], "delta": "0:00:00.183609", "end": "2024-01-15 07:28:18.582124", "msg": "non-zero return code", "rc": 1, "start": "2024-01-15 07:28:18.398515", "stderr": "The connection to the server api.crc.testing:6443 was refused - did you specify the right host or port?", "stderr_lines": ["The connection to the server api.crc.testing:6443 was refused - did you specify the right host or port?"], "stdout": "", "stdout_lines": []}
```

gibizer:
recheck retry limit

slagle (approved):

PR #500, opened 2023-11-20
Added support for configuring telemetry logging

jlarriba:


playbooks/logging.yml
```diff
@@ -0,0 +1,29 @@
+---
```
rebtoor, line 1: maybe `telemetry-logging` or `enable-compute-logging` would be more appropriate for the playbook name?

`logging` seems too generic.

roles/edpm_telemetry/molecule/default/prepare.yml
```diff
@@ -200,3 +200,16 @@
...
+      ansible.builtin.copy:
+        dest: /etc/rsyslog.d/10-telemetry.conf
+        mode: '0644'
+        owner: root
+        group: root
+        content: |
+          module(load="imjournal" PersistStateInterval="10"
+                StateFile="journal_state")
+
+          action(type="omfwd" target="172.17.0.80" port="10514" protocol="tcp"
```
rebtoor, line 213: I don't know if i like these parameters hardcoded here. What about define them as vars with these values as defaults?

roles/edpm_telemetry/tasks/enable_logging.yml
```diff
@@ -0,0 +1,33 @@
...
+    remote_src: "{{ telemetry_test | default('false') }}"
+  tags:
+    - edpm_logging
+
+- name: Restart rsyslog
+  become: true
+  ansible.builtin.systemd:
+    state: restarted
+    name: rsyslog.service
+  tags:
```
rebtoor, line 32: Since this task file is called only by the above mentioned playbook, I'd change this task to be an handler and I'd add the directive:

` notify: Restart rsyslog` 

to the copy task above.

rebtoor (changes requested):

roles/edpm_telemetry/molecule/default/prepare.yml
jlarriba: This is preparing everything for the test. The file will be configured because it will be copied directly from the control plane to the compute node. What I am trying to do here is to have a valid file that wont fail during the test. The values here are irrelevant. I don't really see the need to move them to variables, they would only be testing variables for some feature (writing into a remote server) that wont be tested as it is out of ansible scope.

roles/edpm_telemetry/tasks/enable_logging.yml
jlarriba: This is a very good idea, but as we are using import_tasks to use the tasks in the role, I fail to see where would the handlers be defined. I also can't found an example of handlers usage in the whole edpm-ansible.

Could you please point me in the right direction here?

roles/edpm_telemetry/molecule/default/prepare.yml
rebtoor: sorry @jlarriba , my bad... i missed the fact that it was in molecule testing :)

rebtoor (approved):

jlarriba:
recheck

jlarriba:
recheck

jlarriba:
recheck

fao89 (approved):

PR #501, opened 2023-11-20
Execute `edpm_bootstrap_command` during `download_cache`

rebtoor:
Since the registration of the EL nodes is performed during the `edpm_bootstrap_command` task in the `edpm_bootstrap` role, we moved that task into a new task file called `bootstrap_command.yml` so it can be called by the `edpm_download_cache` role without invoking the entire role. We also added a marker that will avoid to execute the task twice if it's been already executed.

### DIFF
--- a/roles/edpm_bootstrap/defaults/main.yml
+++ b/roles/edpm_bootstrap/defaults/main.yml
@@ -77,3 +77,7 @@ edpm_bootstrap_swap_partition_label: swap1
 # Shell command that is executed before any packages are installed by the role.
 # Can be used to register systems using any arbitrary registration command(s).
 edpm_bootstrap_command: ""
+
+# The path of the file that will be used as a marker when the edpm_bootstrap_command
+# has been executed on the node.
+edpm_bootstrap_command_done_file: /.bootstrap_command_done
--- a/roles/edpm_bootstrap/meta/argument_specs.yml
+++ b/roles/edpm_bootstrap/meta/argument_specs.yml
@@ -94,3 +94,10 @@ argument_specs:
         description: |
           Shell command that is executed before any packages are installed by the role.
           Can be used to register systems using any arbitrary registration command(s).
+
+      edpm_bootstrap_command_done_file:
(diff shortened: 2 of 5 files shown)


rebtoor:
recheck

`NODE_FAILURE`

rebtoor:
recheck

roles/edpm_bootstrap/defaults/main.yml
```diff
@@ -77,3 +77,5 @@ edpm_bootstrap_swap_partition_label: swap1
 # Shell command that is executed before any packages are installed by the role.
 # Can be used to register systems using any arbitrary registration command(s).
 edpm_bootstrap_command: ""
+
+edpm_bootstrap_command_done_file: /.bootstrap_command_done
```
jpodivin, line 81: I know it's in arg spec, but it still ought to have a comment.

roles/edpm_bootstrap/meta/argument_specs.yml
```diff
@@ -94,3 +94,10 @@ argument_specs:
         description: |
           Shell command that is executed before any packages are installed by the role.
           Can be used to register systems using any arbitrary registration command(s).
+
+      edpm_bootstrap_command_done_file:
+        type: str
+        default: /.bootstrap_command_done
+        description: |
+          The name of the file that will be used as a marker when the edpm_bootstrap_command
```
jpodivin, line 102: With the leading slash it's a path. It should have the `path` type and the description should match it.

jpodivin (changes requested):

roles/edpm_bootstrap/defaults/main.yml
rebtoor: :+1:

roles/edpm_bootstrap/meta/argument_specs.yml
rebtoor: :+1:

jpodivin (approved):

PR #502, opened 2023-11-20
[DNM] test for broken playbook behaviour

rebtoor:
Depends-On: https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/270


rebtoor:
not needed anymore, the problem was elsewhere

PR #504, opened 2023-11-22
Set some neutron-ovn-metadata agent options only if necessary

slawqo:
Some metadata agent config options are usually provided by the nova and ovn operators and are set in separate config files.
To avoid confusion with empty values in the main config file used by the neutron-ovn-metadata-agent, those options will be set there only if the are provided in the role's parameters.

### DIFF
--- a/roles/edpm_neutron_metadata/templates/neutron-ovn-metadata-agent.conf.j2
+++ b/roles/edpm_neutron_metadata/templates/neutron-ovn-metadata-agent.conf.j2
@@ -1,8 +1,18 @@
 [DEFAULT]
 debug = {{ edpm_neutron_metadata_agent_DEFAULT_debug }}
+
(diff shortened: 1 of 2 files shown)


roles/edpm_neutron_metadata/templates/neutron-ovn-metadata-agent.conf.j2
```diff
@@ -1,8 +1,18 @@
 [DEFAULT]
 debug = {{ edpm_neutron_metadata_agent_DEFAULT_debug }}
+
+{%- if edpm_neutron_metadata_agent_DEFAULT_nova_metadata_host is defined %}
```
karelyatin, line 4: {%- --> {% seems to be causing it, can drop from all the places in beginning of the block

slawqo: Done

karelyatin (commented):
/lgtm

PR #467, opened 2023-10-19
host install libvirt

SeanMooney:
refactor libvirt to be host installed

Closes: [OSPRH-960](https://issues.redhat.com//browse/OSPRH-960)
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/910

PR #506, opened 2023-11-28
Add libvirt certs

vakwetu:
This follows the same logic as for the nova compute certs.
It does not include any config changes to actually use the certs.  I'm expecting the compute DFG folks will add this logic.

### DIFF
--- a/roles/edpm_libvirt/defaults/main.yml
+++ b/roles/edpm_libvirt/defaults/main.yml
@@ -51,3 +51,8 @@ edpm_libvirt_packages:
   # for qemu and the ceph udev rules
   - ceph-common
 edpm_libvirt_ceph_path: /var/lib/openstack/config/ceph
+
+# certs
+edpm_libvirt_certs_src: /var/lib/openstack/certs
+edpm_libvirt_cacerts_src: /var/lib/openstack/cacerts
+edpm_libvirt_tls_certs_enabled: "{{ edpm_tls_certs_enabled | default(False) }}"
--- a/roles/edpm_libvirt/tasks/configure.yml
+++ b/roles/edpm_libvirt/tasks/configure.yml
@@ -15,6 +15,9 @@
   loop:
     - {"path": "/etc/tmpfiles.d/", "owner": "root", "group": "root"}
     - {"path": "/var/lib/edpm-config/firewall", "owner": "root", "group": "root"}
+    - {"path": "/etc/pki/libvirt", "owner": "root", "group": "root"}
+    - {"path": "/etc/pki/libvirt/private", "owner": "root", "group": "root"}
+    - {"path": "/etc/pki/CA", "owner": "root", "group": "root"}
 
 - name: Render libvirt config files
   tags:
@@ -86,6 +89,7 @@
   register: run_libvirt
   failed_when: run_libvirt.rc not in [0, 1]
   changed_when: false
+
 - name: Restore selinux label for /run/libvirt to virt_var_run_t
   tags:
     - configure
@@ -94,3 +98,20 @@
   ansible.builtin.command: "restorecon -R /run/libvirt"
   changed_when: true
   when: run_libvirt.rc == 0
+
+- name: Copy TLS files to the compute node
+  tags:
+    - configure
+    - libvirt
+  become: true
+  loop:
+    - {"src": "{{ edpm_libvirt_certs_src }}/{{ inventory_hostname }}-tls.crt", "dest": "/etc/pki/libvirt/server-cert.pem"}
+    - {"src": "{{ edpm_libvirt_certs_src }}/{{ inventory_hostname }}-tls.key", "dest": "{/etc/pki/libvirt/private/server-key.pem"}
+    - {"src": "{{ edpm_libvirt_cacerts_src }}/TLSCABundleFile", "dest": "/etc/pki/CA/cacert.pem"}
+  ansible.builtin.copy:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    mode: '0600'
+    owner: "root"
+    group: "root"
+  when: edpm_libvirt_tls_certs_enabled
--- a/roles/edpm_libvirt/meta/argument_specs.yml
+++ b/roles/edpm_libvirt/meta/argument_specs.yml
@@ -40,3 +40,21 @@ argument_specs:
         type: str
         description: The path to the ceph config files.
         default: "/var/lib/openstack/config/ceph"
+      edpm_libvirt_certs_src:
+        type: str
+        default: /var/lib/openstack/certs
+        description: |
+          The path to the directory containing the libvirt cert and key files
+          in the ansibleEE container. This is the directory
+          where all TLS certs and keys for the libvirt service are mounted.
+      edpm_libvirt_cacerts_src:
+        type: str
+        default: /var/lib/openstack/cacerts
+        description: |
+          The path to the directory containing the cacert files
+          in the ansibleEE container. This is the directory
+          where all cacert PEM files for the libvirt service are mounted.
+      edpm_libvirt_tls_certs_enabled:
+        type: bool
+        default: false
+        description: Boolean to specify whether TLS certs are enabled.


fao89:
recheck

roles/edpm_libvirt/tasks/configure.yml
```diff
@@ -64,3 +66,21 @@
...
+  become: true
+  loop:
+    - {"src": "{{ edpm_libvirt_certs_src }}/{{ inventory_hostname }}-tls.crt", "dest": "{{ edpm_libvirt_certs_dest }}/tls.crt"}
+    - {"src": "{{ edpm_libvirt_certs_src }}/{{ inventory_hostname }}-tls.key", "dest": "{{ edpm_libvirt_certs_dest }}/tls.key"}
+    - {"src": "{{ edpm_libvirt_cacerts_src }}/TLSCABundleFile", "dest": "{{ edpm_libvirt_cacerts_dest }}/TLSCABundleFile"}
+  ansible.builtin.copy:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    mode: '0600'
+    owner: "{{ item.owner | default(ansible_user) }}"
```
olliewalsh, line 84: would it be safer to force root ownership?

SeanMooney: in a containerized install these should be owned by root:libvirt
but in a decontaienrsied install then yes root:root

i belive root:root would work in the containerised install too so that is probably a better default in general for these files.

roles/edpm_libvirt/templates/libvirt_virtlogd/libvirt_virtlogd.json.j2
```diff
@@ -8,6 +8,11 @@
         "KOLLA_CONFIG_STRATEGY":"COPY_ALWAYS"
     },
     "volumes": [
+{% if edpm_libvirt_tls_certs_enabled %}
```
SeanMooney, line 11: i need to look back at the nova changes i guess but one thing we are trying to avoid if possible is making this list dynamic


we would generally prefer to always passthrough the folder even if its empty

so i would generally prefer if this was just 

/var/lib/openstack/certs/libvirt/:/etc/pki/libvirt/:ro

and if we set the filet to have the correct name instead when we copy it

were tlse is not enabled it would just end up being an empty dir.

roles/edpm_nova/templates/nova_compute.json.j2
```diff
@@ -10,6 +10,11 @@
     },
     "volumes": [
         "/var/lib/openstack/config/nova:/var/lib/kolla/config_files:ro",
+{% if edpm_nova_tls_certs_enabled %}
```
SeanMooney, line 13: same comment ehre i woudl prefer to kee this static

SeanMooney (changes requested):
99% of this shoudl be the same i think regardless of if we have contaierised or decontaierised libvirt

bogdando:
We should probably land it after https://github.com/openstack-k8s-operators/edpm-ansible/pull/467

roles/edpm_libvirt/tasks/configure.yml
```diff
@@ -94,3 +98,20 @@
...
   changed_when: true
   when: run_libvirt.rc == 0
+
+- name: Copy TLS files to the compute node
+  tags:
+    - configure
+    - libvirt
+  become: true
+  loop:
+    - {"src": "{{ edpm_libvirt_certs_src }}/{{ inventory_hostname }}-tls.crt", "dest": "/etc/pki/libvirt/server-cert.pem"}
```
slagle, line 108: don't the certs need to be under a ''libvirt'' subdirectory in case there are different certs for different services?

roles/edpm_libvirt/defaults/main.yml
```diff
@@ -51,3 +51,8 @@ edpm_libvirt_packages:
   # for qemu and the ceph udev rules
   - ceph-common
 edpm_libvirt_ceph_path: /var/lib/openstack/config/ceph
+
+# certs
+edpm_libvirt_certs_src: /var/lib/openstack/certs
+edpm_libvirt_cacerts_src: /var/lib/openstack/cacerts
+edpm_libvirt_tls_certs_enabled: "{{ tls_certs_enabled | default(False) }}"
```
slagle, line 58: This got changed to ``edpm_tls_certs_enabled`` right?

roles/edpm_libvirt/tasks/configure.yml
SeanMooney: so technially yes. nova and libvirt are just going to use the libvirt cert as we don't have any endpoint on the nova-compute agent but nova-compute need to talk to libvirt so it makes sense for nova and livbirt to share the same certs.

for ovn/ovs or other things like iscisd/multipathd they should likely have spereate certs so it would make sense to move these to a libvirt subdir in the src configmap.

roles/edpm_libvirt/tasks/configure.yml
SeanMooney: we have now decontaienrised libvirt so ya root:root would make more sense.

roles/edpm_libvirt/defaults/main.yml
vakwetu: good catch - yes - will fix

roles/edpm_libvirt/tasks/configure.yml
vakwetu: Just to clarify here, the destination obviously will be in /etc/pki/libvirt/... so thats in a libvirt directory.

The source is where the certs are mounted in the ansibleEE pod, and the ansibleEE pods are created per dataplane service.  That is, there is an ansibleEE pod for nova, and one for libvirt etc.  The certs needed in that pod are not differentiated per service (ie. they are under /var/lib/openstack/certs) because there are only ever certs for that one dataplane service

Does that make sense?

roles/edpm_libvirt/tasks/configure.yml
vakwetu: ack will fix

roles/edpm_libvirt/tasks/configure.yml
slagle: It's correct, but I still think they should be mounted under a subdir with the service name, same as we do for the ConfigMaps and Secrets:

https://github.com/openstack-k8s-operators/dataplane-operator/blob/8d8a1051288a065e8fec16e07c907dc08ebbb4d8/pkg/deployment/deployment.go#L232

bogdando: introducing a dedicated role for certs managment sound like a good idea to me

vakwetu: @slagle ok - I can change it accordingly.

@bogdando  @SeanMooney  The certs that nova and other services would need would be there in /var/lib/openstack/certs/{service} as long as those previous services have already executed.  Each service  copies the certs that are mounted to a specific directory which is then mounted (or not, in the case of libvirt) by the service container.  But this depends on an implied ordering (which probably already existed before TBH).  A dedicated dataplane service for cert management will make this all more predictable.

So, just to clarify, with the new dedicated dataplane service for cert management (lets call it "certs"), we will:
1. mount the secrets containing the certs/keys/cacerts for all nodes for all the enabled services that require certs
2. execute an ansible role that will copy the relevant certs/keys/cacerts  to /var/lib/openstack/certs/{service} or /var/lib/openstack/cacerts/{service} on each node. In the libvirt case, we'd move it to /etc/pki/libvirt etc.
3. each subsequent dataplane service can then use those certs as appropriate.  Most of them will simply mount the required certs in a container.   This simplifies the logic in each dataplane service to just doing a mount.

Is this what you guys had in mind?

SeanMooney: yep more or less
so for nova we would just mount it to the nova contaienrs
for libvirt we we just need to copy it to the correct location or update the libvirt config to know where it is.

so if just haveing the libvirt role copy/move the certs form

 /var/lib/openstack/cacerts/{libvirt} to /etc/pki/libvirt is all that is required we can have the libvirt role do that

jpodivin (approved):

PR #508, opened 2023-11-28
Set edpm_sshd_configure_firewall true by default

bshephar:


roles/edpm_sshd/defaults/main.yml
```diff
@@ -58,7 +58,7 @@ edpm_sshd_server_options:
   Subsystem: 'sftp /usr/libexec/openssh/sftp-server'
 
 # Firewall configuration part
-edpm_sshd_configure_firewall: false
+edpm_sshd_configure_firewall: true
```
rebtoor, line 61: Overall i couldn't agree more on this, but if a user forgets to populate `edpm_sshd_allowed_ranges` he will lock himself out of compute nodes and the deployment will fail.

So, if we want to enable this by default we should:

- provide a default "0.0.0.0/0" to `edpm_sshd_allowed_ranges` (which is just like disabling firewall, so it's useless :) )
- declare `edpm_sshd_allowed_ranges` as a required var that can't be empty.

rebtoor (approved):

rebtoor:
recheck

```
"fatal: [localhost]: FAILED! => {\"changed\": false, \"msg\": \"Failure downloading https://golang.org/dl/go1.19.10.linux-amd64.tar.gz, Request failed: <urlopen error [Errno -2] Name or service not known>\"}",
```

PR #510, opened 2023-11-29
Fixing nova permission denied issues

Jaganathancse:
Getting permission denied error for directory
'/var/lib/openstack/config' with user 'cloud-admin'.
This issue is due to the provolege configuration
for '/var/lib/openstack/config' dir is '0750' in
neutron-sriov role.

Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/540

### DIFF
--- a/roles/edpm_neutron_sriov/molecule/default/converge.yml
+++ b/roles/edpm_neutron_sriov/molecule/default/converge.yml
@@ -17,6 +17,7 @@
 - name: Converge
   hosts: all
   roles:
-    - role: "edpm_neutron_sriov"
+    - role: osp.edpm.edpm_neutron_sriov
       vars:
         edpm_neutron_sriov_agent_config_src: "{{lookup('env', 'MOLECULE_SCENARIO_DIRECTORY')}}/test-data"
+        ansible_user: root
--- a/roles/edpm_neutron_sriov/tasks/install.yml
+++ b/roles/edpm_neutron_sriov/tasks/install.yml
@@ -15,15 +15,18 @@
 # under the License.
 
 - name: Create neutron-sriov-agent directories
+  become: true
   ansible.builtin.file:
     path: "{{ item.path }}"
     setype: "container_file_t"
     state: directory
+    owner: "{{ item.owner | default(omit) }}"
+    group: "{{ item.group | default(omit) }}"
     mode: "{{ item.mode | default(omit) }}"
   loop:
-    - {'path': "/var/lib/openstack/config/containers", "mode": "0750"}
+    - {'path': "/var/lib/openstack/config/containers", "mode": "0755", "owner": "{{ ansible_user }}", "group": "{{ ansible_user }}"}
     - {'path': "/var/lib/neutron", "mode": "0750"}
-    - {'path': "{{ edpm_neutron_sriov_agent_config_dir }}", "mode": "0755"}
+    - {'path': "{{ edpm_neutron_sriov_agent_config_dir }}", "mode": "0755", "owner": "{{ ansible_user }}", "group": "{{ ansible_user }}"}
     - {'path': "/var/log/containers/stdouts"}
     - {'path': "/var/log/containers/neutron"}
   tags:
@@ -31,7 +34,6 @@
     - neutron
 
 - name: Render neutron-sriov-agent container
-  become: true
   ansible.builtin.template:
     src: "neutron_sriov_agent.yaml.j2"
     dest: "/var/lib/openstack/config/containers/neutron_sriov_agent.yaml"
--- a/roles/edpm_nova/tasks/install.yml
+++ b/roles/edpm_nova/tasks/install.yml
@@ -1,7 +1,6 @@
 ---
 
 - name: Render nova container
-  become: true
   tags:
     - install
     - nova


slagle (commented):
This is good to have where you've added it, thanks!

 Should we also set `become: true `in `playbooks/nova.yml`? Most (if not all) of the other playbooks have that set. @gibizer , @SeanMooney  WDYT?

gibizer:
Did this PR trying to solve an actual permission issue seen in some deployment? I'm wondering as I did not see such issue before.
In general I'm OK to add the `become: true` flags if they are needed.

gibizer:
I guess `become:true` in the playbook would replace the all the other become true in the roles. Am I correct?

slagle:
> I guess `become:true` in the playbook would replace the all the other become true in the roles. Am I correct?

It would mean all tasks run as a result of running the playbook would run with `become: true`. Personally, I don't see a need to remove it from individual tasks as well.

SeanMooney:
> > I guess `become:true` in the playbook would replace the all the other become true in the roles. Am I correct?
> 
> It would mean all tasks run as a result of running the playbook would run with `become: true`. Personally, I don't see a need to remove it from individual tasks as well.

in general it think we should limit become:true to blocks or tasks and never use it for playbooks.

roles/edpm_nova/tasks/configure.yml
```diff
@@ -39,6 +39,7 @@
     - {"path": "/var/lib/iscsi"}
     - {"path": "/run/openvswitch"}
 - name: Render nova config files
+  become: true
```
SeanMooney, line 42: this intentionally does not have become true as it should not be required if the permission are correct on the dirs so we should not proceed with this
until we identify what the regression was and fix that.

SeanMooney: on line 16 we are doing

  become: true
  ansible.builtin.file:
    path: "{{ item.path }}"
    state: "directory"
    setype: "container_file_t"
    owner: "{{ item.owner | default(ansible_user) }}"
    group: "{{ item.group | default(ansible_user) }}"
    mode: "{{ item.mode | default(omit) }}"
  loop:
- {"path": "{{ edpm_nova_config_dest }}", "mode": "0755"}

i.e. we are creating edpm_nova_config_dest with 755 permission and the ownership is set to the current ansible user

we should be able to rewite to edpm_nova_config_dest without become 
when we invoke ansible.builtin.template below on line 46

SeanMooney (changes requested):
do you have a log or info about when this permisison issue happens as it sound like something else is modifying the ownership or permsioins of /var/lib/openstack/nova or more spelcicaly edpm_nova_config_dest

Jaganathancse:
> do you have a log or info about when this permisison issue happens as it sound like something else is modifying the ownership or permsioins of /var/lib/openstack/nova or more spelcicaly edpm_nova_config_dest

Mentioned above,  i am getting this error when trying with SRIOV and did not  get for OVN DPDK.

roles/edpm_nova/tasks/configure.yml
rebtoor: @Jaganathancse can you do a `sudo grep denied /var/log/audit/audit.log` in the `edpm-compute-0` node?

SeanMooney: can you provide the content of the config map you are defining the custom nova.conf  and also how sriov is being configured in general.

we have not actually started testing/enabling sriov or dpdk yet.
so you are trying to do something that we don't plan to enable until at least q1 next year. we expect it to work or almost work today but no one has done it yet so its not nessirarly expected to work today.

we should not need to add become here but we may need other changes in the role to make it work

/var/lib/openstack/config/nova/config.json should be owned by the current user
so you should not be getting  a permission denied and then only role that is allowed to modify that or the /var/lib/openstack/config/nova directory in general is the nova role can you confirm that you have not modified it in any way.

please provide "ls -alZ /var/lib/openstack/config/nova"  in addition to the audit log info.

jpodivin (changes requested):
This change needs to be tied to some sort of report about an issue experienced. In general terms, we can't really fix something we know little about.

But that is besides the point. The bigger problem is that become at this level is too powerful of a solution. Everything running under this playbook will now run with all powers of root. That's fine, as long as everybody remembers, and treats the imported roles with care. But people forget.

If there is a permissions issue, it should be solved on the lowest level possible, that is a task, or a block. Not an entire play.

roles/edpm_nova/tasks/configure.yml
Jaganathancse: > sudo grep denied /var/log/audit/audit.log

[cloud-admin@tigon26 ~]$ sudo grep denied /var/log/audit/audit.log
type=AVC msg=audit(1703236288.332:30): avc:  denied  { search } for  pid=2354 comm="modprobe" name="events" dev="tracefs" ino=4099 scontext=system_u:system_r:openvswitch_load_module_t:s0 tcontext=system_u:object_r:tracefs_t:s0 tclass=dir permissive=1
type=AVC msg=audit(1703237550.644:26): avc:  denied  { search } for  pid=2111 comm="modprobe" name="events" dev="tracefs" ino=1040 scontext=system_u:system_r:openvswitch_load_module_t:s0 tcontext=system_u:object_r:tracefs_t:s0 tclass=dir permissive=0
type=AVC msg=audit(1703237550.644:26): avc:  denied  { search } for  pid=2111 comm="modprobe" name="events" dev="tracefs" ino=1040 scontext=system_u:system_r:openvswitch_load_module_t:s0 tcontext=system_u:object_r:tracefs_t:s0 tclass=dir permissive=0

Jaganathancse: > /var/lib/openstack/config/nova/config.json
@SeanMooney 
nova config file is not generated on my env,
[cloud-admin@tigon26 ~]$ ls /var/lib/openstack/config/nova/config.json 
ls: cannot access '/var/lib/openstack/config/nova/config.json': Permission denied
[cloud-admin@tigon26 ~]$ sudo ls /var/lib/openstack/config/nova/config.json 
ls: cannot access '/var/lib/openstack/config/nova/config.json': No such file or directory

Jaganathancse: > "ls -alZ /var/lib/openstack/config/nova

[cloud-admin@tigon26 ~]$ ls -alZ /var/lib/openstack/config/nova
ls: cannot access '/var/lib/openstack/config/nova': Permission denied

rebtoor: sudo?

Jaganathancse: > sudo?

[cloud-admin@tigon26 ~]$ sudo ls -alZ /var/lib/openstack/config/nova
total 0
drwxr-xr-x. 2 cloud-admin cloud-admin unconfined_u:object_r:container_file_t:s0  6 Dec 22 09:43 .
drwxr-x---. 5 root        root        unconfined_u:object_r:container_file_t:s0 48 Dec 22 09:43 ..

SeanMooney: so that is say the permisions are wrong for  /var/lib/openstack/config/nova
drwxr-x--- is not 755 its 750

SeanMooney: line 16 should have fixed that are you perhaps connecting as root instead of cloud-admin for some reason?

we should not be logging into the comptue as root when executing the ansibel roles

SeanMooney: by the way on your system if you just sudo rm -rf /var/lib/openstack/config/nova and reexecut the ansibel it should resolve your current issue

Jaganathancse: > line 16 should have fixed that are you perhaps connecting as root instead of cloud-admin for some reason?
> 
> we should not be logging into the comptue as root when executing the ansibel roles

updated directory permission 755.

roles/edpm_nova/tasks/configure.yml
```diff
@@ -31,7 +31,7 @@
     - {"path": "/var/lib/nova", "mode": "0755"}
     - {"path": "/var/lib/_nova_secontext", "mode": "0755"}
     - {"path": "/var/lib/nova/instances", "mode", "0755"}
-    - {"path": "/var/log/containers/nova", "mode": "0750"}
+    - {"path": "/var/log/containers/nova", "mode": "0755"}
```
SeanMooney, line 34: This should not be required, this folder will be owned by nova when the container starts https://github.com/openstack-k8s-operators/edpm-ansible/blob/c1344d96f00559a34794dedea489f6fe131b84fa/roles/edpm_nova/templates/config.json.j2#L38

We could premtivly change the owner to nova now that we create that user on the host but 755 is only needed for traversability of nested dirs. What requires that in this case?

SeanMooney (changes requested):
We could change to 755 but I would like to see an explanation of why this is needed and a module test to enforce it.

jpodivin (commented):
I appreciate that the become has been removed, and that the problem is now addressed on low level. However @SeanMooney has a point that this shouldn't be necessary given the directory ownership. Which is user is trying to access the dir?

roles/edpm_nova/tasks/configure.yml
Jaganathancse: @jpodivin @SeanMooney 

In latest code, getting permission denied error due to config dir under path '/var/lib/openstack'.

config dir user is root and other user privilege  permission is 0. so we are not able to create any file inside '/var/lib/openstack/nova' dir in cloud-admin user. so need to use root permission for Render nova config files tasks.
or we need to change other user permission for config dir to 'x-r'.

[cloud-admin@edpm-compute-0 ~]$ ls -all /var/lib/openstack/
total 4
drwxr-xr-x.  6 root        root          71 Jan 10 08:30 .
drwxr-xr-x. 49 root        root        4096 Jan 10 08:30 ..
drwxr-xr-x.  3 cloud-admin cloud-admin   18 Jan 10 08:30 cacerts
drwxr-xr-x.  3 cloud-admin cloud-admin   18 Jan 10 08:30 certs
drwxr-x---.  5 root        root          48 Jan 10 08:30 config
drwxr-xr-x.  2 root        root          25 Jan 10 08:16 reboot_required
[cloud-admin@edpm-compute-0 ~]$ ls -all /var/lib/openstack/config/
ls: cannot open directory '/var/lib/openstack/config/': Permission denied

roles/edpm_nova/tasks/install.yml
```diff
@@ -1,7 +1,6 @@
 ---
 
 - name: Render nova container
-  become: true
```
SeanMooney, line 4: oh ya that an existg bug from when i initall ported this

it should not use become here +1

roles/edpm_neutron_sriov/tasks/install.yml
```diff
@@ -15,13 +15,16 @@
 # under the License.
 
 - name: Create neutron-sriov-agent directories
+  become: true
```
SeanMooney, line 18: +1
ideally you would also remove the become true here

https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/playbooks/neutron_sriov.yml#L6

the neutron role should elevate in the tasks that need it now at the playbook level

the same issue is present in all 4 neutron playbooks
this seams to have been cargo culted to all the neutron playbooks from the ovn playbook but since its a preexsitign bug it could be fixed in  different pr

Jaganathancse: > +1 ideally you would also remove the become true here
> 
> https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/playbooks/neutron_sriov.yml#L6
> 
> the neutron role should elevate in the tasks that need it now at the playbook level
> 
> the same issue is present in all 4 neutron playbooks this seams to have been cargo culted to all the neutron playbooks from the ovn playbook but since its a preexsitign bug it could be fixed in different pr

Ack

roles/edpm_nova/tasks/install.yml
Jaganathancse: > oh ya that an existg bug from when i initall ported this
> 
> it should not use become here +1

Ack

roles/edpm_neutron_sriov/tasks/install.yml
```diff
@@ -19,9 +19,11 @@
     path: "{{ item.path }}"
```
SeanMooney, line 19: you should have become: true here

now you are relying on the become form the playbook level which is incorrect

SeanMooney (changes requested):

roles/edpm_neutron_sriov/tasks/install.yml
Jaganathancse: > you should have become: true here
> 
> now you are relying on the become form the playbook level which is incorrect

Ack

roles/edpm_neutron_sriov/tasks/install.yml
```diff
@@ -15,13 +15,16 @@
...
     setype: "container_file_t"
     state: directory
+    owner: "{{ item.owner | default(ansible_user) }}"
+    group: "{{ item.group | default(ansible_user) }}"
     mode: "{{ item.mode | default(omit) }}"
   loop:
-    - {'path': "/var/lib/openstack/config/containers", "mode": "0750"}
+    - {'path': "/var/lib/openstack/config/containers", "mode": "0755"}
     - {'path': "/var/lib/neutron", "mode": "0750"}
```
SeanMooney, line 28: thise seems to work but you might want to chge the owner/group default to default(omit) and pass ansible_user for the files that should that user otherwise for /var/lib/neutron it probably should be owned by neutron

it will get changed when the neutron contianers start but you don't want to be changing it back to anisble_user while the contiaenrs are running

SeanMooney: this might be ok as is im just highlight this line since the user/group will now change and I'm not sure that was intentional

Jaganathancse: > thise seems to work but you might want to chge the owner/group default to default(omit) and pass ansible_user for the files that should that user otherwise for /var/lib/neutron it probably should be owned by neutron
> 
> it will get changed when the neutron contianers start but you don't want to be changing it back to anisble_user while the contiaenrs are running

done

Jaganathancse: > this might be ok as is im just highlight this line since the user/group will now change and I'm not sure that was intentional

We have discussed like, this directories should be owned by 'cloud-admin'. so used ansible_user.

roles/edpm_nova/tasks/configure.yml
Jaganathancse: > on line 16 we are doing
> 
> become: true ansible.builtin.file: path: "{{ item.path }}" state: "directory" setype: "container_file_t" owner: "{{ item.owner | default(ansible_user) }}" group: "{{ item.group | default(ansible_user) }}" mode: "{{ item.mode | default(omit) }}" loop:
> 
> * {"path": "{{ edpm_nova_config_dest }}", "mode": "0755"}
> 
> i.e. we are creating edpm_nova_config_dest with 755 permission and the ownership is set to the current ansible user
> 
> we should be able to rewite to edpm_nova_config_dest without become when we invoke ansible.builtin.template below on line 46

@SeanMooney 
Updated this change based on this comment.

roles/edpm_neutron_sriov/tasks/install.yml
Jaganathancse: @SeanMooney 
For workaround, can we merge this PR, will work and create another PR for above changes later.
and also need to remove become in playbook level.

Jaganathancse: > thise seems to work but you might want to chge the owner/group default to default(omit) and pass ansible_user for the files that should that user otherwise for /var/lib/neutron it probably should be owned by neutron
> 
> it will get changed when the neutron contianers start but you don't want to be changing it back to anisble_user while the contiaenrs are running

Tested and working fine when i use default(omit) instead of ansible_user but wner and group will be root for dir '/var/lib/openstack/config'. will it be fine?

SeanMooney: no /var/lib/openstack/config must be ownewd by ansible_user

the fact it is owned by root is what you need to fix in this series

Jaganathancse: > no /var/lib/openstack/config must be ownewd by ansible_user
> 
> the fact it is owned by root is what you need to fix in this series

Ack

SeanMooney: you need to chagne it to 

- {'path': "/var/lib/openstack/config/containers", "mode": "0755", user: "{{ ansible_user }}", group: "{{ ansible_user }}" }

SeanMooney: same for 
edpm_neutron_sriov_agent_config_dir and /var/log/containers/stdouts

Jaganathancse: done

roles/edpm_neutron_sriov/molecule/default/converge.yml
```diff
@@ -16,7 +16,10 @@
 
 - name: Converge
   hosts: all
+  gather_facts: true
+  become: true
```
SeanMooney, line 20: no... 
this is still wrong you keep elevating to root globally

the role should eleve at the task level
if you need to od it here then there is a bug find the role.

roles/edpm_neutron_sriov/molecule/default/converge.yml
```diff
@@ -16,7 +16,10 @@
...
   roles:
     - role: "edpm_neutron_sriov"
       vars:
         edpm_neutron_sriov_agent_config_src: "{{lookup('env', 'MOLECULE_SCENARIO_DIRECTORY')}}/test-data"
+        ansible_user: root
```
SeanMooney, line 25: you should not need to set this too root.

the correct fix is to add these pre_tasks
```
 pre_tasks:
    - name: set basic user fact
      set_fact:
        ansible_user: "{{ lookup('env', 'USER') }}"
      when:
        - ansible_user is undefined
    - name: set basic home fact
      set_fact:
        ansible_user_dir: "{{ lookup('env', 'HOME') }}"
      when:
        - ansible_user_dir is undefined
```

https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_nova/molecule/default/converge.yml#L3-L13

SeanMooney (changes requested):

roles/edpm_neutron_sriov/molecule/default/converge.yml
Jaganathancse: > you should not need to set this too root.
> 
> the correct fix is to add these pre_tasks
> 
> ```
>  pre_tasks:
>     - name: set basic user fact
>       set_fact:
>         ansible_user: "{{ lookup('env', 'USER') }}"
>       when:
>         - ansible_user is undefined
>     - name: set basic home fact
>       set_fact:
>         ansible_user_dir: "{{ lookup('env', 'HOME') }}"
>       when:
>         - ansible_user_dir is undefined
> ```
> 
> https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_nova/molecule/default/converge.yml#L3-L13

done

Jaganathancse: recheck

Jaganathancse:
recheck

roles/edpm_neutron_sriov/molecule/default/converge.yml
SeanMooney: ok so there are two things happening

the reason root seams to be need for edpm_neutron_sriov is because its running its molecule tests in prow with the podman driver instead of in zuul with the delegated driver.

we should try to figure out how to fix this because the prow jobs are likely not testing things properly as a result.

however we can proceed with this for now to unblock the nfv work.

SeanMooney (approved):

Jaganathancse:
recheck

jpodivin:
rechcek

Jaganathancse:
recheck

roles/edpm_neutron_sriov/molecule/default/converge.yml
SeanMooney: as an aside i did some digging and 

ansible_user is a connect variable only defined when using the ssh connection type .i.e. not when using podma/kubectl

i belive ansible_user_id is available once we gather facts for all connection types and it contains the username ansible will be using on the remote host. in otherwords i has the same info as ansible_user for ssh connection and also provides that info when using podman/kubectl connections too.

so we should refecator our use of ansible_user to ansible_user_id globally in edpm-ansible as a separate pr

Jaganathancse:
recheck

Jaganathancse:
recheck

Jaganathancse:
recheck

Jaganathancse:
Looks like ci edpm-ansible pod hitting below error
TASK [osp.edpm.edpm_users : Create host users and groups] **********************
Monday 22 January 2024  20:58:58 +0000 (0:00:00.085)       0:00:03.609 ******** 
[0;36mincluded: /usr/share/ansible/collections/ansible_collections/osp/edpm/roles/edpm_users/tasks/create_users_and_groups.yml for edpm-compute-0 => (item={'name': 'openvswitch', 'uid': '42476', 'gid': '42476', 'shell': '/sbin/nologin', 'groups': 'hugetlbfs', 'comment': 'openvswitch user'})[0m

TASK [osp.edpm.edpm_users : Ensure group is present on the host [ openvswitch ]] ***
Monday 22 January 2024  20:58:58 +0000 (0:00:00.105)       0:00:03.715 ******** 
[0;32mok: [edpm-compute-0] => {"changed": false, "gid": 42476, "name": "openvswitch", "state": "present", "system": false}[0m

Monday 22 January 2024  20:58:59 +0000 (0:00:01.059)       0:00:04.775 ******** 
[0;31mfatal: [edpm-compute-0]: FAILED! => {"changed": false, "msg": "usermod: user openvswitch is currently used by process 1189\n", "name": "openvswitch", "rc": 8}[0m


above issue fix PR is also hitting same ci issue and not using this PR in CI.
https://github.com/openstack-k8s-operators/edpm-ansible/pull/540

bshephar:
Let's try Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/540

See if it picks up the code from that PR to get this one passing.

lewisdenny:
recheck

bshephar:
recheck

Jaganathancse:
recheck

PR #516, opened 2023-12-11
Stub documentation for all playbooks

jpodivin:
All of these are bare, with few exceptions. Thematically separated PRs will follow. However, if you already have some input about what should be present, don't hesitate to say so in the review.


slagle (commented):
More docs :+1: 

Could some of this be documented inline in the playbooks? Are there auto doc type tools for Ansible that could generate documentation from playbook comments? Given playbooks are meant to be human readable and users will likely be looking directly at them, I was thinking it would be good to have the docs inlined there as much as possible.

jpodivin:
> Could some of this be documented inline in the playbooks? Are there auto doc type tools for Ansible that could generate documentation from playbook comments? Given playbooks are meant to be human readable and users will likely be looking directly at them, I was thinking it would be good to have the docs inlined there as much as possible.

I have considered that option. However, the tools for automatic documentation generation, like antsibull which we already use, are not well suited for this task. Either they lack support for documenting playbooks completely, or they not under very active development. Being dependent on them could force us into very uncomfortable position down the line. I've seen it happen with tripleo.

We could, alternatively, develop our own plugin for sphinx, after all we are already overriding default antisbul-docs behavior to make documentation building possible. But that would have it's own set of challenges. Someone would need to maintain it, devoting time to a work on SW lacking overlap with our product. And if not, we would, eventually get failures. Blocking our pipeline and preventing us from merging patches, just because the documentation plugin is broken.
Again, this is something I have experienced with tripleo.

Finally, we are approaching move to asciidocs, and general harmonization of upstream and downstream documentation. Imho, the fewer moving parts our docs have upstream the better.

We can, and we should, write comments in playbooks, especially those that are non-trivial. But I would refrain from automatically compiling rst files from them.

slagle:
> > Could some of this be documented inline in the playbooks? Are there auto doc type tools for Ansible that could generate documentation from playbook comments? Given playbooks are meant to be human readable and users will likely be looking directly at them, I was thinking it would be good to have the docs inlined there as much as possible.
> 
> I have considered that option. However, the tools for automatic documentation generation, like antsibull which we already use, are not well suited for this task. Either they lack support for documenting playbooks completely, or they not under very active development. Being dependent on them could force us into very uncomfortable position down the line. I've seen it happen with tripleo.
> 
> We could, alternatively, develop our own plugin for sphinx, after all we are already overriding default antisbul-docs behavior to make documentation building possible. But that would have it's own set of challenges. Someone would need to maintain it, devoting time to a work on SW lacking overlap with our product. And if not, we would, eventually get failures. Blocking our pipeline and preventing us from merging patches, just because the documentation plugin is broken. Again, this is something I have experienced with tripleo.
> 
> Finally, we are approaching move to asciidocs, and general harmonization of upstream and downstream documentation. Imho, the fewer moving parts our docs have upstream the better.
> 
> We can, and we should, write comments in playbooks, especially those that are non-trivial. But I would refrain from automatically compiling rst files from them.

That all sounds reasonable. As long as we can get a few comments in the playbooks, I think that would be good.

jpodivin:
recheck

retry limit

jpodivin:
recheck

` k8s.io/client-go/tools/watch/informerwatcher.go:146: Failed to watch *unstructured.Unstructured: the server has asked for the client to provide credentials", "error: timed out waiting for the condition on openstackcontrolplanes/openstack-galera-network-isolation"`

jpodivin:
recheck

retry limit

fao89 (approved):

PR #517, opened 2023-12-12
Install rootwrap.conf where neutron rpm expects it

booxter:
Use kolla to copy the rootwrap config file under /etc/neutron/rootwrap.conf.


slagle (commented):
+2

karelyatin (commented):
/lgtm

PR #518, opened 2023-12-13
Peform DNF update on edpm node via bootstrap_role

raukadah:
In baremetal EDPM job, we use EDPM image to create EDPM node. it has all the required packages pre-installed during image build process.

If an older version of package got included during the image build process.  Then during edpm_deploy, bootstrap role is not going to update the package to latest.

Performing a dnf update will update all the system packages and avoid such issues.

### DIFF
--- a/roles/edpm_bootstrap/tasks/download_cache.yml
+++ b/roles/edpm_bootstrap/tasks/download_cache.yml
@@ -34,3 +34,15 @@
   until: edpm_bootstrap_release_version_package_download.rc == 0
   retries: "{{ edpm_bootstrap_download_retries }}"
   delay: "{{ edpm_bootstrap_download_delay }}"
+
+- name: Download updated packages
+  become: true
+  ansible.builtin.dnf:
+    name: "*"
+    state: latest  # noqa: package-latest
+    download_only: true
+    update_cache: true
+  register: edpm_bootstrap_system_updates_download
+  until: edpm_bootstrap_system_updates_download.rc == 0
+  retries: "{{ edpm_bootstrap_download_retries }}"
+  delay: "{{ edpm_bootstrap_download_delay }}"
--- a/roles/edpm_bootstrap/tasks/update_package.yml
+++ b/roles/edpm_bootstrap/tasks/update_package.yml
@@ -1,5 +1,5 @@
 ---
-# Copyright 2019 Red Hat, Inc.
+# Copyright 2023 Red Hat, Inc.
 # All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may
@@ -14,11 +14,14 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Import download_cache tasks
-  ansible.builtin.import_tasks: download_cache.yml
-
-- name: Import bootstrap tasks
-  ansible.builtin.import_tasks: bootstrap.yml
-
-- name: Import Package update tasks
-  ansible.builtin.import_tasks: update_package.yml
+- name: Perform system update
+  become: true
+  ansible.builtin.dnf:
+    name: "*"
+    state: latest  # noqa: package-latest
+    update_cache: true
+    update_only: true
+  register: edpm_bootstrap_system_update
+  until: edpm_bootstrap_system_update.rc == 0
+  retries: "{{ edpm_bootstrap_download_retries }}"
+  delay: "{{ edpm_bootstrap_download_delay }}"


roles/edpm_bootstrap/tasks/packages.yml
```diff
@@ -39,7 +39,7 @@
 - name: Install required packages to bootstrap EDPM
   ansible.builtin.dnf:
     name: "{{ edpm_bootstrap_packages_bootstrap }}"
-    state: present
+    state: latest
```
slagle, line 42: this isn't the only place we install packages. Should we make this change consistent across all the tasks files? ``state: latest`` probably needs to be added to all of the download_cache.yml task files as well.

raukadah: Thank you @slagle for reviewing it.
https://github.com/openstack-k8s-operators/ci-framework/pull/936 fixes the issue. But the proper solution would be do a dnf update via edpm_ansible itself. I will update this pr to add the dnf update.

rebtoor (changes requested):
You should handle retries and delay like any other `dnf` tasks in the repo (see https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_bootstrap/tasks/packages.yml for an example)

You can create a new task file with this task and call it from the `install.yml` task file inside `download_cache` role.

Also, as suggested by @slagle , you should add this task to `download_cache.yml` task file.

rebtoor:
To pass the pre-lint commit test, add a space before the `noqa: package-latest` comment.

roles/edpm_bootstrap/tasks/update_package.yml
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Perform system update
+  become: true
+  ansible.builtin.dnf:
+    name: "*"
+    state: latest # noqa: package-latest
```
rebtoor, line 21:
```
- name: Perform system update
  become: true
  ansible.builtin.dnf:
    name: "*"
    state: latest  # noqa: package-latest
  register: edpm_bootstrap_system_update
  until: edpm_bootstrap_system_update.rc == 0
  retries: "{{ edpm_bootstrap_download_retries }}"
  delay: "{{ edpm_bootstrap_download_delay }}"
```

rebtoor: and, in `download_cache.yml` task file (from `edpm_bootstrap` role) add:

```
- name: Download updated packages
  become: true
  ansible.builtin.dnf:
    name: "*"
    state: latest  # noqa: package-latest
    download_only: true
  register: edpm_bootstrap_system_updates_download
  until: edpm_bootstrap_system_updates_download.rc == 0
  retries: "{{ edpm_bootstrap_download_retries }}"
  delay: "{{ edpm_bootstrap_download_delay }}"
```

rebtoor (changes requested):

roles/edpm_bootstrap/tasks/update_package.yml
raukadah: Thank you @rebtoor :-)

roles/edpm_bootstrap/tasks/packages.yml
rebtoor: `state: latest` shouldn't be used during packages installation as best practice. (That's why there's the `package-latest` linting test)

Having this task in bootstrap role has the happy side effect that the repo cache will be updated during the installation, so i guess that every package installed after that will be their current version. 

I'm wondering if we need to add `update_cache: true` ([link](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/dnf_module.html#parameter-update_cache)) to this task, though.

roles/edpm_bootstrap/tasks/update_package.yml
rebtoor: @raukadah can you add a `update_cache: true` to both dnf tasks please?

roles/edpm_download_cache/tasks/install.yml
```diff
@@ -22,3 +22,11 @@
...
     - edpm_podman
     - download_cache
+
+- name: Update packages
+  ansible.builtin.include_role:
+    name: osp.edpm.edpm_bootstrap
+    tasks_from: update_package.yml
+  tags:
+    - edpm_bootstrap
+    - download_cache
```
SeanMooney, line 32: this shoudl not have download_cache

SeanMooney (changes requested):
overall i think this is fine but the actual package update should not be part of the download_cache tags

roles/edpm_download_cache/tasks/install.yml
SeanMooney: more contex on this if i rann a deployment limited to the download_cache
tag i would not expect it to do a dnf update of all packages as a side effect.

we likely will document runign download_cache this way to pre stage images and package prior to an update/upgrade. as such we should make sure it only downloads pacakge/images and does not install them.

rebtoor: You're right @SeanMooney 

`download_cache` role should only download the packages without installing them, and it's already done [here](https://github.com/openstack-k8s-operators/edpm-ansible/blob/e53be776f54e6ea1f459e8f5e36af89539eeb47a/roles/edpm_download_cache/tasks/packages.yml#L3-L9)

So i guess we can remove this task entirely.

SeanMooney: the task is correct we jsut need to remove the - download_cache tag form it so that it does not run when you run just that tag.

we need this to do the package update when you run with 

- edpm_bootstrap

SeanMooney: oh actully no wrong file ya we don't need this at all 

roles/edpm_bootstrap/tasks/update_package.yml does the install

roles/edpm_bootstrap/tasks/download_cache.yml
```diff
@@ -34,3 +34,15 @@
   until: edpm_bootstrap_release_version_package_download.rc == 0
   retries: "{{ edpm_bootstrap_download_retries }}"
   delay: "{{ edpm_bootstrap_download_delay }}"
+
+- name: Download updated packages
+  become: true
+  ansible.builtin.dnf:
+    name: "*"
```
bogdando, line 41: have you measured the execution time? the last time I've  checked it (with much older ansible versions, I must admit) that command was taking hours compared to minutes by dnf update

bogdando: yea, that was quite ago... https://github.com/ansible/ansible/issues/26995
the things might get improved since then

rebtoor: if still takes "hours" we should raise a huge issue on ansible core team :)

bogdando: ok, please change for the shell command then, until fixed

rebtoor: Sure, when someone will demonstrate with proper testing that `ansible.builtin.dnf` is much slower than `dnf` called in a `command`/`shell` task, we'll change every instance of it in the repo and raise a bug to the ansible team.

rebtoor (approved):

roles/edpm_bootstrap/tasks/download_cache.yml
bogdando: I checked on my env at hand, the time it takes looks pretty much the same in both cases:

dnf module:
```
real    1m9.585s                            
user    0m43.438s
sys     0m11.120s
```
shell command:
```
real    1m4.617s
user    0m45.630s
sys     0m10.170s
```

rebtoor: we'll keep this aspect monitored though (especially with many repository enabled and lot of packages to install/upgrade)

rebtoor:
recheck

`RETRY_LIMIT`

rebtoor:
recheck

kernel panic on compute node -> https://logserver.rdoproject.org/18/518/c3ec651573761d3ae11949c418074a0bc335fd28/github-check/cifmw-crc-podified-edpm-baremetal/d1a9617/ci-framework-data/logs/bmaas_console_logs/edpm-compute-01_console.log

```
[  636.603289] Kernel panic - not syncing: Jitter RNG permanent health test failure
[  636.603294] CPU: 2 PID: 38381 Comm: dbus-send Not tainted 5.14.0-383.el9.x86_64 #1
[  636.603298] Hardware name: Red Hat KVM/RHEL, BIOS edk2-20230524-4.el9 05/24/2023
[  636.603300] Call Trace:
[  636.603304]  <TASK>
[  636.603311]  dump_stack_lvl+0x34/0x48
[  636.603345]  panic+0xfd/0x2f7
[  636.603350]  jent_kcapi_random.cold+0x15/0x43
[  636.603354]  drbg_seed+0x10a/0x440
[  636.603359]  ? drbg_generate+0xd9/0x320
[  636.603361]  ? _copy_to_iter+0xb7/0x620
[  636.603365]  ? xas_load+0x9/0xa0
[  636.603369]  drbg_generate+0xb9/0x320
[  636.603371]  ? _copy_to_iter+0x1d3/0x620
[  636.603373]  drbg_kcapi_random+0xd3/0x120
[  636.603376]  crypto_devrandom_read_iter+0xb6/0x220
[  636.603380]  ? __do_sys_getrandom+0xa9/0x140
[  636.603384]  ? switch_fpu_return+0x4c/0xd0
[  636.603388]  __do_sys_getrandom+0x9d/0x140
[  636.603392]  do_syscall_64+0x5c/0x90
[  636.603397]  ? __do_softirq+0x16a/0x2ac
[  636.603401]  ? __irq_exit_rcu+0x46/0xc0
[  636.603404]  ? sysvec_apic_timer_interrupt+0x3c/0x90
[  636.603406]  entry_SYSCALL_64_after_hwframe+0x72/0xdc
```

bogdando:
it seems we cannot just upgrade all packages, w/o adopting https://github.com/openstack/tripleo-ansible/blob/e281ae7624774d71f22fbb993af967ed1ec08780/tripleo_ansible/ansible_plugins/modules/tripleo_ovs_upgrade.py as well (CC @karelyatin )

PR #520, opened 2023-12-14
[molecule][podman] Drop remote_tmp config_options

karelyatin:
Running the scenarios with root user fails with
this config, it works fine without this option set. So let's drop it as it's not necessary like other
molecule podman scenarios.


booxter:
I think it's good but I'd love if someone with better knowledge of molecule could chime in. But this patch definitely helps running under root.

booxter:
@jpodivin do you have any thoughts on this patch? It seems to alleviate some issues when running molecule scenarios as root; is it needed for anything? other roles don't have the same config option set. Thanks for any help!

slawqo (approved):

PR #521, opened 2023-12-15
Configure rsyslog to being able to send TLS encrypted logs to the rem…

jlarriba:
…ote server

Depends-On: https://github.com/openstack-k8s-operators/telemetry-operator/pull/251

### DIFF
--- a/roles/edpm_telemetry/tasks/enable_logging.yml
+++ b/roles/edpm_telemetry/tasks/enable_logging.yml
@@ -20,13 +20,30 @@
   become: true
   failed_when: false
   register: rsyslog_service_state
+  tags:
+    - edpm_logging
 
 - name: Configure rsyslog if present
   become: true
   when:
     - (rsyslog_service_state is success) and
       ((rsyslog_service_state.status['SubState'] | lower) == 'running')
+  tags:
+    - edpm_logging
   block:
+    - name: Install openssl module support for rsyslog
+      ansible.builtin.dnf:
+        name: rsyslog-openssl
+        state: present
+
+    - name: Copy Openshift CA to the node
+      become: true
+      ansible.builtin.copy:
+        src: "{{ edpm_telemetry_config_src }}/ca-openshift.crt"
+        dest: "/etc/pki/rsyslog/ca-openshift.crt"
+        mode: "0644"
+        remote_src: "{{ telemetry_test | default('false') }}"
+
     - name: Deploy rsyslog configuration
       become: true
       ansible.builtin.copy:
@@ -35,5 +52,3 @@
         mode: "0644"
         remote_src: "{{ telemetry_test | default('false') }}"
       notify: Restart rsyslog
-      tags:
-        - edpm_logging
--- a/roles/edpm_telemetry/molecule/default/prepare.yml
+++ b/roles/edpm_telemetry/molecule/default/prepare.yml
@@ -207,9 +207,10 @@
         owner: root
         group: root
         content: |
-          module(load="imjournal" PersistStateInterval="10"
-                StateFile="journal_state")
-
           action(type="omfwd" target="172.17.0.80" port="10514" protocol="tcp"
                       action.resumeRetryCount="100"
-                      queue.type="linkedList" queue.size="10000")
+                      queue.type="linkedList" queue.size="10000"
+                      StreamDriver="ossl"
+                      StreamDriverMode="1"
+                      StreamDriverAuthMode="x509/certvalid"
+                      StreamDriver.CAFile="/etc/pki/rsyslog/ca-openshift.crt")


roles/edpm_telemetry/tasks/enable_logging.yml
```diff
@@ -20,13 +20,30 @@
...
+    - name: Install openssl module support for rsyslog
+      ansible.builtin.dnf:
+        name: rsyslog-openssl
+        state: present
+
+    - name: Copy Openshift CA to the node
+      become: true
+      ansible.builtin.copy:
+        src: "{{ edpm_telemetry_config_src }}/ca-openshift.crt"
+        dest: "/etc/ssl/certs/ca-openshift.crt"
```
rebtoor, line 43: can we define a var with this path so we can use it here and on molecule as well?

rebtoor (changes requested):

roles/edpm_telemetry/tasks/enable_logging.yml
jlarriba: I dont really want the user to be able to change this path, as there are other part of the configuration generated by the telemetry-operator which needs to use the same path: https://github.com/openstack-k8s-operators/telemetry-operator/pull/251/files#diff-b9f6028da3c23a90b425ff5d97240a112f428590c3ba0339dee2de54e027912fR11

I will use variables for molecule tests for all the rest of the paths.

jlarriba:
recheck

jlarriba:
recheck

jlarriba:
recheck

jlarriba:
recheck

jlarriba:
recheck

rebtoor:
@jlarriba it's useless to trigger a recheck since it's failing with the same error on the edpm_libvirt molecule check. I think it's better to involve nova folks for that.

(I did a quick check, it seems that it fails on checking if `logvirtd.service` is enabled, and it's not because it's an indirect service)

cc @gibizer @SeanMooney

roles/edpm_telemetry/tasks/enable_logging.yml
SeanMooney: we were disucssing about having a singel role for doing management of the tls ca and certs in https://github.com/openstack-k8s-operators/edpm-ansible/pull/506

looking at the other tls example  (nova)
https://github.com/openstack-k8s-operators/edpm-ansible/pull/491/files

we are installing the certs to 

/var/lib/openstack/certs/<service name>/ and binding that to /etc/pki/<servicename>/... in the container i.e.

/var/lib/openstack/certs/nova/tls.crt:/etc/pki/nova/server-cert.pem:ro


are you sure /etc/ssl/certs/ca-openshift.crt is the correct path?

would it not be /etc/pki/rsyslog/...


@vakwetu any insight here?

SeanMooney:
im currently trying to repoduce but perhaps this is just the openstack-selinux package issue again? could that have been regressed? we are expecting to have the latest version installed
if an older version is used the virtlogd cant start

roles/edpm_telemetry/tasks/enable_logging.yml
jlarriba: Well, according to where the certs should be deployed, it could be anywhere. I chose /etc/ssl/certs for being the default cert path. /etc/pki/rsyslog is as good as the other one, no problem.

But before doing the changes, I am interested in that single role for doing management of the cacerts. The one I need here is the default openshift CA, which signs all the openshift certs. Is nova already getting that cacert from openshift and storing it somewhere (/var/lib/openstack/certs/nova/...)?

SeanMooney:
oh its not installed at all
https://review.rdoproject.org/zuul/build/8b7749f3aac048fc9b6b0b5a6b01545b/log/installed-pkgs.log

so its normally installed by bootstrap https://github.com/openstack-k8s-operators/edpm-ansible/blob/afb124bbc6a799eec51c2785de0e55c81a9b88da/roles/edpm_bootstrap/defaults/main.yml#L36

but i would guess we are not calling bootstrap in the libvirt molecule tests

if so that a pretty easy fix either i can install openstack-selinux in prepare or just add the bootstrap role

SeanMooney:
> > but i would guess we are not calling bootstrap in the libvirt molecule tests
> 
> correct.
> 
> > if so that a pretty easy fix either i can install openstack-selinux in prepare or just add the bootstrap role
> 
> if you only need that package i guess you can install it in prepare.

so i think  there are two things going on here.
first this patch need a rebase it quite a bit behind the tip of main.
second we need to install the newer openstack-selinux package and we need to do the either in prepare directly or by calling bootstrap.

I'm locally testing a rebase of this to confirm

SeanMooney:
ok i have a passing version locally
it would appear that

systemctl is-enabeld virtlogd has changed from "indirect" to "enabled"

i am rebaseing my local changes and testing if that is all that is required to fix this and ill submit that as a separate pr.

since we only need the new selinux rules if we are actually booting a nova VM we don't need to install the openstack-selinux in the molecule env. as such i will drop out the change i have locally to call bootstrap assuming molecule verify passes locally. it running now.

it would still be good to rebase this patch on something closer to main but that is not the reason why this is failing.

my best guess is this is related to an rpm change or something similar that altered the output of is-enabled

SeanMooney:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/523 shoudl fix the issue
once that is merged then you can rebase/recheck this

SeanMooney:
https://review.rdoproject.org/zuul/build/b865a935e7cf4ceeb6e3d11901ccbb8c #523 passed the molecule test so that should unblock this once its merged
I'm going to finish for today o/

SeanMooney:
recheck #523 is now merged

rebtoor:
@SeanMooney thanks a lot for the quick fix!

jlarriba:
@rebtoor all tests passed, could you approve please?

roles/edpm_telemetry/tasks/enable_logging.yml
jlarriba: I changed the path for the cacert to /etc/pki/rsyslog

PR #522, opened 2023-12-15
Use different src directories for telemetry and logging

jlarriba:


rebtoor (changes requested):
PR could be fine, but you should fix a couple of things:

- when you define permissions choose `"` or `'` and use only one of them, especially in the same file.
- molecule test for `edpm_telemetry` role is failing. You should use the same variables defined in `defaults/main.yml` instead of hardcoding paths, otherwise every time you change something, you need to remember to fix them.

rebtoor (approved):

PR #525, opened 2023-12-20
Add role edpm_install_certs

vakwetu:
This is a role that moves certs into the right locations for all the services.  This is executed by a service early in the cycle, so that the services can have all the certs available.  So nova, for example, will be able to use the libvirt certs.

### DIFF
--- a/playbooks/install_certs.yml
+++ b/playbooks/install_certs.yml
@@ -1,0 +1,13 @@
+---
+
+- name: EDPM Install Certs
+  hosts: all
+  strategy: free
+  any_errors_fatal: "{{ edpm_any_errors_fatal | default(true) }}"
+  max_fail_percentage: "{{ edpm_max_fail_percentage | default(0) }}"
+  tasks:
+    - name: Install EDPM Certs
+      ansible.builtin.import_role:
+        name: osp.edpm.edpm_install_certs
+      tags:
+        - edpm_install_certs
--- a/roles/edpm_install_certs/meta/main.yml
+++ b/roles/edpm_install_certs/meta/main.yml
@@ -1,0 +1,33 @@
+---
+# Copyright 2023 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+galaxy_info:
+  namespace: openstack
+  author: OpenStack
+  description: EDPM OpenStack Role -- edpm_install_certs
+  company: Red Hat
+  license: Apache-2.0
+  min_ansible_version: '2.9'
+  platforms:
+    - name: 'EL'
+      versions:
+        - '9'
+
+  galaxy_tags:
+    - edpm
+
+dependencies: []
--- a/roles/edpm_install_certs/tasks/copy_certs_and_keys.yaml
+++ b/roles/edpm_install_certs/tasks/copy_certs_and_keys.yaml
@@ -1,0 +1,49 @@
+---
+# Copyright 2023 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Copy over certs and keys
+  tags:
+    - certs
+  become: true
+  block:
+    - name: Set paths
+      ansible.builtin.set_fact:
+        cert_src_path: "/var/lib/openstack/certs/{{ service }}"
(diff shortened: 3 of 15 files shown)


SeanMooney (commented):
i ran out of time to review this today but I'm around tomorrow so ill try and review it then

roles/edpm_install_certs/meta/main.yml
```diff
@@ -0,0 +1,34 @@
...
+  namespace: openstack
+  author: OpenStack
+  description: EDPM OpenStack Role -- edpm_install_certs
+  company: Red Hat
+  license: Apache-2.0
+  min_ansible_version: '2.9'
+  platforms:
+    - name: 'EL'
+      versions:
+        - '8'
```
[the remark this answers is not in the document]
vakwetu: fixed

vakwetu:
Just a comment for explanation:

The install-certs dataplane service calls this playbook/role to handle all the certs for all the services for all nodes in the nodeset.

The certs/keys will be deployed to the following location on each node:
   cert: /var/lib/openstack/certs/<service>/tls.crt
   key: /var/lib/openstack/certs/<service>/tls.key
   ca certs: /var/lib/openstack/cacerts/<service>/tls-ca-bundle.pem

libvirt is moved to a different location (because its being mounted locally) .
  cert: /etc/pki/libvirt/tls.crt
  key:  /etc/pki/libvirt/private/tls.key
  ca certs: /etc/pki/ca/libvirt/tls-ca-bundle.pem

If you like we can move the special case code for libvirt to the libvirt edpm role if you like.  I'm ambivalent, really.

vakwetu:
recheck

vakwetu:
recheck

vakwetu:
Note: all the jobs appear to have passed in different runs.

roles/edpm_nova/templates/nova_compute.json.j2
```diff
@@ -13,7 +13,7 @@
 {% if edpm_nova_tls_certs_enabled %}
 	"/var/lib/openstack/certs/nova/tls.crt:/etc/pki/nova/server-cert.pem:ro",
 	"/var/lib/openstack/certs/nova/tls.key:/etc/pki/nova/private/server-key.pem:ro",
-	"/var/lib/openstack/cacerts/nova/TLSCABundleFile:/etc/pki/CA/cacert.pem:ro",
+	"/var/lib/openstack/cacerts/nova/tls-ca-bundle.pem:/etc/pki/CA/cacert.pem:ro",
```
SeanMooney, line 16: nova will need the ca key so that it will trust the libvirt certs although I'm not sure if we actually need client certs for nova itself since we are not using two way cert auth

roles/edpm_install_certs/tasks/copy_certs_and_keys.yaml
```diff
@@ -0,0 +1,55 @@
...
+      ansible.builtin.set_fact:
+        cert_src_path: "/var/lib/openstack/certs/{{ service }}"
+        cert_dest_path: "/var/lib/openstack/certs/{{ service }}"
+        key_dest_path: "/var/lib/openstack/certs/{{ service }}"
+
+    - name: Set cert_dest_path for libvirt
+      ansible.builtin.set_fact:
+        cert_dest_path: "/etc/pki/libvirt"
+        key_dest_path: "/etc/pki/libvirt/private"
+      when: service  == "libvirt"
```
SeanMooney, line 32: im ok with this but it would likely be more scalable if you just put
the cert in /var/lib/openstack/certs/libvirt and then modify the libvirt role to copy it form there to /etc/pki/libvirt if it exits

that would allow any other service that need to be host installed to follow the same pattern without modifying this role.

SeanMooney (commented):
overall i like the direction this is taking.

I'm not strongly opposed to inlining the libvirt cert management as you have but i think its likely better long term to externalise it and just put the libvirt certs in the normal place.

the main reason for that is actually third party integrations.

if we want to allow them to use this role to copy there certs then we would like the pattern to be that this role prepares the certs in a standard location and then a second role uses them, either bind mounting them to a container or copying them to a service specific location.

so with that in mind we probably should not special case libvirt in the generic role and just configure libvirt to look in a diffentlocation if it support that or copy the libvirt certs in the libvirt role to the expected location.

i don't have a stong prefrnec however so i would follow your best judgment

jpodivin (changes requested):
Good work. However I think it needs documentation, for both the role and the playbook. Few sentences are enough but it really should be done. 

Furthermore, I'm wondering about the strategy. It may not save a lot of time, but it could be a benefit. Finally, I know there are not arguments to pass here. But argspec could still be implemented to prevent people passing args that don't exist, and as a way to start documentation for the role.

playbooks/install_certs.yml
```diff
@@ -0,0 +1,13 @@
+---
+
+- name: EDPM Install Certs
+  hosts: all
+  strategy: linear
```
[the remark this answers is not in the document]
vakwetu: done

roles/edpm_install_certs/tasks/copy_certs_and_keys.yaml
vakwetu: ack - will do as suggested, and add the relevant code to the libvirt role.

roles/edpm_libvirt/tasks/configure.yml
```diff
@@ -99,18 +99,19 @@
   changed_when: true
   when: run_libvirt.rc == 0
 
-- name: Copy TLS files to the compute node
+- name: Move TLS files to the right location on the compute node
```
SeanMooney, line 102: this is technially still a copy
just not a copy form the AEE container

SeanMooney (approved):

jpodivin (approved):

jpodivin:
recheck

`fatal: [localhost]: FAILED! => {"attempts": 60, "changed": true, "cmd": ["oc", "get", "po", "-l", "osp-provisionserver/name=openstack-edpm-ipam-provisionserver", "-n", "openstack", "-o", "name"], "delta": "0:00:00.131137", "end": "2024-01-09 01:39:50.673721", "msg": "", "rc": 0, "start": "2024-01-09 01:39:50.542584", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}`

PR #526, opened 2023-12-27
[edpm_kernel] Improve `edpm_kernel_args` search in `/proc/cmdline`

rebtoor:
In kernelargs.yml file we have a block of tasks for the definitions of custom kernel arguments. That block has 3 conditionals checks, one of them is the evaluation of kernel parameters defined in /proc/cmdline using the `edpm_kernel_args not in cmdline` notation. The problem is that this kind of notation will evaluate also partial matches.

E.g. if we first launch a deployment defining

```
edpm_kernel_args="isolcpus=1,3,5,7,9,11,13,15"
```

and then we'll do a deployment defining

```
edpm_kernel_args="isolcpus=1,3,5,7,9"
```

the check will evaluate to false because it matches the substring so the new kernel arguments won't be applied.

Also, a new molecule test for this specific case has been added.

### DIFF
--- a/roles/edpm_kernel/molecule/kernelargs-update-substring/converge.yml
+++ b/roles/edpm_kernel/molecule/kernelargs-update-substring/converge.yml
@@ -1,0 +1,63 @@
+---
+# Copyright 2022 Red Hat, Inc.
+# All Rights Reserved.
+#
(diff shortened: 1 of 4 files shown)


rebtoor:
recheck

slagle (approved):

PR #528, opened 2024-01-04
Clarifying tennant_ip argument

jpodivin:


slagle (approved):
This is good in that it documents that current behavior. However, I think we should move away from requiring tenant_ip given that it's a variable set based on a composable network name, which may or may not exist. I think we should update the automatic variable setting logic in dataplane-operator to just set edpm_ovn_encap_ip instead based on one of the ips from the composable networks, which can be annotated in some fashion.

PR #530, opened 2024-01-04
Call package update via bootstrap playbook

raukadah:
Currently tempest tests are failing due to old packages installed on edpm node.

Recently we have added https://github.com/openstack-k8s-operators/edpm-ansible/pull/518 to perform DNF update via edpm_bootstrap role.

But bootstrap-openstack-edpm-ipam pod calls bootstrap.yml playbook not the role that's why update_packages playbook is not called leaving old packages on the node.

Calling update_packages.yml playbook via bootstrap.yml playbook updates the packages and fixes the issue.

Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/532

### DIFF
--- a/roles/edpm_bootstrap/tasks/bootstrap.yml
+++ b/roles/edpm_bootstrap/tasks/bootstrap.yml
@@ -63,6 +63,9 @@
 - name: Import packages tasks
   ansible.builtin.import_tasks: packages.yml
 
+- name: Import Package update tasks
+  ansible.builtin.import_tasks: update_package.yml
+
 - name: Set selinux state
   ansible.posix.selinux:
     policy: targeted
--- a/roles/edpm_bootstrap/tasks/main.yml
+++ b/roles/edpm_bootstrap/tasks/main.yml
@@ -19,6 +19,3 @@
 
 - name: Import bootstrap tasks
   ansible.builtin.import_tasks: bootstrap.yml
-
-- name: Import Package update tasks
-  ansible.builtin.import_tasks: update_package.yml


slagle (commented):
https://github.com/openstack-k8s-operators/edpm-ansible/pull/518 is being reverted in https://github.com/openstack-k8s-operators/edpm-ansible/pull/531

I would prefer to keep the package update as something that is done in CI, or can be enabled from CI by setting an ansible variable. Forcing a package update will break disconnected environments that are using images with pre-installed packages.

slagle:
> #518 is being reverted in #531
> 
> I would prefer to keep the package update as something that is done in CI, or can be enabled from CI by setting an ansible variable. Forcing a package update will break disconnected environments that are using images with pre-installed packages.

@bogdando FYI

jpodivin (commented):
I would prefer the current way. If the CI is failing due to missing updates we should address the issue in CI, not in the collection.

bogdando:
FYI https://github.com/openstack-k8s-operators/edpm-ansible/pull/532
@SeanMooney has some better ideas than reverting the subject. So if you could rebase this on top of the above pr 532?

bogdando (changes requested):
please see my comments above

raukadah:
recheck

bogdando:
> > please see my comments above
> 
> Thank you @bogdando for the review, I have added Depends-On: #532 , I think we donot need to rebase it. Zuul will take care of testing it together.

There likely merge conflicts to happen. Also, it could be that #532 already contains this change?

raukadah:
@bogdando Thank you! I have tested https://github.com/openstack-k8s-operators/edpm-ansible/pull/532 with https://github.com/openstack-k8s-operators/ci-framework/pull/1027 here https://review.rdoproject.org/r/c/testproject/+/51226/2#message-19838ba0d7e193471ac55cfa369f81d7bfc7a477 It fixes the package update issue.

I think we no longer need https://github.com/openstack-k8s-operators/edpm-ansible/pull/530 pr. Hence, closing it.

PR #531, opened 2024-01-04
Revert "Peform DNF update on edpm node"

bogdando:
This reverts commit c3ec651573761d3ae11949c418074a0bc335fd28.

Attempting to solve system-wide conflicts for all OSP17.1 packages, to upgrade them into versions provided by RDO Antelope DLRN repos (deployed as a hack, until we figure a better source of OSP18 repos on rhel9 hosts used in the Adoption CI) yields problems with openvswitch packages upgrades.

It also seems that we might as well have to carry on OVS RPMs upgrade logic from TripleO [1], [2], and cannot simply 'ensure: latest' by Ansible.

[1] https://github.com/openstack/tripleo-ansible/blob/e281ae7624774d71f22fbb993af967ed1ec08780/tripleo_ansible/ansible_plugins/modules/tripleo_ovs_upgrade.py
[2] https://github.com/openstack/tripleo-heat-templates/blob/1393d39be367db3acb02508e0e858395a4e4fefa/deployment/undercloud/undercloud-upgrade.yaml#L146


bogdando:
note: this cannot be tested for adoption CI as we don't have edpm-ansible in the required-projects for that job.
And more importantly we don't build the operators in `data-plane-adoption` repo, so that change would not be applied anyway.

I'm going to test it with my custom AEEE operator setup on my adoption dev env

SeanMooney:
i think we shoudl be able to do https://github.com/openstack-k8s-operators/edpm-ansible/pull/532 instead.

that has the revert but it also has other changes to not regress the update of openstack-selinux


for adoption in think you could just define
```
edpm_ovs_packages:
 - openvswitch2.15.x86_64
```
or 
```
edpm_ovs_packages:
 - openvswitch3.2.x86_64
```
as required instead of the default

```
edpm_ovs_packages:
 - openvswitch
```

bogdando:
> i think we shoudl be able to do #532 instead.

I'l test this in my adoption setup

PR #532, opened 2024-01-08
edpm ovs

SeanMooney:
This commit adds a new edpm_ovs role to manage the
installation of openvswitch. As part of this change,
the installation is moved from the bootstrap playbook to the
configure_network playbook. To aide with testing the
verify_sockets_exists test helper is moved from the edpm_libvirt
role to the top-level molecule test-helpers and refactored to
work with any unix socket. The edpm_ovs role uses this to validate
that the ovs-db socket is created with the expected permission.

As part of this refactor a latent bug where the OVS user was not
created before the OVS package was installed has been fixed.
This will ensure that the OVS user on the host and in the containers
are aligned to the static user/group ID used in the container images.

Finally, a new zuul job is added to run the edpm_ovs tests.
Closes: [OSPRH-3451](https://issues.redhat.com//browse/OSPRH-3451)


bogdando:
I'' test this with https://github.com/openstack-k8s-operators/data-plane-adoption/pull/225

bogdando:
I tested it, and it worked for my dev setup. I'm still struggling to confirm the ansibleee jobs logs have had the subject ovs tasks included though :)

```
TASK [dataplane_adoption : verify if Nova services can start the existing test VM instance] ***                                                                                  
FAILED - RETRYING: [localhost]: verify if Nova services can start the existing test VM instance (60 retries left).                                                               
FAILED - RETRYING: [localhost]: verify if Nova services can start the existing test VM instance (59 retries left).                                                               
changed: [localhost] => {"attempts": 3, "changed": true, "cmd": "set -euxo pipefail\n\nalias openstack=\"oc exec -t openstackclient -- openstack\"\nFIP=192.168.122.20\n\n${BASH_
ALIASES[openstack]} server list | grep -qF '| test | SHUTOFF |' && ${BASH_ALIASES[openstack]} server start test\n${BASH_ALIASES[openstack]} server list | grep -F '| test | ACTIV
E |'\n${BASH_ALIASES[openstack]} server --os-compute-api-version 2.48 show --diagnostics test --fit-width -f json | jq -r '.state' | grep running\n", "delta": "0:00:12.233542", 
"end": "2024-01-10 13:53:59.978917", "msg": "", "rc": 0, "start": "2024-01-10 13:53:47.745375", "stderr": "+ alias 'openstack=oc exec -t openstackclient -- openstack'\n+ FIP=192
.168.122.20\n+ oc exec -t openstackclient -- openstack server list\n+ grep -qF '| test | SHUTOFF |'\n+ oc exec -t openstackclient -- openstack server list\n+ grep -F '| test | A
CTIVE |'\n+ oc exec -t openstackclient -- openstack server --os-compute-api-version 2.48 show --diagnostics test --fit-width -f json\n+ jq -r .state\n+ grep running", "stderr_li
nes": ["+ alias 'openstack=oc exec -t openstackclient -- openstack'", "+ FIP=192.168.122.20", "+ oc exec -t openstackclient -- openstack server list", "+ grep -qF '| test | SHUT
OFF |'", "+ oc exec -t openstackclient -- openstack server list", "+ grep -F '| test | ACTIVE |'", "+ oc exec -t openstackclient -- openstack server --os-compute-api-version 2.4
8 show --diagnostics test --fit-width -f json", "+ jq -r .state", "+ grep running"], "stdout": "| 08e221c2-8627-4f9d-a8cc-9ad00ff2f17d | test | ACTIVE | private=192.168.0.21, 19
2.168.122.20 | cirros | m1.small |\nrunning", "stdout_lines": ["| 08e221c2-8627-4f9d-a8cc-9ad00ff2f17d | test | ACTIVE | private=192.168.0.21, 192.168.122.20 | cirros | m1.small
 |", "running"]} 
```

roles/edpm_bootstrap/tasks/bootstrap.yml
```diff
@@ -98,16 +98,5 @@
         name: NetworkManager
         state: reloaded
```
bogdando, line 99: I can still see 'Enable openvswitch service if installed' and other removed tasks in aee jobs logs on my test env. I haven't tested this change properly, sorry for the noise

slagle (commented):
I think this is looking ok to me. It would be nice if the commit messages were fixed to accurately describe the scope of the change. Also, dropping the WIP from the one of the commit messages, and fixing the spelling errors would help us with accuracy.

SeanMooney:
@slagle i had that marked as WIP because if this work for adoption i planned to update that commit with molecule tests to verify the correct installation and execution of ovs.

so yes ill try and respin this and address the typos ectra if people are ok with this direction
probably not today but i can see if i can do it tomorrow.

roles/edpm_ovs/molecule/default/verify.yml
```diff
@@ -0,0 +1,103 @@
+---
```
SeanMooney, line 1: ill work and updatign this to actully verify ovs installation
this is still just a copy paste form the libvirt role.

roles/edpm_ovs/meta/argument_specs.yml
```diff
@@ -0,0 +1,60 @@
+---
```
SeanMooney, line 1: i also need to update this

playbooks/configure_network.yml
```diff
@@ -7,6 +7,11 @@
   any_errors_fatal: "{{ edpm_any_errors_fatal | default(true) }}"
   max_fail_percentage: "{{ edpm_max_fail_percentage | default(0) }}"
   tasks:
+    - name: Import  edpm_ovs to install ovs packages
```
SeanMooney, line 10: extra space

bogdando:
re-testing this with https://github.com/openstack-k8s-operators/data-plane-adoption/pull/225

bogdando:
I confirm this worked as the part of https://github.com/openstack-k8s-operators/data-plane-adoption/pull/193, with https://github.com/openstack-k8s-operators/data-plane-adoption/pull/225 embedding this change

SeanMooney:
great ok ill correct the outstaing issues on Monday and get this into a mergable state for people to review

roles/edpm_ovs/tasks/download_cache.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: Downloading required packages
+  ansible.builtin.dnf:
+    name: "{{ edpm_ovs_packages }}"
+    download_only: true
```
slagle, line 5: is state:latest not needed here?

roles/edpm_bootstrap/tasks/download_cache.yml
```diff
@@ -34,15 +34,3 @@
   until: edpm_bootstrap_release_version_package_download.rc == 0
   retries: "{{ edpm_bootstrap_download_retries }}"
   delay: "{{ edpm_bootstrap_download_delay }}"
-
-- name: Download updated packages
```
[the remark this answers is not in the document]
slagle: Ok, I see it's part of the revert, but don't you need to add state:latest to the task at line 3?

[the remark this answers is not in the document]
bogdando: my env set up for adoption testing showed ovs rpms conflicts, when we have ensure latest set for 'Download needed packages' (line 3)

roles/edpm_ovs/tasks/download_cache.yml
bogdando: that was exactly the source problem for ovs packages repos conflicts blocking adoption development

SeanMooney: yes the latest is imporant

what im trying to fix here is the fact that if we do a dnf update we have no ability to pin the package to a specific version

by swapping to latest "openvswitch" will install the latest version available
but we can override the default packages with openvswtich27 or openvswitch32
to pin to 2.7 or 3.2

my assertion is we should use latest whenever we are installing packages and then pin the package via ansible_vars if required during adoption

this approach means that minor update is just a redeployment.

roles/edpm_bootstrap/tasks/download_cache.yml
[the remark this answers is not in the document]
slagle: I want to make sure that download_cache.yml caches the right versions that will be installed by packages.yml. I assume that if packages.yml has state:latest, then it's also needed in this file for the task at line 3. That needs to be tested.

bogdando:
please let's check if the user exists and only created it if it does not (tripleo created it for the source cloud)

SeanMooney:
> I totally got the revert part in the series as that cause issues with ovs update, and ideally ovs update should be done considering dataplane outages, tripleo time there used to be tripleo_ovs_update module for that case.
> 
> To me not clear what's the requirement to always install latest packages whenever these roles get's executed/reexecuted? those bits seems to me only done with update/upgrade phases are executed.

we don't currently have update/upgrade phases and unless we have miscondicored our repos i would like to propose that we do minor update simply with a normal deploy.
if we use latest then that is all that is required. we don't need to have any special upgrade service to do minor upgrade.

downstream we will be configuring the subscription to a given stream i.e. in the past it would have been 16.1 or 16.2.

it should be possible to update to the latest rpm in a given stream.

using latest you should be able to pin the packages 1 of 2 ways
either by overriding the package name in ansible_vars to a versioned packs
or by deploying with your repos pointed to satellite and managing them using that toolchain.

the original patch that added the effective dnf update was required to ensure the selinux package was upgrade to make libvirt decontierisation work

if we are concerned with using latest in general i could restrict this to the libvirt and or bootstrap roles and not ovs.

ie.. i can add openstack-selinux to the list of libvirt packages and use latest there to make sure we are always running with the latest libvirt/openstack-selinux.

roles/edpm_bootstrap/tasks/download_cache.yml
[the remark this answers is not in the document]
SeanMooney: i woudl have to check but i can certnely add it on line 6
and i should also add update_cache: true

karelyatin:
> miscondicored

Thanks for sharing the insights here.

I am missing full details about the openstack-selinux issue but may be that also arised mixing rhel 9.2 nodes and antelope rpms? If it was just for openstack-selinux update then the original patch should have restricted the updated only to the required package or if full update is done for any CI case then a flag could have added to do the update only for the CI part.
Also if there is need to update some specific package then that would be better to be done by bumping minimal version requirement for that package where ever that's deployed as a dependency.

Personally with the context i have right now for the problems mentioned i would avoid doing generic update as part of this patch.
Someone from update/upgrade team could share better insights on such global change.

roles/edpm_ovs/tasks/download_cache.yml
bogdando: @SeanMooney would you please update to latest?

bogdando (changes requested):
> Thanks for sharing the insights here.
> 
> I am missing full details about the openstack-selinux issue but may be that also arised mixing rhel 9.2 nodes and antelope rpms? If it was just for openstack-selinux update then the original patch should have restricted the updated only to the required package or if full update is done for any CI case then a flag could have added to do the update only for the CI part.
> Also if there is need to update some specific package then that would be better to be done by bumping minimal version requirement for that package where ever that's deployed as a dependency.
> 
> Personally with the context i have right now for the problems mentioned i would avoid doing generic update as part of this patch.
> Someone from update/upgrade team could share better insights on such global change.

Please let's reduce this change scope to only unblock libvirt decontainerization during adoption of workloads, leaving everything else aside

bogdando:
As this work needs more inputs from OVN and upgrade teams, I suggest to simply revert the breaking change https://github.com/openstack-k8s-operators/edpm-ansible/pull/531 instead

SeanMooney:
> > miscondicored
> 
> Thanks for sharing the insights here.
> 
> I am missing full details about the openstack-selinux issue but may be that also arised mixing rhel 9.2 nodes and antelope rpms? If it was just for openstack-selinux update then the original patch should have restricted the updated only to the required package or if full update is done for any CI case then a flag could have added to do the update only for the CI part. Also if there is need to update some specific package then that would be better to be done by bumping minimal version requirement for that package where ever that's deployed as a dependency.
> 
> Personally with the context i have right now for the problems mentioned i would avoid doing generic update as part of this patch. Someone from update/upgrade team could share better insights on such global change.

no the issue was because for installer provsioned image and adoption openstack-selinux is alreway installed but it was using an old version that was incompatble with host installed libvirt

so we need to make sure that the greenfiled code always upgrades it.

we cant do a min version bump for two reasons one its not an indrect depency of another package its a direct depenciy we are installing.
second while we did update the policy a formal release was not done with a new version. just a new version fo the package was built and published like an async

form my standpoint we are starting form a point where we update every package on the system with a dnf update. in this change i am restricting the package update to only the packages we installed. so im not introducing the updating of all packages im restricing it to only the packages we install and providing a way for the adopting folks to pin which is not in the current code.

a straight revert without updating the openstack-selinux package will not work for adoption as the existing rpm form 17.1 does not support deconteinerised libvirt so at a very minium we need to update openstack-selinux explicitly to latest.

SeanMooney:
as a compromise solution, i have reverted the ovs behaviour back to present instead of latest.
i still think latest is the correct approach but i won't have time to work on edpm again for a few weeks so i would like to proceed with this in its current form if there is no objection or hand this off to someone else. i have upstream features in nova that need to be completed before the feature freeze at the end of February so i will be exclusively working on that form this week until then.

bogdando:
> If it was just for openstack-selinux update then the original patch should have restricted the updated only to the required package

global dnf upgrade was alse needed for periodic jobs pipeline AFAIK

> or if full update is done for any CI case then a flag could have added to do the update only for the CI part.

perhaps, this i why I suggested to revert the initial change and address periodic pipeline in ci-framework scope.

> Also if there is need to update some specific package then that would be better to be done by bumping minimal version requirement for that package where ever that's deployed as a dependency.

this is how openstack-selinux gets updated already, we don't call for its upgrade command explicitly

roles/edpm_bootstrap/tasks/packages.yml
```diff
@@ -39,7 +39,7 @@
 - name: Install required packages to bootstrap EDPM
   ansible.builtin.dnf:
     name: "{{ edpm_bootstrap_packages_bootstrap }}"
-    state: present
+    state: latest
```
bogdando, line 42: lgtm, if this doesn't break the periodic pipeline, like a clean revert would do

SeanMooney:
check-rdo ERROR! Error when getting collection version metadata for community.general:5.6.0 from default (https://galaxy.ansible.com/api/) (HTTP Code: 504, Message: Gateway Time-out Code: Unknown)

roles/edpm_bootstrap/tasks/packages.yml
slagle: After further thought, and some [discussion on slack](https://redhat-internal.slack.com/archives/CQXJFGMK6/p1705590739161989), I think we need to halt on the approach to switching to use state:latest. We need to think carefully about if we want every ansible execution to perform a minor update. I don't think we do, so I'd like to hold this PR due to that for the time being.

If we can fix the specific issue related to decontainerized libvirt and openstack-selinux with an adoption specific fix (I think that's the intent of https://github.com/openstack-k8s-operators/edpm-ansible/pull/539), then let's go with that for now.

slagle (dismissed):

playbooks/configure_network.yml
```diff
@@ -7,6 +7,11 @@
   any_errors_fatal: "{{ edpm_any_errors_fatal | default(true) }}"
   max_fail_percentage: "{{ edpm_max_fail_percentage | default(0) }}"
   tasks:
+    - name: Import edpm_ovs to install ovs packages
+      ansible.builtin.import_role:
+        name: osp.edpm.edpm_ovs
+      tags:
+        - edpm_network_config
```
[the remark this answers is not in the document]
slagle: edpm_network_config is for the edpm_network_config role (see the next task), so this should be edpm_ovs

slagle (commented):
thanks for the update. just one comment about the tag

roles/edpm_ovs/molecule/default/README.md
```diff
@@ -0,0 +1,18 @@
+************
+edpm_libvirt
+************
+
```
karelyatin, line 4: The README needs to be updated for edpm_ovs role

roles/edpm_ovs/meta/main.yml
```diff
@@ -0,0 +1,42 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+galaxy_info:
+  author: OpenStack
+  description: EDPM OpenStack Role -- edpm_libvirt
```
karelyatin, line 20: edpm_ovs

roles/edpm_ovs/tasks/install.yml
```diff
@@ -0,0 +1,23 @@
...
+    - ovs
+  become: true
+  ansible.builtin.package:
+    name: "{{ item }}"
+    state: present
+  loop: "{{ edpm_ovs_packages }}"
+  notify:
+    - Restart ovs
+
+- name: Ensure ovs services are enabled and running
```
karelyatin, line 14: may be add state: started here itself(also to match it with task name) instead of adding a "Restart" notify handler and using flush_handlers?

SeanMooney:
> thanks for the update. just one comment about the tag
oh i had fixed that in i guess one of the commits i dropped ill fix it again, no worries thanks for the review

roles/edpm_ovs/meta/main.yml
SeanMooney: good spot. can you geuess what role i used as a basis of this :)
will fix

roles/edpm_ovs/molecule/default/README.md
SeanMooney: yep i tought i did this already but i guess i missed some ill do that now

roles/edpm_ovs/tasks/install.yml
SeanMooney: so im trying to prefer using the handeler over using started in general to have a single way to restart/start the service once if we have configtion action that require the process to be bounded.

in the case of ovs most of its config is done by the ovs db so we can change some of it at run time but not all its config.

since I'm intentionally not doing any config here and derfing that to os-net-config i guess swaping to start might be fine for ovs so ill make that refactoring but for service that are config file based i think the handeler approach is cleaner.

slagle (commented):
+2, this LGTM now

bogdando (commented):
lgtm

bogdando (approved):

roles/edpm_ovs/tasks/user.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Configure openvswitch user and group on the host
+  ansible.builtin.import_role:
+    name: edpm_users
+  vars:
+    edpm_users_users:
+      # 42476 is matching with the uid and gid created by kolla in the openstack containers
+      - {"name": "openvswitch", "uid": "42476", "gid": "42476", "shell": "/sbin/nologin", groups: "hugetlbfs", "comment": "openvswitch user"}
```
rabi, line 8: This has broken the baremetal job. It's trying to change the uid when `openvswitch` has been already installed and running and openvswitch user has a different uid (which is the case with the edpm images used in baremetal job). It fails with the below error. This PR has passed CI, so I guess the job is not using images built with the PR. 

https://logserver.rdoproject.org/14/614/91067d34d1c3d575e9f1151d3b043dcf5b404911/github-check/cifmw-crc-podified-edpm-baremetal/c8d647b/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/configure-network-openstack-edpm-ipam-942ls/logs/openstackansibleee.log

```
Monday 22 January 2024  10:19:04 +0000 (0:00:03.313)       0:00:06.826 ******** 
[0;31mfatal: [edpm-compute-0]: FAILED! => {"changed": false, "msg": "usermod: user openvswitch is currently used by process 1191\n", "name": "openvswitch", "rc": 8}[0m
```
`openvswitch:x:987:42476:Open vSwitch Daemons:/:/sbin/nologin
`

bogdando: There is a fix https://github.com/openstack-k8s-operators/edpm-ansible/pull/540/files
and it seems I should switch it back to using try/except blocks?

bogdando: > which is the case with the edpm images used in baremetal job

Shouldn't also that BM job using that same ID, that's likely originates from kolla users mappings?

rabi: > There is a fix https://github.com/openstack-k8s-operators/edpm-ansible/pull/540/files

I don't think the change here is relevant to the issue. AFAIK user would be created during installation and edpm image used in baremetal job already installs openvswitch[1]. IMO we can get rid of the user creation completely here.

[1] https://github.com/openstack-k8s-operators/edpm-image-builder/blob/main/images/edpm-hardened-uefi-centos-9-stream.yaml#L21

SeanMooney: we shoudl not be installing OVS in the baremetal image as not all EDPM nodes should have ovs install. 

so we should remove the ovs installation at https://github.com/openstack-k8s-operators/edpm-image-builder/blob/main/images/edpm-hardened-uefi-centos-9-stream.yaml#L21

SeanMooney: it looks like the fix i had added to only set the user if its not already created got lost when i removed the last two commits form this pr...

lets proceed with #540

rabi: > we shoudl not be installing OVS in the baremetal image as not all EDPM nodes should have ovs install.

If it's not needed in the base images, it can be dropped from edpm-image-builder. However, I don't understand why we want to unnecessarily have the check to create the user, as what I can see it would surely be created with when installing the rpm.

[1] https://pkgs.devel.redhat.com/git/rpms/openvswitch/tree/openvswitch.spec?h=rhel-9.0.0#n329

SeanMooney: if the rpm was correctly pinnign the user to the one expect on OpenStack system you would not have had the ci issue.

the container expect ovs to have a specific user and group that this not the same as the one used by rhel when packaging ovs.

the ovs we install in our openstack distribution comes form the fast datapath repository
and is not part of our openstack repos so it is not packaged to run as the users we use historically in our product when building the container.

as a result we need to ensure we create the user with the expect user id and groups before the rpm is installed.

bogdando: mismatching IDs will hit us if we'd want to use UNIX sockets for ovn/ovs/nova compute interactions, IIUC. But this is good the fix doesn't change the status quo at least

bogdando: the fix helps with "upgrade" of workloads during adoption, where we don't touch OVS bits yet

SeanMooney: creating the user/group with the uid/group id specified in tcib is not for upgrades
it to fix greenfield to ensure they are assigned.

skipping the user/group creation in #540 is a workaround for the fact the tcib/tripleo uid/gid are out of sync.

we can and likely should correct the uid/gid in tcib 
https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/kolla/base/uid_gid_manage.sh#L62
to match those used in tripleo

https://github.com/openstack/tripleo-common/blob/stable/wallaby/container-images/kolla/base/uid_gid_manage.sh#L66

however looking at stable wallaby we used the same user/group id dere as tcib is using today.

_SUPPORTED_USERS['openvswitch']='openvswitch 42476 42476'

it is also the same in stable/train

https://github.com/openstack/tripleo-common/blob/stable/train/container-images/kolla/base/uid_gid_manage.sh#L65

so either the users in the container and the hsot have always been broken or there is something else happening to correct this that I'm not aware of.

PR #533, opened 2024-01-10
OVN role arguments refactor

jpodivin:
We really don't need to expose all of these. In many cases we were essentially setting customer up for configuration conflicts if they were to supply different values in multiple ways. This way we have much cleaner interface.

### DIFF
--- a/roles/edpm_ovn/meta/argument_specs.yml
+++ b/roles/edpm_ovn/meta/argument_specs.yml
@@ -27,44 +27,20 @@ argument_specs:
         type: bool
       edpm_enable_internal_tls:
         default: false
-        description: ''
+        description: >
+          Should OVN use tls for default protocol?
+          If set to false, the OVN will default to tcp.
         type: bool
       edpm_ovn_bridge:
         default: br-int
-        description: ''
+        description: >
+          Passed to the `edpm_ovn_ovs_external_ids` as value for the `ovn-bridge`.
         type: str
       edpm_ovn_bridge_mappings:
         default:
           - datacentre:br-ex
         description: ''
         type: list
-      edpm_ovn_chassis_mac_first_octets:
-        default:
-          - 0e
-          - 1e
-          - 2e
-          - 3e
-        description: ''
-        type: list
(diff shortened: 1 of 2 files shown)


jpodivin:
recheck

`error: You must be logged in to the server (Unauthorized)`

jpodivin:
recheck

retry limit

roles/edpm_ovn/meta/argument_specs.yml
```diff
@@ -105,35 +81,18 @@ argument_specs:
...
-          ovn-monitor-all: true
-          ovn-ofctrl-wait-before-clear: '{{ edpm_ovn_ofctrl_wait_before_clear }}'
-          ovn-remote: '{% set db_addresses = [] %}{% for host in edpm_ovn_dbs %}{{ db_addresses.append([edpm_ovn_protocol,
-            host, edpm_ovn_sb_server_port] | join('':'')) }}{% endfor %}{{ db_addresses
-            | join('','') }}'
-          ovn-remote-probe-interval: '{{ edpm_ovn_remote_probe_interval }}'
-          rundir: /var/run/openvswitch
-        description: Sets external_id data from provided variables using Jinja templating
-        type: dict
+      ovn_match_northd_version:
```
karelyatin, line 84: where these used? also naming convention differs - vs _

roles/edpm_ovn/meta/argument_specs.yml
```diff
@@ -105,35 +81,18 @@ argument_specs:
         default: 8000
         description: ''
```
karelyatin, line 82: So just for clarity, the arguments which are not in argument_specs, if still passed those will be noop or will get considered?

jpodivin: If you pass them to the role it will produce error. Argspecs are something like function API, in this context.

roles/edpm_ovn/meta/argument_specs.yml
jpodivin: Thanks for catching this, variable is now passed properly.

slawqo (approved):

PR #534, opened 2024-01-11
Make edpm_user the default for nova persistent directories

fao89:
see https://github.com/openstack-k8s-operators/data-plane-adoption/pull/219#discussion_r1446079023

### DIFF
--- a/roles/edpm_nova/tasks/configure.yml
+++ b/roles/edpm_nova/tasks/configure.yml
@@ -24,8 +24,8 @@
     path: "{{ item.path }}"
     setype: "container_file_t"
     state: "directory"
-    owner: "{{ item.owner | default(omit) }}"
-    group: "{{ item.group | default(omit) }}"
+    owner: "{{ item.owner | default(ansible_user) }}"
+    group: "{{ item.group | default(ansible_user) }}"
     mode: "{{ item.mode | default(omit) }}"
   loop:
     - {"path": "/var/lib/nova", "mode": "0755"}


fao89:
recheck

bogdando:
I think this is correct, let's wait for @SeanMooney review

fao89:
recheck

roles/edpm_nova/tasks/configure.yml
```diff
@@ -24,8 +24,8 @@
     path: "{{ item.path }}"
     setype: "container_file_t"
     state: "directory"
-    owner: "{{ item.owner | default(omit) }}"
-    group: "{{ item.group | default(omit) }}"
+    owner: "{{ item.owner | default(ansible_user) }}"
+    group: "{{ item.group | default(ansible_user) }}"
```
SeanMooney, line 28: default omit was used so that we did not modify the permission of the diretores unless we specified them

we could make this ansible_user but it should not be required.

if we want to proceed with this then /var/lib/nova, /var/lib/nova/instances and  "/var/log/containers/nova", should probably be set to the nova user

SeanMooney:
in general i dont think this is correct as written the nova persistent dirs should be owned by the nova user

they are set to be owned by nova when the nova-compute container starts by kolla today.

roles/edpm_nova/tasks/configure.yml
fao89: what do you think is the best way to handle https://github.com/openstack-k8s-operators/data-plane-adoption/pull/219#discussion_r1446963270 ?

PR #535, opened 2024-01-11
Nova compute parameters changes

Jaganathancse:
This change is to provide option to configure nova compute parameters config as edpm ansible variables.

### DIFF
--- a/roles/edpm_nova/tasks/configure_compute.yml
+++ b/roles/edpm_nova/tasks/configure_compute.yml
@@ -1,0 +1,67 @@
+---
+# Copyright 2024 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Configure compute configs
+  block:
+    - when: edpm_nova_cpu_shared_set|string
+      name: Set CPU shared set
+      community.general.ini_file:
+        path: "{{ edpm_nova_config_dest }}/03_compute_config.conf"
+        setype: container_file_t
+        mode: "0644"
+        create: true
+        section: "compute"
+        option: "cpu_shared_set"
+        value: "{{ edpm_nova_cpu_shared_set }}"
+
+    - when: edpm_nova_cpu_dedicated_set|string
+      name: Set CPU dedicated set
+      community.general.ini_file:
+        path: "{{ edpm_nova_config_dest }}/03_compute_config.conf"
+        setype: container_file_t
+        mode: "0644"
+        create: true
+        section: "compute"
+        option: "cpu_dedicated_set"
+        value: "{{ edpm_nova_cpu_dedicated_set }}"
+
+    - when: edpm_nova_reserved_host_memory|int > 0
+      name: Set reserved host memory
+      community.general.ini_file:
+        path: "{{ edpm_nova_config_dest }}/03_compute_config.conf"
+        setype: container_file_t
+        mode: "0644"
+        create: true
+        section: "compute"
+        option: "reserved_host_memory_mb"
+        value: "{{ edpm_nova_reserved_host_memory }}"
+  notify:
+    - Restart nova
+
+- name: Configure SRIOV PCI Passthrough
+  when: (edpm_nova_pci_passthrough | list | length) > 0
+  community.general.ini_file:
+    path: "{{ edpm_nova_config_dest }}/04_sriov_config.conf"
+    setype: container_file_t
+    mode: "0644"
+    create: true
+    section: "pci"
+    option: "passthrough_whitelist"
+    value: "{{ item | to_json }}"
+    exclusive: false
+  loop: "{{ edpm_nova_pci_passthrough | list }}"
+  notify:
+    - Restart nova
--- a/roles/edpm_nova/defaults/main.yml
+++ b/roles/edpm_nova/defaults/main.yml
@@ -31,3 +31,8 @@ edpm_nova_compute_image: "quay.io/podified-antelope-centos9/openstack-nova-compu
 
 # certs
 edpm_nova_tls_certs_enabled: "{{ edpm_tls_certs_enabled | default(False) }}"
+
+# compute parameters
+edpm_nova_cpu_shared_set: ""
+edpm_nova_cpu_dedicated_set: ""
+edpm_nova_reserved_host_memory: 4096
--- a/roles/edpm_nova/meta/argument_specs.yml
+++ b/roles/edpm_nova/meta/argument_specs.yml
@@ -37,3 +37,25 @@ argument_specs:
           This image is used to create the nova compute container
           on the compute node and generaly will be set via the container
           bundle by the dataplane operator.
+      edpm_nova_cpu_shared_set:
+        type: str
+        default: ""
+        description: |
+          Set to a comma-separated string or range of physical host
+          CPU numbers used to.
+      edpm_nova_cpu_dedicated_set:
+        type: str
+        default: ""
+        description: |
+          Set to a comma-separated string or range of host CPU cores to which
+          processes for pinned instance CPUs (PCPUs) can be scheduled.
+      edpm_nova_reserved_host_memory:
+        type: int
+        default: 4096
+        description: Reserved RAM for host processes
+      edpm_nova_pci_passthrough:
+        type: list
+        default: []
+        description: |
+          Set to list of SRIOV PCI passthrough dictinary with option's like
+          devname, physical network and trusted etc..
--- a/roles/edpm_nova/tasks/main.yml
+++ b/roles/edpm_nova/tasks/main.yml
@@ -18,6 +18,9 @@
 - name: Configure nova
   ansible.builtin.include_tasks: configure.yml
 
+- name: Configure compute parameters
+  ansible.builtin.include_tasks: configure_compute.yml
+
 - name: Install nova
   ansible.builtin.include_tasks: install.yml
 


slagle (commented):
We had some functionality similar to this in the previous edpm_nova_compute role. Although I don't completely agree, I think it was an intentional design choice to not expose nova configuration with ansible vars any longer. @SeanMooney @gibizer WDYT, can you provide input on this one?

gibizer (commented):
I disagree with this change. Nova configuration can be passed via ConfigMaps or Secrets in the DataPlaneService, aligned with the documented strategy in https://github.com/openstack-k8s-operators/docs/blob/main/service_config.md 

* Having multiple ways to configure the same thing makes understanding the nova configuration hard during troubleshooting.
* Adding ansible vars to each nova configuration parameter is creating a lot of unnecessary work.

Jaganathancse:
> I disagree with this change. Nova configuration can be passed via ConfigMaps or Secrets in the DataPlaneService, aligned with the documented strategy in https://github.com/openstack-k8s-operators/docs/blob/main/service_config.md
> 
> * Having multiple ways to configure the same thing makes understanding the nova configuration hard during troubleshooting.
> * Adding ansible vars to each nova configuration parameter is creating a lot of unnecessary work.

@gibizer 
Created this PR to provide the edpm ansible variables to configure compute and pci section related option's in nova conf and also avoid duplicating openstackdataplaneservice like 'nova-custom-sriov' for multiple roles.
This PR is also required for OVN DPDK compute related configs and also  derive PCI whitelist automation for NIC Partitioning.

gibizer:
> > I disagree with this change. Nova configuration can be passed via ConfigMaps or Secrets in the DataPlaneService, aligned with the documented strategy in https://github.com/openstack-k8s-operators/docs/blob/main/service_config.md
> > 
> > * Having multiple ways to configure the same thing makes understanding the nova configuration hard during troubleshooting.
> > * Adding ansible vars to each nova configuration parameter is creating a lot of unnecessary work.
> 
> @gibizer Created this PR to provide the edpm ansible variables to configure compute and pci section related option's in nova conf and also avoid duplicating openstackdataplaneservice like 'nova-custom-sriov' for multiple roles. This PR is also required for OVN DPDK compute related configs and also derive PCI whitelist automation for NIC Partitioning.

I disagree to duplicate each nova configuration variable as ansible variable. We don't expose nova config variables as CRD fields in the control plane side either. Instead we allow the user to pass through full config file snippets via CMs or Secrets. So please do the configuration via adding CM/Secrets to the DataPlaneService.  This PR is not required for your DPDK and PCI work, you can use the already working way to pass configuration to nova compute via the DataPlaneService.

gibizer:
As a note. Nova [config doc](https://docs.openstack.org/nova/latest/configuration/config.html) lists 1446 config variables. I don't think you want to add each as an edpm variable. That is unmanageable. Please use the existing interface to pass config

Jaganathancse:
> > > I disagree with this change. Nova configuration can be passed via ConfigMaps or Secrets in the DataPlaneService, aligned with the documented strategy in https://github.com/openstack-k8s-operators/docs/blob/main/service_config.md
> > > 
> > > * Having multiple ways to configure the same thing makes understanding the nova configuration hard during troubleshooting.
> > > * Adding ansible vars to each nova configuration parameter is creating a lot of unnecessary work.
> > 
> > 
> > @gibizer Created this PR to provide the edpm ansible variables to configure compute and pci section related option's in nova conf and also avoid duplicating openstackdataplaneservice like 'nova-custom-sriov' for multiple roles. This PR is also required for OVN DPDK compute related configs and also derive PCI whitelist automation for NIC Partitioning.
> 
> I disagree to duplicate each nova configuration variable as ansible variable. We don't expose nova config variables as CRD fields in the control plane side either. Instead we allow the user to pass through full config file snippets via CMs or Secrets. So please do the configuration via adding CM/Secrets to the DataPlaneService. This PR is not required for your DPDK and PCI work, you can use the already working way to pass configuration to nova compute via the DataPlaneService.

@gibizer
Need to derive pci passthrough whitelist based on compute dynamically for SRIOV Nic Partitioning. we need to use edpm-ansible for module scripts.
Whether we need to use VA  with kustomization.yaml for automation also?

gibizer:
> > > > I disagree with this change. Nova configuration can be passed via ConfigMaps or Secrets in the DataPlaneService, aligned with the documented strategy in https://github.com/openstack-k8s-operators/docs/blob/main/service_config.md
> > > > 
> > > > * Having multiple ways to configure the same thing makes understanding the nova configuration hard during troubleshooting.
> > > > * Adding ansible vars to each nova configuration parameter is creating a lot of unnecessary work.
> > > 
> > > 
> > > @gibizer Created this PR to provide the edpm ansible variables to configure compute and pci section related option's in nova conf and also avoid duplicating openstackdataplaneservice like 'nova-custom-sriov' for multiple roles. This PR is also required for OVN DPDK compute related configs and also derive PCI whitelist automation for NIC Partitioning.
> > 
> > 
> > I disagree to duplicate each nova configuration variable as ansible variable. We don't expose nova config variables as CRD fields in the control plane side either. Instead we allow the user to pass through full config file snippets via CMs or Secrets. So please do the configuration via adding CM/Secrets to the DataPlaneService. This PR is not required for your DPDK and PCI work, you can use the already working way to pass configuration to nova compute via the DataPlaneService.
> 
> @gibizer Need to derive pci passthrough whitelist based on compute dynamically for SRIOV Nic Partitioning. we need to use edpm-ansible for module scripts. Whether we need to use VA with kustomization.yaml for automation also?

Do you have document or PR I can look at that describe what you need to / want to do with pci passthrough whitelist generation?

SeanMooney:
doing config generation on the compute node via ansible is expiclty against our design goals so we should not do this.

SeanMooney:
> We had some functionality similar to this in the previous edpm_nova_compute role. Although I don't completely agree, I think it was an intentional design choice to not expose nova configuration with ansible vars any longer. @SeanMooney @gibizer WDYT, can you provide input on this one?

yes it was an intentional choice and i would like to convert the neutron roles to follow the same pattern we have established eventually.

slagle:
> As a note. Nova [config doc](https://docs.openstack.org/nova/latest/configuration/config.html) lists 1446 config variables. I don't think you want to add each as an edpm variable. That is unmanageable. Please use the existing interface to pass config

We don't need ansible variables to map one to one to the 1446 nova config variables, in order to do configuration with ansible. I hope that's clear. If that's a reason for supporting the design choice, I'd be happy to show why that's not the case.

"We only want to have 1 way to do configuration"...is a reason. I don't totally agree with it, but I understand the reasoning why that choice was made.

roles/edpm_nova/tasks/configure_compute.yml
```diff
@@ -0,0 +1,67 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Configure compute configs
+  block:
+    - when: edpm_nova_cpu_shared_set|string
+      name: Set CPU shared set
+      community.general.ini_file:
```
SeanMooney, line 21: communit.general is not part of the validated content supported in our downstream distobution fo ansible

https://access.redhat.com/support/articles/ansible-automation-platform-certified-content
and per https://access.redhat.com/articles/4916901
that means that there is no official support for those modules in our product.

roles/edpm_nova/tasks/configure_compute.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    - when: edpm_nova_reserved_host_memory|int > 0
+      name: Set reserved host memory
+      community.general.ini_file:
+        path: "{{ edpm_nova_config_dest }}/03_compute_config.conf"
+        setype: container_file_t
+        mode: "0644"
+        create: true
+        section: "compute"
+        option: "reserved_host_memory_mb"
```
SeanMooney, line 49: this is in the Default section and it is not enough to reserve memory when using CPU pinnign

roles/edpm_nova/tasks/configure_compute.yml
```diff
@@ -0,0 +1,67 @@
...
+    setype: container_file_t
+    mode: "0644"
+    create: true
+    section: "pci"
+    option: "passthrough_whitelist"
+    value: "{{ item | to_json }}"
+    exclusive: false
+  loop: "{{ edpm_nova_pci_passthrough | list }}"
+  notify:
+    - Restart nova
```
SeanMooney, line 67: the current proposal is also not correct.
passthrough_whitelist is deprecated and we had a company wide goal to remove all reference to blacklist/whitelist form our product device_spec is the replacement name.

https://docs.openstack.org/nova/latest/configuration/config.html#pci.device_spec and passthrough_whitelist were both multi opt values meaning that 
the behavior is different from in format in that each time is merged.

```
- name: Configure SRIOV PCI Passthrough
  when: (edpm_nova_pci_passthrough | list | length) > 0
  community.general.ini_file:
    path: "{{ edpm_nova_config_dest }}/04_sriov_config.conf"
    setype: container_file_t
    mode: "0644"
    create: true
    section: "pci"
    option: "passthrough_whitelist"
    value: "{{ item | to_json }}"
    exclusive: false
  loop: "{{ edpm_nova_pci_passthrough | list }}"
  notify:
    - Restart nova
 ```

 exclusive: false

changes the behaviour somewhat
```
If set to false, only the specified [value](https://docs.ansible.com/ansible/latest/collections/community/general/ini_file_module.html#ansible-collections-community-general-ini-file-module-parameter-value)/[values](https://docs.ansible.com/ansible/latest/collections/community/general/ini_file_module.html#ansible-collections-community-general-ini-file-module-parameter-values) are added when [state=present](https://docs.ansible.com/ansible/latest/collections/community/general/ini_file_module.html#ansible-collections-community-general-ini-file-module-parameter-state), or removed when [state=absent](https://docs.ansible.com/ansible/latest/collections/community/general/ini_file_module.html#ansible-collections-community-general-ini-file-module-parameter-state), and existing ones are not modified.
```
the default state is present so this would just addtivly append to the file across multiple iterations if we change the input values.

SeanMooney:
> > As a note. Nova [config doc](https://docs.openstack.org/nova/latest/configuration/config.html) lists 1446 config variables. I don't think you want to add each as an edpm variable. That is unmanageable. Please use the existing interface to pass config
> 
> We don't need ansible variables to map one to one to the 1446 nova config variables, in order to do configuration with ansible. I hope that's clear. If that's a reason for supporting the design choice, I'd be happy to show why that's not the case.
> 
> "We only want to have 1 way to do configuration"...is a reason. I don't totally agree with it, but I understand the reasoning why that choice was made.

the other reasons i don't want to just add this with out detailed discussion is when we were discussing this in the past we had explicit instruction from pm  that we do not internet to provide an introspection interface in the the datapalne operator and we should not automate dynamically caluslting CPU pinning or PCI device filtering.

this was one of the example usecase we discussed when comparing the two ways for the control plane and datapalne to interface.


while its not perfect i would much rather proceed with the method doucmente  in the sriov_va

https://github.com/openstack-k8s-operators/architecture/pull/76

gibizer:
> > As a note. Nova [config doc](https://docs.openstack.org/nova/latest/configuration/config.html) lists 1446 config variables. I don't think you want to add each as an edpm variable. That is unmanageable. Please use the existing interface to pass config
> 
> We don't need ansible variables to map one to one to the 1446 nova config variables, in order to do configuration with ansible. I hope that's clear. If that's a reason for supporting the design choice, I'd be happy to show why that's not the case.
> 
> "We only want to have 1 way to do configuration"...is a reason. I don't totally agree with it, but I understand the reasoning why that choice was made.

My comment about the 1446 nova config variable probably make more sense if you consider that I want to keep a single way for the end user to pass configuration to nova. If there is a single way, and that is edpm ansible variables, then we need to model a lot of variables (I agree we might not need all 1446) not just the 3 proposed today.

PR #540, opened 2024-01-19
Create host user and group, if not exists

bogdando:
Create host user and group, if not exists
    
    During Wallaby to Anteleope FFU (EDPM adoption), users that exist
    should not be attempted to get changed IDs, home paths, or the like.
    
    In particular, changing openwswitch user ID to a constant agreed upon
    for OSP-related services use from Kolla, breaks its upgrades.
    
    Use getent to check if a user being created is already exists.
    Only create it in a rescue block (getent fails if user does not exist).
    While the user module is idempotent per se, it doesn't allow changin
    user IDs.
    
    This fixes ovs edpm role during EDPM adoption, in particular.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/702

### DIFF
--- a/roles/edpm_users/tasks/create_users_and_groups.yml
+++ b/roles/edpm_users/tasks/create_users_and_groups.yml
@@ -14,11 +14,16 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Create host user and group
+- name: Create host user and group, if not exists
   become: true
   tags:
     - edpm_users
   block:
+    - name: Detect when user exists and cannot be created
+      ansible.builtin.getent:
+        database: passwd
+        key: "{{ item.name }}"
+  rescue:
     - name: Ensure group is present on the host [ {{ item.name }} ]
       ansible.builtin.group:
         name: "{{ item.name }}"


bogdando:
tested this on my adoption dev env

jpodivin (commented):
I'm slightly confused here. The `group` and `user` modules should already take care of this check. Their behavior is meant to be idempotent. If it is not we should report a bug to ansible devs.

bogdando:
> I'm slightly confused here. The `group` and `user` modules should already take care of this check. Their behavior is meant to be idempotent. If it is not we should report a bug to ansible devs.

Please see https://github.com/openstack-k8s-operators/edpm-ansible/pull/532#pullrequestreview-1836323967
We will need a try/except block to keep it "kolla FFU" compatible

bogdando:
@jpodivin I've reset the impl back to https://github.com/openstack-k8s-operators/edpm-ansible/commit/243047f089906e2ece0ec21f94e30e6241e5b0f2 , the rescue block to cover FFU cases

roles/edpm_users/tasks/create_users_and_groups.yml
```diff
@@ -14,11 +14,16 @@
...
+- name: Create host user and group, if not exists
   become: true
   tags:
     - edpm_users
   block:
+    - name: Detect when user exists and cannot be created
+      ansible.builtin.getent:
+        database: passwd
+        key: "{{ item.name }}"
+  rescue:
```
SeanMooney, line 26: right the code for the conditional creation 
https://github.com/openstack-k8s-operators/edpm-ansible/commit/eb342de63c06046dc2fcf645771e030a66b4027c#diff-a6614876ec5d5aafe6387d59d602cf4b5c7a7c1967b7d3e69e971b78c38fce70
was lost when i dropped the final 2 commits in the series


your approach is slightly different but it achieve the same result

SeanMooney (approved):

roles/edpm_users/tasks/create_users_and_groups.yml
bogdando: thanks for the link, whatever unblocks folks is WFM

jpodivin (approved):
Please make a note about this behavior somewhere in docs. We may hit something similar later.

bogdando:
> Please make a note about this behavior somewhere in docs. We may hit something similar later.

do you mean the adoption docs? this role provides a broader scope than just ovs and also covers greenfield

SeanMooney:
> So to clarify, the problem is we have a process owned by this user already running and we can't shut it down. Which is something that only occurs during adoption.

so the issue with ovs is if we shut it down it will cause a datapalne outage to the workload during adoption.

for ovs-dpdk in particular the dataplane is executed in userspace as part of the ovs-vswitchd process. for kernel and hardware offloaded ovs the ovs-vswichd is primarlly responsible for calculating and propagating the ovs flow to he kernel dataplaneor hardware dataplane but the first few packets of any new connection are also processed in user space. for kernel/hardware ovs stoping the ovs-vswticd wont terminate any active traffic imeditaly as the flow rules are cached in the offloaded dataplane but new connections wont be able to function and the cached roles will also eventually expire.

we may revisit this in the future and deciend that we ant to force them to be stop so we can align the uid/group id but we likely would want that to be opt in or a separate post adoption procedure.

by default we do not connect to ovs using the command line client form nova/neutron
we use TCP instead as a result if the users are missalinged that is not fatal however that is somewhat insecure as currently implemented so we may need to revert to using the UNIX sockets instead in the future which would then use the user/group id fo the socket to authenticate if we are in the appropriate user/group when connecting.

hardening access to ovs is a separate topic that we should discuss later for now, we want to avoid impacting the running service and only create the ovs user if it does not already exist.

roles/edpm_users/tasks/create_users_and_groups.yml
```diff
@@ -14,11 +14,16 @@
...
 # under the License.
 
-- name: Create host user and group
+- name: Create host user and group, if not exists
   become: true
   tags:
     - edpm_users
   block:
+    - name: Detect when user exists and cannot be created
+      ansible.builtin.getent:
```
rabi, line 23: I can see the user created[1] when the rpm is installed. What scenario do we expect the user not to be created during install and we need to do all this? 

[1] [1] https://pkgs.devel.redhat.com/git/rpms/openvswitch/tree/openvswitch.spec?h=rhel-9.0.0#n329

bogdando: when user exists, like comes from RPM spec, it will not be created by this role.

bogdando: do we need extra conditions and tasks here? For example, we can revert logic and add rescue block to user role instead, to make it bypassing creation of a user w/o error if there is that 'usermod' message in the error?

bogdando:
hm, BM job has failed https://review.rdoproject.org/zuul/build/9b8c45bee9254982a73acd69bc80717a

bogdando:
[cross posting](https://github.com/openstack-k8s-operators/edpm-ansible/pull/532#discussion_r1462034577) here a related question:

**Question**:
> this does not update the uid/gid, if the user/group already exist, so I don't understand how it would help with upgrade?
```
[tripleo-admin@compute-0 ~]$ id openvswitch
uid=989(openvswitch) gid=988(openvswitch) groups=988(openvswitch),987(hugetlbfs)

[tripleo-admin@compute-0 ~]$ sudo podman exec -ti ovn_controller /bin/bash
[root@compute-0 /]# id openvswitch
uid=993(openvswitch) gid=993(openvswitch) groups=993(openvswitch)
```

**Answer**:
the fix helps with "upgrade" of workloads during adoption, where we don't touch OVS bits yet. We might hit issues later on, during OVN/OVS ffu though (yet again that's status quo, not a regression)

bogdando:
recheck cifmw-crc-podified-edpm-baremetal

SeanMooney:
the ovn base contaiern is not invokeing uid_gid_manage to pin the uid/gid of the ovs user/group

https://github.com/openstack/tripleo-common/blob/stable/wallaby/container-images/tcib/base/ovn-base/ovn-base.yaml#L2

and the neurton container are only doing it for neutron 
https://github.com/openstack/tripleo-common/blob/stable/wallaby/container-images/tcib/base/ovn-base/ovn-base.yaml#L2

so this seam like its an existing bug in the tripleo container and the ovs uid/gid is not stable which can cause upgrade issues

we had to fix that for in the past for example for the nova_migration user

https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/kolla/base/uid_gid_manage.sh#L62

this tend to break mixed rhel if we rely on package installation order to provide the uid/gid when building the containers

https://bugzilla.redhat.com/show_bug.cgi?id=2181897
so we don't have to fix this now but this is likely technical debt that should be adressed.

bshephar:
Looks like the changes here aren't being executed in the CI job. It still has the old content in the failing configure-network pod:
https://review.rdoproject.org/zuul/build/9cdb540c6e2142b1a1b769f16970ebde/log/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/configure-network-openstack-edpm-ipam-svt85/logs/openstackansibleee.log#21-31

lewisdenny:
recheck

Weird, rechecking with node hold to check it out.

lewisdenny:
Looks like Zuul pulled the correct code [1] 

[1] https://review.rdoproject.org/zuul/build/9cdb540c6e2142b1a1b769f16970ebde/log/zuul-info/inventory.yaml#190

bshephar:
recheck

bshephar:
We put a new hold node for this job. Re-running CI jobs to capture the node and investigate failures.

rabi:
@karelyatin pointed out that https://github.com/openstack-k8s-operators/install_yamls/pull/650/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R740 added a change to patch it in `edpm_deploy_baremetal` target which would break it. Let's see if https://github.com/openstack-k8s-operators/install_yamls/pull/702 fixes the issue

rabi:
recheck

rabi:
recheck

PR #543, opened 2024-01-23
Support NVMEe-oF storage protocols

ASBishop:
First commit:

Revamp deploying storage services required by nova
    
    Storage services (namely the iSCSI and multipath daemons) required
    by nova are now deployed using a dedicated playbook, which in turn
    is imported by the nova playbook. This ensures the services only
    get installed on EDPM nodes hosting nova.

Second commit:

NVMe-oF support has these components:
- Load the nvme-fabrics kernel module
- Install the nvme-cli package on the host (which initializes several files in /etc/nvme, and also some udev rules)
- Bind mount /etc/nvme into the nova-compute container

Jira: [OSPRH-1579](https://issues.redhat.com//browse/OSPRH-1579)

### DIFF
--- a/roles/edpm_nova/templates/nova_compute.json.j2
+++ b/roles/edpm_nova/templates/nova_compute.json.j2
@@ -27,6 +27,7 @@
         "/etc/multipath:/etc/multipath:z",
         "/etc/multipath.conf:/etc/multipath.conf:ro",
         "/etc/iscsi:/etc/iscsi:ro",
+        "/etc/nvme:/etc/nvme",
         "/var/lib/openstack/config/ceph:/var/lib/kolla/config_files/ceph:ro",
         "/etc/ssh/ssh_known_hosts:/etc/ssh/ssh_known_hosts:ro"
     ]
--- a/playbooks/configure_os.yml
+++ b/playbooks/configure_os.yml
@@ -50,18 +50,6 @@
         tasks_from: configure.yml
       tags:
         - edpm_logrotate_crond
-    - name: Configure edpm_iscsid
-      ansible.builtin.import_role:
-        name: osp.edpm.edpm_iscsid
-        tasks_from: configure.yml
-      tags:
-        - edpm_iscsid
-    - name: Configure edpm_multipathd
-      ansible.builtin.import_role:
-        name: osp.edpm.edpm_multipathd
-        tasks_from: configure.yml
-      tags:
-        - edpm_multipathd
     - name: Configure nftables
       ansible.builtin.import_role:
         name: osp.edpm.edpm_nftables
--- a/playbooks/install_os.yml
+++ b/playbooks/install_os.yml
@@ -33,15 +33,3 @@
         tasks_from: install.yml
       tags:
         - edpm_logrotate_crond
-    - name: Install edpm_iscsid
-      ansible.builtin.import_role:
-        name: osp.edpm.edpm_iscsid
-        tasks_from: install.yml
-      tags:
-        - edpm_iscsid
-    - name: Install edpm_multipathd
-      ansible.builtin.import_role:
-        name: osp.edpm.edpm_multipathd
-        tasks_from: install.yml
-      tags:
-        - edpm_multipathd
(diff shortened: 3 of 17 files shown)


roles/edpm_bootstrap/defaults/main.yml
```diff
@@ -31,6 +31,8 @@ edpm_bootstrap_packages_bootstrap:
   - crudini
   - jq
   - nftables
+  - nvme-cli
+  - openvswitch
```
gibizer, line 35: I'm wondering if this is a rebase artifact

ASBishop: I assume so. I rebased on origin/main, got a conflict, and it was due to these lines. The pushed update fixed the conflict.

SeanMooney: ovs has been moved to its own role so you should drop it form here

SeanMooney: one other thing we shoudl not install this by default on all hosts.

not all hosts will have vms so we should only install nvme-cli on host that need it.

edpm nodes are not 1:1 mapping with compute nodes

so you should probably install this in a different role.

i also am not sure why the nvme-cli package is required on the host?
os-brick runs in the nova_compute container.

is there a package that just provides the kernel modules that you can use instead?

roles/edpm_nova/templates/nova_compute.json.j2
```diff
@@ -27,6 +27,7 @@
         "/etc/multipath:/etc/multipath:z",
         "/etc/multipath.conf:/etc/multipath.conf:ro",
         "/etc/iscsi:/etc/iscsi:ro",
+        "/etc/nvme:/etc/nvme:z",
```
SeanMooney, line 30: +1 to uncodiotnally adding this.

z will force /etc/nvme to be relabled as contaienr_t

so if anything on the host is going to use this that is probably incorrect

roles/edpm_bootstrap/defaults/main.yml
ASBishop: > one other thing we shoudl not install this by default on all hosts.
> 
> not all hosts will have vms so we should only install nvme-cli on host that need it.
> 
> edpm nodes are not 1:1 mapping with compute nodes
> 
> so you should probably install this in a different role.

Hmm, OK. Where is the host-to-role mapping maintained? I need to know how to get the role included on the appropriate host.

> i also am not sure why the nvme-cli package is required on the host? os-brick runs in the nova_compute container.
> 
> is there a package that just provides the kernel modules that you can use instead?

I tried to explain that in the commit message. The nvme-cli package initializes several files and udev rules.

SeanMooney: if you dont want to have a separate role i would add the package to the libvirt role instead.

SeanMooney: we are installing ceph-common to provide udev rules
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_libvirt/defaults/main.yml#L52

so nvme-cli would be similar in that regard

roles/edpm_nova/templates/nova_compute.json.j2
ASBishop: Ack, will change this in the next update.

roles/edpm_bootstrap/defaults/main.yml
ASBishop: I'm about to push an update that adds another commit, which will isolate the current iscsid and multipathd roles in their own playbook, to which I'll add (in the other commit) a new nvme role.

I'll invoke this new playbook directly from nova's existing playbook. That will:
- ensure storage stuff is only configured on nova nodes
- ensure storage stuff is always configured wherever nova is deployed
- eliminates the need for creating a new dataplane service

slagle (commented):
+2 LGTM

ASBishop:
recheck

gibizer (commented):
Looks OK to me but I will let Sean to take a look too.

fultonj (approved):

PR #545, opened 2024-01-30
Ensure hugetlbfs group exists

lewisdenny:
Due to the changes in [1] the openvswitch user is created before the
openvswitch package is installed. In RHEL systems the edpm_ovs role now
fails with [2]

This patch ensures the hugetlbfs group is created before the openvswitch
users is created and added to the hugetlbfs group.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/pull/532
[2]
```
task path: /usr/share/ansible/collections/ansible_collections/osp/edpm/roles/edpm_users/tasks/create_users_and_groups.yml:32
fatal: [compute-0]: FAILED! => {"changed": false, "msg": "Group hugetlbfs does not exist"}
```

JIRA: https://issues.redhat.com/browse/OSPCIX-173

### DIFF
--- a/roles/edpm_ovs/tasks/user.yml
+++ b/roles/edpm_ovs/tasks/user.yml
@@ -5,6 +5,7 @@
   vars:
     edpm_users_users:
       # 42476 is matching with the uid and gid created by kolla in the openstack containers
+      - {"name": "hugetlbfs", "gid": "42477", "group_only": true}
       - {"name": "openvswitch", "uid": "42476", "gid": "42476", "shell": "/sbin/nologin", groups: "hugetlbfs", "comment": "openvswitch user"}
     edpm_users_extra_dirs: []
   tags:


rabi:
I think `bootstrap` service is expected to create the group, if it does not exist which should run before this.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/playbooks/bootstrap.yml#L26-L30
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_users/defaults/main.yml#L28

roles/edpm_ovs/tasks/user.yml
```diff
@@ -5,6 +5,7 @@
   vars:
     edpm_users_users:
       # 42476 is matching with the uid and gid created by kolla in the openstack containers
+      - {"name": "hugetlbfs", "gid": "42477", "group_only": true}
```
marios, line 8: this change lgtm but is not being picked up in the test run (and I'm trying to understand why... zuul thinks it is checked out correctly, but it is not being used by the create users & groups tasks:

```
TASK [osp.edpm.edpm_users : Create host users and groups] **********************
[1;30mtask path: /usr/share/ansible/collections/ansible_collections/osp/edpm/roles/edpm_users/tasks/main.yml:17[0m
[0;36mincluded: /usr/share/ansible/collections/ansible_collections/osp/edpm/roles/edpm_users/tasks/create_users_and_groups.yml for compute-0 => (item={'name': 'openvswitch', 'uid': '42476', 'gid': '42476', 'shell': '/sbin/nologin', 'groups': 'hugetlbfs', 'comment': 'openvswitch user'})[0m
```

i see we have this in the role default https://github.com/openstack-k8s-operators/edpm-ansible/blob/c8ed1ac822813bfe27d42cd98573ab93cb25b68d/roles/edpm_users/defaults/main.yml#L26-L28 

```
edpm_users_users:
  - {"name": "qemu", "uid": "107", "gid": "107", "shell": "/sbin/nologin", "comment": "qemu user"}
  - {"name": "hugetlbfs", "gid": "42477", "group_only": true}
```

so what you're adding here looks correct eg the group id and since the openvswitch user needs that group to exist it makes sense that we'd create it here (it seems like that is a requirement and should have been included here earlier?)

marios (approved):
lgtm

roles/edpm_ovs/tasks/user.yml
rabi: This is not required. The issue was something else, fixed in https://github.com/openstack-k8s-operators/ci-framework/pull/1082

marios: oh @rabi it looks like this merged anyway after my review 

do we need to revert it then? I don't think it should hurt?

marios: as discussed in slack @lewisdenny will revert this

PR #547, opened 2024-01-30
Use journald as log_driver for the containers

jlarriba:
As highlighted by @SeanMooney on https://issues.redhat.com/browse/OSPRH-153, we would like Podman log behaviour to be the one provided by the journald log_driver, which has been the default for a while.

The centralized logging retrieval system for the EDPM will expect the logs for all the containers to be stored in the journal.

### DIFF
--- a/molecule/test-helpers/verify_podman.yaml
+++ b/molecule/test-helpers/verify_podman.yaml
@@ -23,13 +23,3 @@
           - running_containers.stdout_lines | length == 1
           - "'Up' in running_containers.stdout_lines | first"
         fail_msg: "Podman container {{ item }} is not running"
-    - name: Check the container stdout log exists {{ item }}
-      become: true
-      ansible.builtin.stat:
-        path: "/var/log/containers/stdouts/{{ item }}.log"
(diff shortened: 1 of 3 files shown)


rebtoor:
recheck

jlarriba:
The issue is caused, at least in nova, because the change I introduced is making tests fail. This test:

```
TASK [Assert the container stdout log exists nova_compute] *********************
fatal: [compute-1]: FAILED! => changed=false
assertion: log_exists.stat.exists
evaluated_to: false
msg: Container stdout log for nova_compute does not exist
```

It makes sense. As we change the log_driver to podman, the file "/var/log/containers/stdouts/{{ item }}.log" is not written.

I am removing that test and let that run again.

jlarriba:
I have decided to remove the test for the moment. Currently, nova does not write their logs to stdout so podman does not pick anything and send it to journald, so there is nothing actually to test.

An appropiate test can be written down when nova starts writing logs to stdout in the container.

jpodivin (approved):

PR #548, opened 2024-01-30
Allow hosts override from extra vars

jpodivin:
Sometimes we may want to apply our play to a subset of nodes, or to all nodes, even though it was originally targeting 'all' or just some of them.

Right now, all of our playbooks target all hosts by default, so the override can be hypothetically used to target subset of nodes. But eventually, when we have playbooks for specific groups in inventory, this will allow us to instead go for all of them.

We just need to document it.

### DIFF
--- a/docs/source/contributing_playbooks.rst
+++ b/docs/source/contributing_playbooks.rst
@@ -1,0 +1,88 @@
+Contributing Playbooks
+----------------------
+
+Addition of new playbooks to the edpm-ansible collection is easy,
+thanks to specialized development tooling and CI pipeline.
+However, certain requirements must be met by any new playbook,
+in order to ensure quality of the collection and a smooth operation
+of software consuming it.
+
+These requirements, outlined in the following document, go beyond
+syntactic and functional correctness of the code.
+
+Where there isn't an explicit requirement, or recommendation,
+provided by this document, the primary `Ansible documentation`_ is considered
+to be the source of truth. When this document and official Ansible docs
+are in substantial conflict, an issue should be raised.
+
+Playbook design principles
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Playbook must not duplicate other playbooks.
+
+If a newly developed playbook requires functionality not already present in existing
+roles, a new role should be submitted in the same PR implementing it.
+
+If possible, playbooks should not use tasks directly, only trough a role.
+
+Playbooks must use following basic layout. Exposing variables, `edpm_override_hosts`,
+`edpm_max_fail_percentage` and `edpm_any_errors_fatal`.
+
+.. code-block:: YAML
+
+    - name: NEWPLAYNAME
+      hosts: "{{ edpm_override_hosts | default('all') }}"
+      strategy: linear
+      become: true
+      any_errors_fatal: "{{ edpm_any_errors_fatal | default(true) }}"
+      max_fail_percentage: "{{ edpm_max_fail_percentage | default(0) }}"
+
+This allows for more granular error handling and playbook application by operators.
+
+Error handling
+++++++++++++++
+
+Errors encountered while executing playbooks should be handled in roles which
+caused them.
+
+Playbook test development
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Playbooks themselves don't require tests. However, all of the roles they invoke do.
+For further information about testing roles, please refer to the role contribution
+section of this guide.
+
+
+Documenting a new playbook
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Playbooks must accurately explain their effects in their documentation, with in-line
+comments serving only as a supplement. In complex cases the documentation should link
+to other material, but only as a source of additional information.
+
+
+Primary playbook documentation file must reside at `docs/source/playbooks/<NEWPLAYBOOKNAME>.rst`
+and must be a valid, readable rst document.
+
+This file must contain a title and a `literalinclude` directive of the playbook itself
+at the `../../../playbooks/<NEWPLAYBOOKNAME>.yml` path.
+
+
+.. code-block:: rst
+
+    .. literalinclude:: ../../../playbooks/bootstrap.yml
+       :language: YAML
+
+Optionally, you can write further information about the playbook,
+such as section of examples, in the `docs/source/playbooks/<NEWPLAYBOOKNAME>.rst`.
+However, this documentation must be updated manually, together with the playbook.
+
+Create a new playbook
+~~~~~~~~~~~~~~~~~~~~~
+
+Each new playbook must follow form and basic outline of existing playbooks.
+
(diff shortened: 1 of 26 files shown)


rebtoor (commented):
My +2 on this, i only left a comment for a typo in the doc.

Does it worth to add a `, true` after the default value in each hosts definition to fallback to default even if `edpm_override_hosts` is defined but empty?

jpodivin:
> My +2 on this, i only left a comment for a typo in the doc.
> 
> Does it worth to add a `, true` after the default value in each hosts definition to fallback to default even if `edpm_override_hosts` is defined but empty?

That is a good point. I'll add that right away.

jpodivin:
This has to wait until we merge crun regression mitigation reverts.

jpodivin:
recheck - crun regression workaround removed

docs/source/contributing_playbooks.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Playbook design principles
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Playbooks tie multiple roles and tasks together into cohesive whole.
+If roles represent subroutines of Ansible world, playbooks are programs and scripts.
+Still intended for a certain purpose, but utilizing diverse functionalities of different
+roles to achieve it.
+
+Playbook must duplicate other playbooks.
```
slagle, line 26: must not?

docs/source/contributing_playbooks.rst
```diff
@@ -0,0 +1,92 @@
...
+Where there isn't an explicit requirement, or recommendation,
+provided by this document, the primary `Ansible documentation`_ is considered
+to be the source of truth. When this document and official Ansible docs
+are in substantial conflict, an issue should be raised.
+
+Playbook design principles
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Playbooks tie multiple roles and tasks together into cohesive whole.
+If roles represent subroutines of Ansible world, playbooks are programs and scripts.
```
slagle, line 22: I don't really agree with the analogy. Could we link to ansible docs describing playbooks instead of trying to do it ourselves? It seems redundant. I'd just link to https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_intro.html

docs/source/contributing_playbooks.rst
```diff
@@ -0,0 +1,92 @@
...
+
+
+Documenting a new playbook
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Playbooks must accurately explain their effects in their documentation, with in-line
+comments serving only as a supplement. In complex cases the documentation should link
+to other material, but only as a source of additional information. Beware the link rot.
+
+Playbook license must be compatible with the primary repo license `Apache v2.0`.
```
slagle, line 66: why would we allow any license other than the repo license? I don't feel like this statement is needed. Perhaps general contributing guidance should say something like:

All content contributed to edpm-ansible must be licensed the same as edpm-ansible (`Apache v2.0`).

docs/source/contributing_playbooks.rst
```diff
@@ -0,0 +1,92 @@
...
+For further information about testing roles, please refer to the role contribution
+section of this guide.
+
+
+Documenting a new playbook
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Playbooks must accurately explain their effects in their documentation, with in-line
+comments serving only as a supplement. In complex cases the documentation should link
+to other material, but only as a source of additional information. Beware the link rot.
```
slagle, line 64: I don't think "Beware the link rot" is a useful idiom in our official docs

docs/source/contributing_playbooks.rst
```diff
@@ -0,0 +1,92 @@
...
+       :language: YAML
+
+Optionally, you can write further information about the playbook,
+such as section of examples, in the `docs/source/playbooks/<NEWPLAYBOOKNAME>.rst`.
+However, this documentation must be updated manually, together with the playbook.
+
+Create a new playbook
+~~~~~~~~~~~~~~~~~~~~~
+
+New playbook must follow form and basic outline of existing playbooks.
```
slagle, line 87: `A new` or `Each new`

docs/source/contributing_playbooks.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Where there isn't an explicit requirement, or recommendation,
+provided by this document, the primary `Ansible documentation`_ is considered
+to be the source of truth. When this document and official Ansible docs
+are in substantial conflict, an issue should be raised.
+
+Playbook design principles
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Playbooks tie multiple roles and tasks together into cohesive whole.
```
slagle, line 21: Generally, yes. But for the purposes of edpm-ansible I'd prefer that all tasks are in roles and playbooks only `import_role`. A quick check of all existing playbooks show this to be the case. Could we make this a rule, and only consider exceptions as needed?

slagle (commented):
change looks good, i have several comments on the playbook docs

docs/source/contributing_playbooks.rst
jpodivin: yep.

docs/source/contributing_playbooks.rst
jpodivin: In general we could just say "use apache" but there are compatible licenses and I recall we had some cases in tripleo when we had them rather than then the primary. I suppose it's not really likely to be a deal breaker for anyone so I'll just remove the sentence.

docs/source/contributing_playbooks.rst
jpodivin: I suppose that is a valid approach.

docs/source/contributing_playbooks.rst
```diff
@@ -0,0 +1,88 @@
...
+
+Playbook design principles
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Playbook must not duplicate other playbooks.
+
+If a newly developed playbook requires functionality not already present in existing
+roles, a new role should be submitted in the same PR implementing it.
+
+If possible, playbooks should not use tasks directly, only trough a role.
```
slagle, line 26: through

docs/source/contributing_playbooks.rst
```diff
@@ -0,0 +1,88 @@
...
+
+Playbooks must use following basic layout. Exposing variables, `edpm_override_hosts`,
+`edpm_max_fail_percentage` and `edpm_any_errors_fatal`.
+
+.. code-block:: YAML
+
+    - name: NEWPLAYNAME
+      hosts: "{{ edpm_override_hosts | default('all') }}"
```
slagle, line 34: `default('all', true)`

slagle (approved):
LGTM.
please fix the couple of inline comments in a follow up.

PR #552, opened 2024-02-02
Fix metadata name for testing with an NFS PV

ASBishop:
The previous "edpm-ansible" metadata name was too generic, and other CRs could accidentally claim the PV.


fultonj (commented):
/lgtm

fao89 (approved):

PR #555, opened 2024-02-02
Libvirt qemu group config changes

Jaganathancse:
Missing libvirt qemu group config and ovn bridge
datapath type is not updated correctly.

### DIFF
--- a/roles/edpm_libvirt/templates/qemu.conf.j2
+++ b/roles/edpm_libvirt/templates/qemu.conf.j2
@@ -8,3 +8,6 @@ nbd_tls = 0
 # port usage conflicts. See https://review.openstack.org/#/c/561784
 migration_port_min = 61152
 migration_port_max = 61215
+{% if edpm_nova_libvirt_qemu_group is defined%}
+group = "{{ edpm_nova_libvirt_qemu_group }}"
+{% endif %}
--- a/roles/edpm_ovs_dpdk/tasks/configure.yml
+++ b/roles/edpm_ovs_dpdk/tasks/configure.yml
@@ -196,9 +196,30 @@
       ansible.builtin.set_fact:
         edpm_ovs_other_config: "{{ edpm_ovs_other_config | combine(pmd_auto_lb_rebal_interval) }}"
 
+- name: OVN bridge datapath type config
+  when: edpm_ovn_datapath_type|string
+  block:
+    - name: Apply PMD cores config
+      ansible.builtin.set_fact:
+        ovn_bridge_datapath_type:
+          ovn-bridge-datapath-type: "{{ edpm_ovn_datapath_type }}"
+
+    - name: Append external_ids with ovn datapath type
+      ansible.builtin.set_fact:
+        edpm_ovs_external_ids: "{{ {} | combine(ovn_bridge_datapath_type) }}"
+
 - name: Configure OVS other_config
   ansible.builtin.shell: >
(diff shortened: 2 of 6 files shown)


roles/edpm_libvirt/templates/qemu.conf.j2
```diff
@@ -8,3 +8,6 @@ nbd_tls = 0
 # port usage conflicts. See https://review.openstack.org/#/c/561784
 migration_port_min = 61152
 migration_port_max = 61215
+{% if edpm_nova_libvirt_qemu_group is defined%}
+group = "{{ edpm_nova_libvirt_qemu_group }}"
```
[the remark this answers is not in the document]
Jaganathancse: This group change is specific to DPDK, so setting only for DPDK ansible variable is defined. otherwise no impact on existing config.
Need to configure group value 'hugetlbfs'. otherwise we can't set vhostuser DPDK port.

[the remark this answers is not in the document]
Jaganathancse: i think, group default value is kvm but not used previously and may create issue in other features. 
we can add new  PR later if any default value required.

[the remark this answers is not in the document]
bogdando: @SeanMooney please check if this is the way we are supposed to configure things like that for libvirt, or should we instead provide direct config snippets on osdpd level

roles/edpm_ovs_dpdk/tasks/configure.yml
```diff
@@ -196,9 +196,30 @@
       ansible.builtin.set_fact:
         edpm_ovs_other_config: "{{ edpm_ovs_other_config | combine(pmd_auto_lb_rebal_interval) }}"
 
+- name: OVN bridge datapath type config
+  when: edpm_ovn_datapath_type|string
+  block:
+    - name: Apply PMD cores config
+      ansible.builtin.set_fact:
+        ovn_bridge_datapath_type:
```
bogdando, line 204: using facts is not free for deployment time, could you please instead call combine directly with edpm_ovn_datapath_type ?

roles/edpm_ovs_dpdk/tasks/configure.yml
```diff
@@ -196,9 +196,30 @@
...
 - name: Configure OVS other_config
   ansible.builtin.shell: >
     ovs-vsctl set open . {% for key, value in edpm_ovs_other_config.items() %} other_config:{{ key }}={{ value }} {% endfor %}
   register: ovs_other_config
   changed_when: ovs_other_config.rc == 0
   failed_when: ovs_other_config.rc != 0
+  when: edpm_ovs_other_config | length > 0
+
+- name: Configure OVS external_ids
+  ansible.builtin.shell: >
```
bogdando, line 220: this could be just:
```
vars:
  edpm_ovs_external_ids:
    ovn-bridge-datapath-type: "{{ edpm_ovn_datapath_type }}"
```

roles/edpm_ovs_dpdk/tasks/configure.yml
Jaganathancse: @bogdando This section is created to cover any new OvS DPDK external ids config later similar like ovs dpdkother configs but i have added only one parameter right now.

roles/edpm_ovs_dpdk/tasks/configure.yml
Jaganathancse: we may need to add new OvS DPDK external_ids configs based on conditions.

PR #557, opened 2024-02-06
Global MTU size for the interfaces managed by neutron

Jaganathancse:


roles/edpm_neutron_metadata/templates/neutron-ovn-metadata-agent.conf.j2
```diff
@@ -1,6 +1,10 @@
 [DEFAULT]
 debug = {{ edpm_neutron_metadata_agent_DEFAULT_debug }}
 
+{% if edpm_neutron_global_physnet_mtu is defined %}
+global_physnet_mtu = {{ edpm_neutron_global_physnet_mtu }}
```
karelyatin, line 5: global_physnet_mtu config is for neutron server but not ovn metadata

karelyatin (changes requested):

roles/edpm_neutron_metadata/templates/neutron-ovn-metadata-agent.conf.j2
Jaganathancse: @karelyatin 
Ack, Looks like we can configure default MTU size using openstack controlplane patch similar like sample below
oc patch openstackcontrolplane openstack-galera-network-isolation --type=merge --patch '
spec:
neutron:
template:
      customServiceConfig: |
        [DEFAULT]
        **global_physnet_mtu = 9000**
        [ml2]
        mechanism_drivers = ovn
        **path_mtu = 9000**
        [ovn]
        vhost_sock_dir = /var/lib/vhost_sockets
        [ml2_type_vlan]
        network_vlan_ranges = dpdk1:206:209,dpdk2:206:209

slagle: is this PR still needed in that case?

Jaganathancse: @slagle not required. will close it.

Jaganathancse:
Not required this PR. only controlplane patch config needed.

PR #558, opened 2024-02-07
Add rdoproject repo to get latest libvirt 10 version

arxcruz:
There is a bug on libvirt 9.10 and libvirt 10 which fixes this bug is
not yet available. In order to workaround this situation, the rdoproject
build latest libvirt 10 version on their repo and we are adding it here
in order to fix our long standing block on edpm jobs.
I used shell to configure the repo, because I wasn't able to find in the
ansible documentation how to do so with a .repo file.

### DIFF
--- a/roles/edpm_libvirt/tasks/install.yml
+++ b/roles/edpm_libvirt/tasks/install.yml
@@ -1,4 +1,11 @@
 ---
+# TODO(arxcruz): Remove this once libvirt 10 is available on centos stream
+- name: Add master repository containing libvirt 10
+  shell:
+    cmd: |
+      dnf config-manager --add-repo https://trunk.rdoproject.org/centos9-master/delorean-deps-test.repo && dnf update -y
+  become: true
+
 - name: Install libvirt packages
   tags:
     - install


roles/edpm_libvirt/tasks/install.yml
```diff
@@ -1,4 +1,11 @@
 ---
+# TODO(arxcruz): Remove this once libvirt 10 is available on centos stream
+- name: Add master repository containing libvirt 10
+  shell:
+    cmd: |
+      dnf config-manager --add-repo https://trunk.rdoproject.org/centos9-master/delorean-deps-test.repo && dnf update -y
```
viroel, line 6: I don't think that we should add a fix related to centos here in this repo.
Can we do this in install_yamls or ci-framework instead?

SeanMooney: ya this should not be done in the production code this is a tempeory ci issue.

also I'm not sure that the rdo project is the correct place to pull this form 
if we wanted newer libvirt we shoudl use the virt-preview corp repos

we don't maintain libvirt in rdo.
it is not the upstream of what lands in our downstream product
libvirt comes form rhel directly.
OSP had a libvirt package in osp 13 but not in 16 or 17

SeanMooney (changes requested):
this seams like the wrong repo to work around this in

a better apprch would be to pin this in the ci job or via a package mirror not in the ansibel code

PR #561, opened 2024-02-08
unpin libvirt

SeanMooney:
- unpin the libvirt package

### DIFF
--- a/roles/edpm_libvirt/meta/argument_specs.yml
+++ b/roles/edpm_libvirt/meta/argument_specs.yml
@@ -23,7 +23,7 @@ argument_specs:
       edpm_libvirt_range:
         type: str
         description: The version range for libvirt package.
-        default: "<9.10.0"
+        default: ""
       edpm_libvirt_packages:
         type: list
         description: The list of packages to install for libvirt.
--- a/roles/edpm_libvirt/meta/argument_specs.yml
+++ b/roles/edpm_libvirt/meta/argument_specs.yml
@@ -20,14 +20,18 @@ argument_specs:
           - "virtproxyd"
           - "virtqemud"
           - "virtsecretd"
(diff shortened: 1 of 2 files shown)


slagle:
recheck

roles/edpm_libvirt/meta/argument_specs.yml
```diff
@@ -20,14 +20,18 @@ argument_specs:
...
         description: The list of packages to install for libvirt.
         default:
-          - libvirt
-          - libvirt-admin
-          - libvirt-client
-          - libvirt-daemon
+          - libvirt {{edpm_libvirt_range}}
+          - libvirt-admin {{edpm_libvirt_range}}
+          - libvirt-client {{edpm_libvirt_range}}
+          - libvirt-daemon {{edpm_libvirt_range}}
```
slagle, line 34: the space between name and range looks odd here, but apparently this works with dnf. this worked in my local test:

`
sudo dnf install libvirt\ \<9.7.0-2
`

I guess the ansible module will handle this properly.

slagle (commented):
+2

roles/edpm_libvirt/meta/argument_specs.yml
SeanMooney: the ansible module example had a space when i looked at it which is why i added it here.

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/dnf_module.html#examples


- name: Install Apache >= 2.4
  ansible.builtin.dnf:
    name: httpd >= 2.4
    state: present

the package module support the same syntax.
but it does not have an example so i kept the style form the dnf module


libvirt 10 is now available in the mirrors so we can proceed with unpinning the version

slagle (approved):

PR #562, opened 2024-02-09
[edpm_nova]Support provider*.yaml custom config

gibizer:
The kolla config of the nova-compute container is extended to move the provider*.yaml files to /etc/nova/provider_config/

Implements: [OSPRH-2503](https://issues.redhat.com//browse/OSPRH-2503)


mrkisaolamb (approved):
lgtm

PR #569, opened 2024-02-20
Add support for TLS to edpm_libvirt

SeanMooney:
This change adds molecule testing for deploying libvirt
with tls. to enable this the prepare playbook is enhanced
to generate a ca and issue client and server certs to mimic
what will be generated by cert manager.

The edpm_libvirt role is updated to use the defualt locations
expect by libvirt as documented in https://libvirt.org/kbase/tlscerts.html
This is done to ensure that virsh and other clients automaticaly work
and to avoid overriding the libvirt socket activation files to use custom
values.

Closes: [OSPRH-2387](https://issues.redhat.com//browse/OSPRH-2387)

### DIFF
--- a/roles/edpm_libvirt/molecule/default/prepare.yml
+++ b/roles/edpm_libvirt/molecule/default/prepare.yml
@@ -16,6 +16,8 @@
 
 - name: Setup DUT
   hosts: all
+  vars:
+    edpm_libvirt_tls_cert_src_dir: /tmp/pki
   pre_tasks:
     - name: set basic user fact
       set_fact:
@@ -41,12 +43,16 @@
         popd
         rm -rf repo-setup-main
 
-    # NOTE(gibi): this is done by the boostrap role in a real deployment
-    - name: Install openstack-selinux
+    - name: Install host packages
       become: true
       ansible.builtin.dnf:
-        name: "openstack-selinux"
+        name: "{{ item }}"
         state: present
+      loop:
+        # NOTE(gibi): this is done by the boostrap role in a real deployment
+        - openstack-selinux
+        # NOTE(sean-k-mooney): this is needed to generate certs for tls testing
+        - python3-cryptography
 
     - name: set /etc/localtime
       become: true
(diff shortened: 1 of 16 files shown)


SeanMooney:
i need to enable/mask some of the socket activation files before this will actually be funcitonal

roles/edpm_libvirt/tasks/configure.yml
```diff
@@ -115,14 +115,16 @@
     - libvirt
   become: true
   loop:
-    - {"src": "/var/lib/openstack/certs/libvirt/tls.crt", "dest": "/etc/pki/libvirt/tls.crt"}
-    - {"src": "/var/lib/openstack/certs/libvirt/tls.key", "dest": "/etc/pki/libvirt/private/tls.key"}
-    - {"src": "/var/lib/openstack/cacerts/libvirt/tls-ca-bundle.pem", "dest": "/etc/pki/CA/libvirt/tls-ca-bundle.pem"}
+    - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.crt", "dest": "/etc/pki/libvirt/servercert.pem"}
+    - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.key", "dest": "/etc/pki/libvirt/private/serverkey.pem"}
+    - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.crt", "dest": "/etc/pki/libvirt/clientcert.pem"}
+    - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.key", "dest": "/etc/pki/libvirt/private/clientkey.pem", "mode": "0644"}
```
olliewalsh, line 121: 0644 doesn't look right for the key, should that have been set on the cacert?

olliewalsh: If the intention is to allow access to non-root users (is that wise?) then 0640 and a specific group perhaps

SeanMooney: to the client key yes. because no one shoudl be logging in to the compute ndoes as root and i want ti work for the non admin user by default
with that said i could make it 0640 and change the ownership to root:libvirt

that woudl mean only members of the libvirt group could use it.

the virt-pki-validate script is actully what proped me to change it 

https://github.com/libvirt/libvirt/blob/11839c8cc1d204ec3abdb3a0d9499b64dc83fc83/tools/virt-pki-validate.in#L230

 if [ "$MOD" != "-rw-r--r--" ]
            then
                echo The client private key need to be read by client tools
                echo "as root do: chmod 644 $LIBVIRTP/clientkey.pem"
            fi

so they recommend using 644 but that because they do not have a libvirt group by default.
we do so we can make it tigher if we want too.

SeanMooney: i guess it would work with sudo without this and the user you will conenct in as to debug should have that so i can likely revert to 600 ill need to test that but thats not unreasonabel to do.

roles/edpm_libvirt/molecule/default/verify.yml
```diff
@@ -20,8 +20,23 @@
...
+      register: virt_pki_validate
+    - name: Assert that virt-pki-validate returns no errors
+      assert:
+        that:
+          - "virt_pki_validate.rc == 0"
+
+    - name: ensure we can connect to libvirt with virsh via tls
+      # Note we need become because the client cert is owned by root
+      become: true
+      ansible.builtin.command: "virsh -c qemu+tls://compute-1/system version"
```
SeanMooney, line 36: this proably needs to be 

ansible.builtin.command: "virsh -c qemu+tls://{{ inventory_hostname }}/system version"

but ill update that if/when it fails in ci

roles/edpm_libvirt/molecule/default/prepare.yml
```diff
@@ -125,3 +131,61 @@
...
+
+    - name: Create a certificate signing request for ca
+      community.crypto.openssl_csr_pipe:
+        privatekey_path: "{{ edpm_libvirt_tls_cert_src_dir }}/ca.key"
+        organization_name: fake-corp
+        basic_constraints:
+          - 'CA:TRUE'
+        subject_alt_name:
+          - "DNS:compute-1.example.com"
+          - "DNS:compute-1"
```
SeanMooney, line 156: if i have to update the hostname in the virsh line ill need to update it here too to inventory_hostname

compute-1 is the host name used in the vagrant molecule env but in our ci job we are suing the deleaged plugin and the hostname wont be static for the zuul vm

SeanMooney:
the molecule test is now passing so this is good to be reviewed https://review.rdoproject.org/zuul/build/d51d20e51cd9494cb813a1724cb15e9b

gibizer (approved):
looks good to me

vakwetu:
looks good to me

roles/edpm_libvirt/tasks/install.yml
```diff
@@ -21,6 +21,11 @@
     enabled: false
     masked: true
     state: stopped
+  loop:
```
olliewalsh, line 24: line 20 should be `name: "{{ item }}"`

olliewalsh (changes requested):

roles/edpm_libvirt/tasks/install.yml
SeanMooney: ... yes

so they are disabeld and stopped by default but not masked

which means i also am missing a verify test fo the sockets.

ill respin this tomorrow thanks

gibizer (approved):
looks good to me

PR #572, opened 2024-02-27
Add script to derive passthrough device_spec for nova-sriov

vcandapp:
**Compute Derive list of allowed pci devices for nova-compute from user requested pci_addresses/devname :**

- SRIOV deployment will define ConfigMap for nova compute service with all NICs enabled for SRIOV. (Note: This input will be a dict of nics similar to supported osp releases)
- The input will contain list of SRIOV nics, to be written in xx-sriov-nova.conf file
- Additional flag will be used to specify if the device_spec has to checked for nic partition or not (same as 17)
- If true, copy the service python script to compute node
- Sriov_config.map from os-net will contain VFs used by NIC partitioning
- Execute and derive_allowed_devspec module ( read from xx-sriov-nova.conf file and update the config map)
- Restart nova compute service (if pci dev_spec is to be modified) 

Note:  xx-sriov-nova.conf will be updated to proper config map name in VA

Status: After discussing with all stakeholders on the SR-IOV NIC-partitioing deployment, it came out that ease-of-use is not a requirement for NIC paritioning. Also, modifying nova configmap (which this module updates, for single-step deployment) is not permitted
**Decision taken:**
Even though we have reservations from customers point-of-view on this approach, we will created a new role edpm_derive_pci_device_spec for the pci device_spec derivation. 
Follow the two step-deployment approach (1. Do a provision upto a point where this role generates the configmap 2. SSH into compute to get the exact sriov device_spec configmap and continue  by using this map for nova, complete the deployment for remaining services)

### DIFF
--- a/plugins/modules/edpm_nova_derive_pci_passthrough_devicespec.py
+++ b/plugins/modules/edpm_nova_derive_pci_passthrough_devicespec.py
@@ -1,0 +1,649 @@
+#!/usr/bin/python3
+
+#
+# Copyright 2024 Red Hat Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+__metaclass__ = type
+
+import json
+import os
+import re
+import yaml
+
+
+from ansible.module_utils.basic import AnsibleModule
+
+ANSIBLE_METADATA = {
+    'metadata_version': '1.0',
+    'status': ['preview'],
+    'supported_by': 'community'
+}
+
+DOCUMENTATION = '''
+---
+module: edpm_nova_derive_pci_passthrough_devicespec
+author:
+    - OpenStack EDPM Contributors
+version_added: '1.0'
+short_description: Derive pci passthrough devspec for nova for sriov devices
+notes: []
+description:
+  - This module updates the pci device_spec addr list for nova,
+    when sriov nic partitioning is configured in EDPM nodes.
+    See openstack user guide for details of
+    nova PCI passthrough formats.
+options:
+  sriov_nova_conf_file:
+    description:
+      - User specified nova conf file for nic device lookup
+    type: str
+    required: true
+    default: '/var/lib/openstack/config/nova/04-sriov-nova.conf'
+  sriov_phydev_map:
+    description:
+      -  SR-IOV <-> phydev mapping str
+    type: str
+    required: false
+    default: ''
+'''
+
+EXAMPLES = '''
+- name: Create pci device_spec from defaults
+  edpm_nova_derive_pci_passthrough_devicespec:
+    - 04-sriov-nova.conf
+    - 'sriov1:ens2f0np0,sriov2:ens2f1np1'
+'''
+
+RETURN = '''
+rc:
+  description:
+    - Integer for the return code
+  returned: always
+  type: int
+'''
+
+_PASSTHROUGH_DEVSPEC_KEY = 'device_spec'
+_PCI_DEVICES_PATH = '/sys/bus/pci/devices'
+_SYS_CLASS_NET_PATH = '/sys/class/net'
+
+MAX_FUNC = 0x7
+MAX_DOMAIN = 0xFFFF
+MAX_BUS = 0xFF
+MAX_SLOT = 0x1F
+ANY = '*'
+REGEX_ANY = '.*'
+
+ADD_PF_PCI_ADDRESS = 0x1
+ADD_VF_PCI_ADDRESS = 0x2
+DEL_USER_CONFIG = 0x3
+KEEP_USER_CONFIG = 0x4
+
+
+class InvalidConfigException(ValueError):
+    pass
+
+
+def get_pci_field_val(
+        prop: str, maxval: int, hex_value: str):
+    if prop == ANY:
+        return REGEX_ANY
+    try:
+        v = int(prop, 16)
+    except ValueError:
+        raise InvalidConfigException('Invalid PCI address specififed {!r}'.format(prop))
+    if v > maxval:
+        raise InvalidConfigException('PCI address specififed {!r} is out of range'.format(prop))
+    return hex_value % v
+
+
+def get_pciaddr_dict_from_usraddr(pci_addr: str):
+    """Convert PCI address in STRING format to DICT
+    (this is done for uniformity in PCI address)
+    """
+    pci_dict = {}
+    dbs, sep, func = pci_addr.partition('.')
+    pci_dict['function'] = ANY
+    if func:
+        func = func.strip()
+        pci_dict['function'] = func
+    if dbs:
+        dbs_fields = dbs.split(':')
+        if len(dbs_fields) > 3:
+            raise InvalidConfigException('Invalid PCI address specififed {!r}'.format(pci_addr))
+        # If we got a partial address like ":00.", we need to turn this
+        # into a domain of ANY, a bus of ANY, and a slot of 00. This code
+        # allows the address,bus and/or domain to be left off
+        dbs_all = [ANY] * (3 - len(dbs_fields))
+        dbs_all.extend(dbs_fields)
+        dbs_checked = [s.strip() or ANY for s in dbs_all]
+
+        ''' domain, bus, slot = dbs_checked '''
+        pci_dict['domain'], pci_dict['bus'], pci_dict['slot'] = dbs_checked
+
+    pci_dict['domain'] = get_pci_field_val(pci_dict['domain'], MAX_DOMAIN, '%04x')
+    pci_dict['slot'] = get_pci_field_val(pci_dict['slot'], MAX_SLOT, '%02x')
+    pci_dict['bus'] = get_pci_field_val(pci_dict['bus'], MAX_BUS, '%02x')
+    pci_dict['function'] = get_pci_field_val(pci_dict['function'], MAX_FUNC, '%1x')
+    return pci_dict
+
+
+def get_pci_regex_pattern(config_regex: str, size: int, maxval: int, hex_value: str):
+    if config_regex in [ANY, REGEX_ANY]:
+        config_regex = "[0-9a-fA-F]{%d}" % size
+
+    try:
+        re.compile(config_regex)
+    except re.error:
+        msg = "Invalid regex pattern identified %s" % config_regex
+        raise InvalidConfigException(msg)
+    try:
+        v = int(config_regex, 16)
+    except ValueError:
+        return config_regex
+    if v > maxval:
+        msg = "Invalid pci address"
+        raise InvalidConfigException(msg)
+    return hex_value % v
+
+
+def get_user_regex_from_addrdict(addr_dict):
+    if isinstance(addr_dict, dict):
+        domain_regex = get_pci_regex_pattern(addr_dict['domain'], 4, MAX_DOMAIN, '%04x')
+        bus_regex = get_pci_regex_pattern(addr_dict['bus'], 2, MAX_BUS, '%02x')
+        slot_regex = get_pci_regex_pattern(addr_dict['slot'], 2, MAX_SLOT, '%02x')
+        function_regex = get_pci_regex_pattern(addr_dict['function'], 1, MAX_FUNC, '%1x')
+
+        user_address_regex = '%s:%s:%s.%s' % (domain_regex,
+                                              bus_regex,
+                                              slot_regex,
+                                              function_regex)
+        return user_address_regex
+    else:
+        return None
+
+
+def get_pci_addr_from_ifname(ifname: str):
+    """Given the device name, returns the PCI address of a device
+    and returns True if the address is in a physical function.
+    """
+    dev_path = os.path.join(_SYS_CLASS_NET_PATH, ifname, "device")
+    if os.path.isdir(dev_path):
+        try:
+            return (os.readlink(dev_path).strip("./"))
+        except OSError:
+            raise
+    return None
+
+
+def get_sriov_configs():
+    configs = []
+    try:
+        with open('/var/lib/os-net-config/sriov_config.yaml') as sriov_config:
+            configs = yaml.safe_load(sriov_config)
+    except IOError:
+        configs = []
+    return configs
+
+
+def get_sriov_nic_partition_pfs(configs):
+    pfs = []
+    for config in configs:
+        device_type = config.get('device_type', None)
+        if device_type and 'vf' in device_type:
+            device = config.get('device', None)
+            if device:
+                dev_name = device.get('name', None)
+                if dev_name and dev_name not in pfs:
+                    pfs.append(dev_name)
+    return pfs
+
+
+def get_sriov_non_nic_partition_pfs(configs):
+    all_pfs = []
+    non_nicp_pfs = []
+    nicp_pfs = get_sriov_nic_partition_pfs(configs)
+
+    for config in configs:
+        device_type = config.get('device_type', None)
+        if device_type and 'pf' in device_type:
+            name = config.get('name', None)
+            if name and name not in all_pfs:
+                all_pfs.append(name)
+    non_nicp_pfs = [x for x in all_pfs if x not in nicp_pfs]
+    return non_nicp_pfs
+
+
+def get_pci_device_info_by_ifname(pci_dir, sub_dir):
+    if not os.path.isdir(os.path.join(pci_dir, sub_dir)):
+        return None
+    try:
+        # ids located in files inside PCI devices
+        # directory are stored in hex format (0x1234 for example)
+        with open(os.path.join(pci_dir, sub_dir,
+                               'vendor')) as vendor_file:
+            vendor = vendor_file.read().strip()
+        with open(os.path.join(pci_dir, sub_dir,
+                               'device')) as product_file:
+            product = product_file.read().strip()
+        return vendor, product
+    except IOError:
+        return None
+
+
+def get_available_vf_pci_addresses_by_ifname(pf_name, allocated_pci):
+    """It gets the list of all VF's of a PF minus the VFs allocated for
+    NIC Partitioning. If the PF has 10 VFs and VFs 0,1 are allocated then
+    the PCI address of VFs from 2 to 9 along with the device info wout be returned
+    """
+    vf_pci_addresses = {}
+    pci_dir = _PCI_DEVICES_PATH
+    if os.path.isdir(pci_dir):
+        for sub_dir in os.listdir(pci_dir):
+            if sub_dir in allocated_pci:
+                continue
+            pci_phyfn_dir = os.path.join(pci_dir, sub_dir, 'physfn/net')
+            if os.path.isdir(pci_phyfn_dir):
+                phyfn_dirs = os.listdir(pci_phyfn_dir)
+                for phyfn in phyfn_dirs:
+                    if phyfn in pf_name:
+                        if phyfn not in vf_pci_addresses:
+                            vf_pci_addresses[phyfn] = [sub_dir]
+                        else:
+                            vf_pci_addresses[phyfn].append(sub_dir)
+    return vf_pci_addresses
+
+
+def get_allocated_pci_addresses(configs):
+    alloc_pci_info = []
+    for config in configs:
+        pci = config.get('pci_address', None)
+        if pci:
+            alloc_pci_info.append(pci)
+    return alloc_pci_info
+
+
+def get_pci_passthrough_devicespec(user_config, pf, pci_addresses):
+    pci_passthrough_list = []
+
+    for pci in pci_addresses:
+        pci_passthrough = dict(user_config)
+        pci_passthrough['address'] = str(pci)
+
+        # devname and address fields can't co exist
+        if 'devname' in pci_passthrough:
+            del pci_passthrough['devname']
+
+        pci_passthrough_list.append(pci_passthrough)
+    return pci_passthrough_list
+
+
+def match_pf_details(user_config, pf_name, is_non_nic_pf: bool):
+    """Decide the action for devicespec_pci_addr, based on user config
+
+    :param user_configs: THT param NovaPCIPassthrough
+    :param pf_name: Interface/device name (str)
+    :param is_non_nic_pf: Indicates whether the PF is noc-partitioned or not
+    :return: Return the actions to be done, based on match criteria
+    """
+
+    # get the vendor and product id of the PF and VF
+    pf_path = os.path.join(_SYS_CLASS_NET_PATH, pf_name, "device")
+    vendor, vf_product = get_pci_device_info_by_ifname(pf_path, 'virtfn0')
+    vendor, pf_product = get_pci_device_info_by_ifname(pf_path, '')
+
+    if ('product_id' not in user_config
+            or vf_product[-4:] == user_config['product_id'][-4:]):
+        if is_non_nic_pf:
+            # If NON NIC Part PF matches, then add the complete device
+            return ADD_PF_PCI_ADDRESS
+        else:
+            """ In case of NIC Partitioning PF, add the VFs only (excluding NIC Part VFs)
+                when the product id is not given or product_id matches that of VF
+            """
+            return ADD_VF_PCI_ADDRESS
+    elif ('product_id' not in user_config
+          or pf_product[-4:] == user_config['product_id'][-4:]):
+        if is_non_nic_pf:
+            # If product id of NON NIC Part VF matches, then add the complete device
+            return ADD_PF_PCI_ADDRESS
+        else:
+            """ When the user_config address matches that of the NIC Partitioned PF,
+                the PF must be removed from the user_config. So return the status such
+                that the caller ignores this user_config if this user_config is very
+                specific to the NIC Partitioned PF
+            """
+            return DEL_USER_CONFIG
+    else:
+        """ The Product ID neither belongs to VF nor PF, simply ignore matching
+        """
+        return KEEP_USER_CONFIG
+
+
+# +------------+-----------------------+---------------------------+-------------------+
+# |   USER     |     product_id        |       product_id          |  Product ID       |
+# |   CONFIG   |     is VF             |       is PF               |  NOT mentioned    |
+# +------------+-----------------------+---------------------------+-------------------+
+# |   PCI addr |                       |                           |                   |
+# |   is VF    |     Matching VF       |       INVALID             |  add only VFs     |
+# |            |                       |                           |                   |
+# |            |                       |                           |                   |
+# +------------+-----------------------+---------------------------+-------------------+
+# |            |                       |                           |                   |
+# |   PCI addr |     ALL VFs of this   |       Matching PFs        | Add both PF and   |
+# |   is PF    |     addr -            |      - NIC Part PFs       | available VFs     |
+# |            |     NIC Part VFs      |                           |                   |
+# |            |                       |                           |                   |
+# +------------+-----------------------+---------------------------+-------------------+
+# |            |                       |                           |                   |
+# |  PCI addr  |     All matching      |       All matching        |  INVALID CASE     |
+# |     not    |     VFs               |       PFs - NIC Part PF   |                   |
+# |  specified |     NIC Part VFs      |                           |                   |
+# |            |                       |                           |                   |
+# +------------+-----------------------+---------------------------+-------------------+
+def get_passthrough_config(user_config, pf_name,
+                           allocated_pci, is_non_nic_pf: bool):
+    """Handle all variations of user specifid pci addr format
+
+    Arrive at the address fields of the devicespec. Check the address fields of
+    the pci.passthrough_devicespec configuration option, validating the address fields.
+
+    :param user_configs: THT param NovaPCIPassthrough
+    :param pf_name: Interface/device name (str)
+    :param allocated_pci: List of VFs (for nic-partitioned PF), which are used by host
+    :param is_non_nic_pf: Indicates whether the PF is non-partitioned or not
+    :return: pci_passthrough: Derived config list
+    :return del_user_config: Flag to state if the user_config is to be deleted or not
+
+    Example format for user_config:
+    | [pci] in standard string format
+    | passthrough_devicespec = {"address":"*:0a:00.*",
+                          "physical_network":"physnet1"}
+    | passthrough_devicespec = {"address": {"domain": ".*",
+                                       "bus": "02",
+                                       "slot": "01",
+                                       "function": "[0-2]"},
+                            "physical_network":"net1"}
+    """
+    sel_addr = []
+    pci_passthrough = []
+    del_user_config = False
+
+    """ Get the regex of the address fields """
+    if 'address' in user_config:
+        if isinstance(user_config['address'], dict):
+            addr_dict = user_config['address']
+        elif isinstance(user_config['address'], str):
+            addr_dict = get_pciaddr_dict_from_usraddr(user_config['address'])
+        user_address_pattern = get_user_regex_from_addrdict(addr_dict)
+    else:
+        user_address_pattern = ("[0-9a-fA-F]{4}:[0-9a-fA-F]{2}:"
+                                "[0-9a-fA-F]{2}.[0-7]")
+
+    """ If address mentioned in user_config is PF, then get all VFs belonging to that PF
+    """
+    available_vfs = get_available_vf_pci_addresses_by_ifname(pf_name, allocated_pci)
+
+    # get the pci address of the PF
+    parent_pci_address = get_pci_addr_from_ifname(pf_name)
+    user_address_regex = re.compile(user_address_pattern)
+
+    # Match the user_config address with the PF's address
+    if user_address_regex.match(parent_pci_address):
+        match_id = match_pf_details(user_config, pf_name, is_non_nic_pf)
+        """ if the address matches and there is no mismatch in
+        product id's of PF, then add the PF's provided its not a
+        NIC Partitioning PF
+        """
+        if match_id == ADD_PF_PCI_ADDRESS:
+            sel_addr.append(parent_pci_address)
+        elif match_id == ADD_VF_PCI_ADDRESS:
+            if pf_name in available_vfs:
+                for vf_addr in available_vfs[pf_name]:
+                    sel_addr.append(vf_addr)
+            if not sel_addr:
+                del_user_config = True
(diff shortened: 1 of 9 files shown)


roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,585 @@
...
+
+import json
+import os
+import re
+import yaml
+
+_PASSTHROUGH_WHITELIST_KEY = 'device_spec'
+_PCI_DEVICES_PATH = '/sys/bus/pci/devices'
+_SYS_CLASS_NET_PATH = '/sys/class/net'
+_DERIVED_PCI_WHITELIST_FILE = '/var/lib/openstack/pci_passthrough_whitelist.json'
```
Jaganathancse, line 26: @vcandapp Need to update conf file format and path.

roles/edpm_nova/defaults/main.yml
```diff
@@ -31,3 +31,8 @@ edpm_nova_compute_image: "quay.io/podified-antelope-centos9/openstack-nova-compu
 
 # certs
 edpm_nova_tls_certs_enabled: "{{ edpm_tls_certs_enabled | default(False) }}"
+
+#Defaults for  PCI whitelist derive script, if nic-partitioning is enabled
+edpm_nova_sriov_derive_pci_whitelist: true
+#TBD: This can be an ENV variable
+edpm_nova_default_sriov_conf_file: 03-sriov-nova.conf
```
Jaganathancse, line 38: @vcandapp is thi required? can we get the file name list from configmaps flattened for nova role?

roles/edpm_nova/tasks/derive_pci.yml
```diff
@@ -0,0 +1,74 @@
...
+# under the License.
+
+- name: EDPM Node derive_novapci_whitelist Config
+  hosts: "{{ edpm_override_hosts | default('all', true) }}"
+  strategy: linear
+  become: true
+  tags:
+    - nova
+  tasks:
+    - name: Check if edpm node has the sriov nova conf
```
Jaganathancse, line 24: @vcandapp instead of hard coding filename for source filename, can we consider all the user input configmap files to identify soucrce file using pci section key?

roles/edpm_nova/tasks/derive_pci.yml
```diff
@@ -0,0 +1,74 @@
...
+              "{{ edpm_nova_config_dest }}/{{ edpm_nova_default_sriov_conf_file }}"
+
+    - name: Derive the sriov pci whitelist addr for nic-partition
+      when:
+        - sriov_nova_conf_check.stat.exists
+        - edpm_nova_sriov_derive_pci_whitelist|bool 
+      - name: Create folder for the derive_whitelist script
+        file:
+          state: directory
+          path: /var/lib/openstack/nova_pci_whitelist_script
```
Jaganathancse, line 42: @vcandapp can we use module instead of using script file?

roles/edpm_nova/defaults/main.yml
```diff
@@ -31,3 +31,8 @@ edpm_nova_compute_image: "quay.io/podified-antelope-centos9/openstack-nova-compu
 
 # certs
 edpm_nova_tls_certs_enabled: "{{ edpm_tls_certs_enabled | default(False) }}"
+
+#Defaults for  PCI whitelist derive script, if nic-partitioning is enabled
+edpm_nova_sriov_derive_pci_whitelist: true
```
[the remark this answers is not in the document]
vcandapp: For ease-of-use, in osp17 it was set to true (for all sriov role, irrespective of nic-partition is used ot not)

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
vcandapp: @Jaganathancse, this file is retained temporarily (so that users can run the derive script and refer the whitelist like osp17). Intend to remove it later

roles/edpm_nova/defaults/main.yml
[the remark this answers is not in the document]
Jaganathancse: @vcandapp  This is part of neutron-sriov-agent in osp17, so setting default will not be problem and used for only SRIOV or NIC Partitioning nodes.
Now derive logic based on edpm-nova role and may common for multiple features. so we need to make this variable as false default. we can enable only for SRIOV NIC partition nodeset.

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,587 @@
...
+        ''' Check the data type of first json decode '''
+        if not isinstance(phys_dev_mappings, list):
+            msg = f'ml2::sriov::physical_device_mappings specified is not a list {phys_dev_mappings}'
+            raise InvalidConfigException(msg)
+
+        for phy_dev_map in phys_dev_mappings:
+            net_name, nic_name = phy_dev_map.split(':')
+            if net_name == physical_network:
+                return nic_name
+        return None
```
Jaganathancse, line 472: @vcandapp Need to change this logic based on edpm_neutron_sriov_agent_SRIOV_NIC_physical_device_mappings edpm-ansible variable.
you can default nova variable with getting value from above variable and use added variable for this logic.

roles/edpm_nova/tasks/derive_pci.yml
```diff
@@ -0,0 +1,72 @@
...
+          ansible.builtin.file:
+            state: directory
+            path: /var/lib/openstack/nova_pci_whitelist_script
+            owner: root
+            group: root
+            mode: '0750'
+        - name: Copy the Derive pci passthrough whitelist script
+          ansible.builtin.copy:
+            # TBD: This destination path can be updates
+            src: /usr/share/roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
```
Jaganathancse, line 48: @vcandapp source path is incorrect.

Jaganathancse: destination file will not be there.

roles/edpm_nova/tasks/configure.yml
```diff
@@ -174,3 +174,7 @@
     mode: "0400"
     attributes: "+i"
   when: compute_id.stat.exists
+
+# Nova should be restarted, if derive script modifies nova conf
```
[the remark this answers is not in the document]
vcandapp: Fixed

roles/edpm_nova/tasks/derive_pci.yml
vcandapp: For feature parity, will leave this as script for now. There seems to be no need for this script to be run as module outside of nic-partition

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,587 @@
...
+                    else:
+                        configs.append(obj[1])
+    except IOError:
+        configs = []
+    return (configs, src_config)
+
+
+if __name__ == "__main__":
+    pci_passthro = {}
+    sriov_nova_conf_file = '/var/lib/openstack/config/nova/03-sriov-nova.conf'
```
Jaganathancse, line 555: @vcandapp Can we change the file name used similar like VA architecture.

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,587 @@
+#!/usr/bin/python3
+
+#
+# Copyright 2019 Red Hat Inc.
```
jpodivin, line 4: We can keep this here, if this is an unmodified copy of an existing script. Otherwise it must be updated to 2024

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,587 @@
...
+        device_type = config.get('device_type', None)
+        if device_type and 'pf' in device_type:
+            name = config.get('name', None)
+            if name and name not in all_pfs:
+                all_pfs.append(name)
+    non_nicp_pfs = [x for x in all_pfs if x not in nicp_pfs]
+    return non_nicp_pfs
+
+
+def get_pci_device_info_by_ifname(pci_dir, sub_dir):
```
jpodivin, line 175: Function can return `None` or a tuple. All invocations of this function must perform a check, or the function must be rewritten so it's not possible.

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,587 @@
...
+
+        user_address_regex = '%s:%s:%s.%s' % (domain_regex,
+                                              bus_regex, slot_regex,
+                                              function_regex)
+        return user_address_regex
+    else:
+        return None
+
+
+def get_pci_addr_from_ifname(ifname: str):
```
jpodivin, line 124: Function can return `None` or a path. Consuming functions need a check, or a rewrite.

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,587 @@
...
+        bus_regex = get_pci_regex_pattern(addr_dict['bus'], 2, MAX_BUS, '%02x')
+        slot_regex = get_pci_regex_pattern(addr_dict['slot'], 2, MAX_SLOT, '%02x')
+        function_regex = get_pci_regex_pattern(addr_dict['function'], 1, MAX_FUNC, '%1x')
+
+        user_address_regex = '%s:%s:%s.%s' % (domain_regex,
+                                              bus_regex, slot_regex,
+                                              function_regex)
+        return user_address_regex
+    else:
+        return None
```
jpodivin, line 121: Function can return `None` or a string. Consuming functions need a check, or a rewrite.

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,587 @@
...
+
+
+def get_user_regex_from_addrdict(addr_dict):
+    if isinstance(addr_dict, dict):
+        domain_regex = get_pci_regex_pattern(addr_dict['domain'], 4, MAX_DOMAIN, '%04x')
+        bus_regex = get_pci_regex_pattern(addr_dict['bus'], 2, MAX_BUS, '%02x')
+        slot_regex = get_pci_regex_pattern(addr_dict['slot'], 2, MAX_SLOT, '%02x')
+        function_regex = get_pci_regex_pattern(addr_dict['function'], 1, MAX_FUNC, '%1x')
+
+        user_address_regex = '%s:%s:%s.%s' % (domain_regex,
```
jpodivin, line 116: Switch to fstring would be warranted.

jpodivin (changes requested):

roles/edpm_nova/tasks/derive_pci.yml
vcandapp: Updated

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
vcandapp: Updated default.yml with sriov conf file name as user configurable input

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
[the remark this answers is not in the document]
vcandapp: Updated year

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,587 @@
...
+
+
+# Todo: (remove)This function will not be needed now
+def get_neutron_devmap_configs():
+    ml2_sriov = []
+    valid_spec = 'physical_device_mappings'
+
+    # Read phydev mapping from neutron conf file
+    try:
+        with open('/var/lib/os-net-config/01-neutron-sriov-agent.conf') as neutron_config:
```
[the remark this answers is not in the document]
vcandapp: Since the final config map files for sriov-nova and sriov-neutron (for NFV) are not yet finalised in VA design, using the path which QE uses (i.e ci-framework file name)
This will be updated later once VA is updated (but updated plugin to take this file as user input)

roles/edpm_nova/defaults/main.yml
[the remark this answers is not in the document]
vcandapp: The wallaby release had this param set to true:
https://github.com/openstack-archive/tripleo-heat-templates/blob/stable/wallaby/deployment/neutron/neutron-sriov-agent-container-puppet.yaml#L75

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
vcandapp: This will be take as user input

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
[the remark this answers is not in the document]
vcandapp: As per latest code:
https://github.com/openstack-k8s-operators/architecture/blob/c85878642bcd64eb67ad2ea19e4e6e8bdb790cfb/va/nfv/sriov/edpm/nova_sriov.yaml#L14

roles/edpm_nova/defaults/main.yml
vcandapp: Will take this feature as part of feature pack 1

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
vcandapp: It is handled: https://github.com/openstack-archive/tripleo-heat-templates/blob/stable/wallaby/deployment/neutron/derive_pci_passthrough_whitelist.py#L370

roles/edpm_nova/tasks/derive_pci.yml
```diff
@@ -0,0 +1,48 @@
...
+        owner: root
+        group: root
+        mode: '0750'
+    - name: Copy the Derive pci passthrough whitelist script
+      ansible.builtin.copy:
+        src: "{{ role_path }}/scripts/derive_pci_passthrough_whitelist.py"
+        dest: "{{ edpm_nova_dst_pciwl_dir }}/{{ edpm_nova_dst_pciwl_file }}"
+        mode: '0700'
+    - name: Run derive_pci_passthrough_whitelist.py
+      ansible.builtin.command: "{{ edpm_nova_dst_pciwl_dir }}/{{ edpm_nova_dst_pciwl_file }}"
```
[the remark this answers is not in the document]
vcandapp: Done

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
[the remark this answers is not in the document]
vcandapp: Done

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,587 @@
...
+            for line in neutron_config:
+                line = line.replace('\n', '')
+                if line.startswith(valid_spec):
+                    phy_map = re.split(', |,|=|= ', line)
+                    for i in phy_map:
+                        if ':' in i:
+                            ml2_sriov.append(i)
+    except IOError:
+        ml2_sriov = []
+    return ml2_sriov
```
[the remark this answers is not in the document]
vcandapp: Updated comment

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,587 @@
...
+
+_PASSTHROUGH_WHITELIST_KEY = 'device_spec'
+_PCI_DEVICES_PATH = '/sys/bus/pci/devices'
+_SYS_CLASS_NET_PATH = '/sys/class/net'
+_DERIVED_PCI_WHITELIST_FILE = '/var/lib/openstack/pci_passthrough_whitelist.json'
+
+MAX_FUNC = 0x7
+MAX_DOMAIN = 0xFFFF
+MAX_BUS = 0xFF
+MAX_SLOT = 0x1F
```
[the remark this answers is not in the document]
vcandapp: These are as per standards: https://docs.openstack.org/nova/latest/admin/pci-passthrough.html

roles/edpm_nova/tasks/derive_pci.yml
vcandapp: Done

Jaganathancse:
Looking good and please remove copying derive pci scripts and using neutron sriov conf file instead of using user input neutron ansible variable.

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
vcandapp: The user input of the derive script is valid pf/vf

roles/edpm_nova/scripts/derive_pci_passthrough_whitelist.py
vcandapp: The user input of the derive script is valid pf/vf addr

vcandapp:
> Looking good and please remove copying derive pci scripts and using neutron sriov conf file instead of using user input neutron ansible variable.

Ack

vcandapp:
recheck

SeanMooney:
/hold this is expcitly somethign compute has said we do not want to support.
please talks to use and don't proceed with this until we are in agreement if we should support this or not.
right now this is explicitly not in scope of edpm ansible based on past PM feedback so this should not be added.

plugins/modules/edpm_nova_derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,671 @@
...
+    - 03-sriov-nova.conf
+'''
+
+RETURN = '''
+'''
+
+_PASSTHROUGH_WHITELIST_KEY = 'device_spec'
+_PCI_DEVICES_PATH = '/sys/bus/pci/devices'
+_SYS_CLASS_NET_PATH = '/sys/class/net'
+_DERIVED_PCI_WHITELIST_FILE = '/var/lib/openstack/pci_passthrough_whitelist.json'
```
SeanMooney, line 79: again this si not an accpaable name but also I'm not sure this is a good location to use.

if this file was owned by the nova roles then it should be in /var/lib/openstack/config/nova

plugins/modules/edpm_nova_derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,671 @@
...
+    the pci.passthrough_whitelist configuration option, validating the address fields.
+
+    :param user_configs: THT param NovaPCIPassthrough
+    :param pf_name: Interface/device name (str)
+    :param allocated_pci: List of VFs (for nic-partitioned PF), which are used by host
+    :param is_non_nic_pf: Indicates whether the PF is non-partitioned or not
+    :return: pci_passthrough: Derived config list
+    :return del_user_config: Flag to state if the user_config is to be deleted or not
+
+    Example format for user_config:
```
SeanMooney, line 366: https://github.com/openstack/nova/blob/master/nova/pci/devspec.py#L203-L211

     | [pci]
        | device_spec = {"address":"*:0a:00.*",
        |                "physical_network":"physnet1"}
        | device_spec = {"address": {"domain": ".*",
                                     "bus": "02",
                                     "slot": "01",
                                     "function": "[0-2]"},
                         "physical_network":"net1"}
        | device_spec = {"vendor_id":"1137","product_id":"0071"}

SeanMooney (changes requested):
as writen this should not merge as this would regress our removal of the term whitelist form our product and its not approatte for the edpm_nova role to modify the sriov agent config.

on higher level we are try not to generate config on the edpm nodes via ansible and when we raised this question with gil in teh past they said we should not spend time automating this for GA.


this is not required for the validated architectures

https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/nfv/sriov/edpm/values.yaml#L166-L169

and we already have the capability to define this config via configmap associated with the custom nova datapalen service.

Jaganathancse:
> > /hold this is expcitly somethign compute has said we do not want to support. please talks to use and don't proceed with this until we are in agreement if we should support this or not. right now this is explicitly not in scope of edpm ansible based on past PM feedback so this should not be added.
> 
> @SeanMooney We are using configmaps only to configure device_spec and other nova parameters for SRIOV and DPDK. This automation will be used for only SRIOV and NIC partitioning to figure out only available pci addresses and manually configure this device_spec for NIC Partitioning is hard for the user. No updates will be done for SRIOV and nova config is updated for only NIC partitioning mostly.

also we cannot generate configmap using introspect data, need to run scripts inside compute to implement this logic based on  used PCI addresses and configmap device_spec user input.

Jaganathancse:
> as writen this should not merge as this would regress our removal of the term whitelist form our product and its not approatte for the edpm_nova role to modify the sriov agent config.
> 
> on higher level we are try not to generate config on the edpm nodes via ansible and when we raised this question with gil in teh past they said we should not spend time automating this for GA.
> 
> this is not required for the validated architectures
> 
> https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/nfv/sriov/edpm/values.yaml#L166-L169
> 
> and we already have the capability to define this config via configmap associated with the custom nova datapalen service.
@SeanMooney 
Configuring device_spec for NIC Partitioning manaualy is not easy and also we cannot derive outside compute node using introspect data. Need to run script inside compute using used pci addresses and sriov os-net-config details.

SeanMooney:
> > as writen this should not merge as this would regress our removal of the term whitelist form our product and its not approatte for the edpm_nova role to modify the sriov agent config.
> > on higher level we are try not to generate config on the edpm nodes via ansible and when we raised this question with gil in teh past they said we should not spend time automating this for GA.
> > this is not required for the validated architectures
> > https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/nfv/sriov/edpm/values.yaml#L166-L169
> > and we already have the capability to define this config via configmap associated with the custom nova datapalen service.
> > @SeanMooney
> > Configuring device_spec for NIC Partitioning manaualy is not easy and also we cannot derive outside compute node using introspect data. Need to run script inside compute using used pci addresses and sriov os-net-config details.


this has nothing to do  with intersopection data.
then intent was not to make this generitive at all and instead make it declaritive.

meaning the intent was for the admin to determine the correct values and statically configure it in a configmap.


that means if you need to have different config data on different hosts then you would have different nodesets for each host with idffent config and separate config maps for each.

replicating the tripleo behaviour was an non-goal of this approach.

plugins/modules/edpm_nova_derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,671 @@
...
+      - User specified nova conf file for nic device lookup
+    type: str
+    required: false
+    default: '/var/lib/config-data/ansible-generated/neutron-sriov-agent/01-neutron-sriov-agent.conf'
+  save_conf_file:
+    description:
+      - If enabled, the derived file will be saved (backward compatible)
+    type: bool
+    required: false
+    default: false
```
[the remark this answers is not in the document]
vcandapp: We don't update/modify any neutron config in this script - we just read from this (eg):
    NeutronPhysicalDevMappings:
      - "sriov-1:eno35"
      - "sriov-2:eno36"
NovaPCIPassthrough uses deprecated devname (until latest osp17.x release)
The sriov conf file will be "READ" to get the PF name

plugins/modules/edpm_nova_derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,671 @@
...
+    when sriov nic partitioning is configured in EDPM nodes.
+    See openstack user guide for details of
+    NovaPCIPassthrough format.
+options:
+  sriov_nova_conf_file:
+    description:
+      - User specified nova conf file for nic device lookup
+    type: str
+    required: true
+    default: '/var/lib/openstack/config/nova/03-sriov-nova.conf'
```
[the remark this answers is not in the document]
vcandapp: This param has been made as user_configurable. Currently used this "03-sriov-nova.conf", as the NFV QE has used this in thier automation scripts

plugins/modules/edpm_nova_derive_pci_passthrough_whitelist.py
vcandapp: This was just for backward compatible, where the derived pci addr would be written if "save_conf_file" is set. The actual update will be done in "03-sriov-nova.conf"

plugins/modules/edpm_nova_derive_pci_passthrough_whitelist.py
vcandapp: Both "device_spec" and "passthrough_whitelist" will be supported - but the output conf file will have only "device_spec"

plugins/modules/edpm_nova_derive_pci_passthrough_whitelist.py
```diff
@@ -0,0 +1,671 @@
...
+
+ANSIBLE_METADATA = {
+    'metadata_version': '1.0',
+    'status': ['preview'],
+    'supported_by': 'community'
+}
+
+DOCUMENTATION = '''
+---
+module: edpm_nova_derive_pci_passthrough_whitelist
```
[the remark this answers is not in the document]
vcandapp: Will plan to do the removal of term "whitelist" in subsequent  PR updates

vcandapp:
> as writen this should not merge as this would regress our removal of the term whitelist form our product and its not approatte for the edpm_nova role to modify the sriov agent config.
> 
> on higher level we are try not to generate config on the edpm nodes via ansible and when we raised this question with gil in teh past they said we should not spend time automating this for GA.
> 
> this is not required for the validated architectures
> 
> https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/nfv/sriov/edpm/values.yaml#L166-L169
> 
> and we already have the capability to define this config via configmap associated with the custom nova datapalen service.

@SeanMooney
As we have got the +1 from Compute for this merge, so proceeding to submit the patch <removing reference to the "whitelist" keywork. 
Requesting you to do the necessary

SeanMooney:
> > as writen this should not merge as this would regress our removal of the term whitelist form our product and its not approatte for the edpm_nova role to modify the sriov agent config.
> > on higher level we are try not to generate config on the edpm nodes via ansible and when we raised this question with gil in teh past they said we should not spend time automating this for GA.
> > this is not required for the validated architectures
> > https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/nfv/sriov/edpm/values.yaml#L166-L169
> > and we already have the capability to define this config via configmap associated with the custom nova datapalen service.
> 
> @SeanMooney This derive automation is for ease-of-use purpose As we have got the +1 from Compute for this merge, so proceeding to submit the patch <removing reference to the "whitelist" keywork> Requesting you to do the necessary

you have not updated this with the different changes we requested so i don't think we whoudl merve this until those refactorings are done.
namely we should have molecule and unit testing for the python script and the removal of whitelist ectra should be done.

ideally the removal of reading form the neutron config should also be done.

this change is introducing and ordering depency between the edpm_nova and edpm_neutron_sriov roles

instead this should be refactored to use the same input data source as edpm_neutron_sriov instead of reading the rendered config.

gibizer:
Here is my understanding and acceptance criteria based on our yesterday's call:
* For GA:
  * minor comments about naming and role dependencies are solved.
  * automated testing is added. Unit test for the python module. Molecule test for the ansible role changes. (Also if you did manual end to end testing then a link to the description of the manual test steps would be nice)
  * adoption is verified. 
* For FP1:
  * design a generic and well defined node introspection interface that gathers HW data and presents an API to all ansible roles that needs such data. This is needed to prevent individual introspection implementations per feature and pre role.

roles/edpm_nova/tasks/derive_pci.yml
```diff
@@ -0,0 +1,38 @@
...
+         User provided nova conf for pci device_spec is
+         "{{ edpm_nova_config_dest }}/{{ edpm_nova_default_sriov_conf_file }}"
+
+- name: Derive the sriov nic-partitioned pci device_spec addr
+  when:
+    - sriov_nova_conf_check.stat.exists
+  become: true
+  block:
+    - name: Run derive_pci_passthrough_devicespec
+      edpm_nova_derive_pci_passthrough_devicespec:
```
bogdando, line 34: As Sean mentioned, configuring services in EDPM violates today architecture, where we only provide ready configs as raw snippets to apply it for openstack dataplane nodesets (osdpns).

Perhaps this ansible automation and this custom module should be used on the other side of the data inputs, where we should call kustomize to build osdpns with expected config overrides, before to apply it for EDPM?

Imagine a large multi-cell deployment of pre-provisioned nodes.
We will need to produce an ansible inventory then, so that such automation could split it into "osdpns chunks" and populate it with expected roles to deploy, and configs snippets to inject.

Finally, kustomize should apply it according to the supported validated architecture (or a deployment topology).

For infrastructure provisioned hosts, the situation is more complicated as we will need this ansible inventory prepared with the help of the baremetal operator perhaps? @rabi WDYT?

SeanMooney:
this brnach should not have any merge commits in it so it need to be rebased insted.

plugins/modules/edpm_nova_derive_pci_passthrough_devicespec.py
```diff
@@ -0,0 +1,649 @@
...
+
+    :param user_configs: THT param NovaPCIPassthrough
+    :param pf_name: Interface/device name (str)
+    :param is_non_nic_pf: Indicates whether the PF is noc-partitioned or not
+    :return: Return the actions to be done, based on match criteria
+    """
+
+    # get the vendor and product id of the PF and VF
+    pf_path = os.path.join(_SYS_CLASS_NET_PATH, pf_name, "device")
+    vendor, vf_product = get_pci_device_info_by_ifname(pf_path, 'virtfn0')
```
SeanMooney, line 302: i am not sure that virtfn0 is portable across vendors.

plugins/modules/edpm_nova_derive_pci_passthrough_devicespec.py
```diff
@@ -0,0 +1,649 @@
...
+
+    allocated_pci = get_allocated_pci_addresses(system_configs)
+    nic_partition_pfs = get_sriov_nic_partition_pfs(system_configs)
+
+    for user_config in user_configs:
+        if 'address' in user_config and 'devname' in user_config:
+            msg = f"Both devname and address can't be present in {_PASSTHROUGH_DEVSPEC_KEY}"
+            raise InvalidConfigException(msg)
+
+        keys = ['address', 'product_id', 'devname']
```
SeanMooney, line 531: if product id is specified then vendor id should be also.

plugins/modules/edpm_nova_derive_pci_passthrough_devicespec.py
```diff
@@ -0,0 +1,649 @@
...
+        if not any(k in user_config for k in keys):
+            # address or product_id or devname not present in user_config
+            pf = get_pf_name_from_phy_network(user_config['physical_network'],
+                                              phys_dev_map)
+            user_config['address'] = get_pci_addr_from_ifname(pf)
+
+        if 'devname' in user_config:
```
SeanMooney, line 538: we should be removing all usages fo devname.

its not something that customer should be using in there config.

plugins/modules/edpm_nova_derive_pci_passthrough_devicespec.py
```diff
@@ -0,0 +1,649 @@
...
+    :return derived_config: All nic partinioned cfg will be returned after derivation in this list
+    """
+
+    user_config_copy = []
+    derived_config = []
+
+    allocated_pci = get_allocated_pci_addresses(system_configs)
+    nic_partition_pfs = get_sriov_nic_partition_pfs(system_configs)
+
+    for user_config in user_configs:
```
SeanMooney, line 526: if we have this script it should just hard error if devname is present.

plugins/modules/edpm_nova_derive_pci_passthrough_devicespec.py
```diff
@@ -0,0 +1,649 @@
...
+    if user_config_copy != user_configs:
+        for item in (user_config_copy + derived):
+            line = _PASSTHROUGH_DEVSPEC_KEY + '=' + json.dumps(item)
+            lines += "".join(line) + '\n'
+        # Replace the new derived devspec into conf
+        with open(sriov_nova_conf_file) as nova_config:
+            file_data = nova_config.read()
+            update = file_data.replace(input_config, lines)
+        with open(sriov_nova_conf_file, 'w') as nova_config:
+            nova_config.write(update)
```
SeanMooney, line 612: this script should not be using a nova config files as input but even if we ignore that its not ok for this script to modify that.

SeanMooney (changes requested):
reviewing the python script in more detail it does not seam to be fit for purpose.

it has code for merging config which should not exits
the python script should be providing a conf snipit with a lower precedence than the user provided custom config which will be merged by oslo config.

we have fundamentally changes how we do configmanagment in the new installer.

this script is not following any of the conventions for that and should be rewritten to align in my view before we consider merging this pr.

Jaganathancse:
@SeanMooney Mostly conf files are designed using unique key value pair. May i know why device_spec configuration format is supporting duplicate entries for list of different device_spec values [instead of key and list of dict]?

SeanMooney:
> @SeanMooney Mostly conf files are designed using unique key value pair. May i know why device_spec configuration format is supporting duplicate entries for list of different device_spec values [instead of key and list of dict]?

it is a multiopt filed
```
multi valued (MultiStrOpt)

    A multi-valued option is a string value and can be given more than once, all values will be used.

    # Driver or drivers to handle sending notifications. (multi valued)
    notification_driver = nova.openstack.common.notifier.rpc_notifier
    notification_driver = ceilometer.compute.nova_notifier
```

https://docs.openstack.org/nova/latest/configuration/config.html#pci.device_spec

it has always been a multi valued field even before the rename


we do support a json list style syntax also 

device_spec = [{"product_id":"0001", "vendor_id":"8086"},
               {"product_id":"0002", "vendor_id":"8086"}]

but the primary way this is used is relying on the multi value merging provide by oslo.

for other related config options like the PCI alias

https://docs.openstack.org/nova/latest/configuration/config.html#pci.alias
we only support the multiple line syntax.

we do not support the json list form so the muli value syntax is the common approach that works for both.

vcandapp:
Status: After discussing with all stakeholders on the SR-IOV NIC-partitioing deployment, it came out that ease-of-use is not a requirement for NIC paritioning. Also, modifying nova configmap (which this module updates, for single-step deployment) is not permitted
**Decision taken:**
Even though we have reservations from customers point-of-view on this approach, we will created a new role edpm_derive_pci_device_spec for the pci device_spec derivation. 
Follow the two step-deployment approach (1. Do a provision upto a point where this role generates the configmap 2. SSH into compute to get the exact sriov device_spec configmap and continue  by using this map for nova, complete the deployment for remaining services)

PR #573, opened 2024-02-27
[edpm_ovn] Add support for configuring Availability zones

karelyatin:
Role var is added to configure availability zones in external-ids:ovn-cms-options.

Related-Issue: [OSPRH-5009](https://issues.redhat.com//browse/OSPRH-5009)

### DIFF
--- a/roles/edpm_ovn/meta/argument_specs.yml
+++ b/roles/edpm_ovn/meta/argument_specs.yml
@@ -31,6 +31,10 @@ argument_specs:
           Should OVN use tls for default protocol?
           If set to false, the OVN will default to tcp.
         type: bool
+      edpm_ovn_availability_zones:
+        default: []
+        description: Availability zones for the chassis
+        type: list
       edpm_ovn_bridge:
         default: br-int
         description: >
--- a/roles/edpm_ovn/molecule/availability_zones/converge.yml
+++ b/roles/edpm_ovn/molecule/availability_zones/converge.yml
@@ -1,0 +1,29 @@
+---
+# Copyright 2024 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Converge
+  hosts: all
+  gather_facts: true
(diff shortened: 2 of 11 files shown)


roles/edpm_ovn/meta/argument_specs.yml
```diff
@@ -31,6 +31,10 @@ argument_specs:
           Should OVN use tls for default protocol?
           If set to false, the OVN will default to tcp.
         type: bool
+      edpm_ovn_availability_zones:
+        default: []
+        description: Availaiblity zones for the chassis
```
booxter, line 36: availab_i_lity

roles/edpm_ovn/molecule/availability_zones/converge.yml
```diff
@@ -0,0 +1,29 @@
+---
+# Copyright 2021 Red Hat, Inc.
```
booxter, line 2: 2024

roles/edpm_ovn/tasks/cleanup.yml
```diff
@@ -43,6 +43,20 @@
     - not edpm_enable_chassis_extport | bool
     - cleanup_ovn_cms_options.rc == 0
 
+- name: Cleanup availability-zones
+  block:
+    - name: Cleanup availability-zones from stdout when undefined
+      ansible.builtin.set_fact:
+        filtered_azs:
+          stdout: "{{ cleanup_ovn_cms_options.stdout | regex_replace('availability-zones=[^\"]*' ~ ',?', '') }}"
```
booxter, line 51: AFAIU this has an assumption that may fire back later, namely that the zones is the last entry in the options list (hence "read till double-quote" instruction in the regexp.)

My first preference would be to make it not make this assumption; my second preference would be to add a comment here documenting this assumption (perhaps also in the `configure` playbook, so that whenever we touch the latter, we are aware that some other piece of code relies on the order.)

My final concern is that cleanup is not validated in tests, so if there's some problem with the regex used, we won't notice it until production deployments.

booxter: Perhaps...?

```python
>>> x
'availability-zones=az1:az2,blabla=xxx'
>>> re.sub('availability-zones=[^\"]*,?', '', x)
''
>>> re.sub('availability-zones=[^,]*,?', '', x)
'blabla=xxx'
```

booxter (changes requested):

roles/edpm_ovn/tasks/cleanup.yml
karelyatin: Done, my bad somehow local changes got missed for this

roles/edpm_ovn/molecule/availability_zones/converge.yml
karelyatin: Done

roles/edpm_ovn/meta/argument_specs.yml
karelyatin: Done

PR #577, opened 2024-02-28
Use the same variable name as all other variables

jlarriba:


fao89 (approved):

jpodivin (changes requested):
Hi. The change is good, but you need to replace the `telemetry_node_exporter_image` in the role as well. https://github.com/openstack-k8s-operators/edpm-ansible/blob/5cc279c1bc69e06719d06d8a0707054d5fa2a1a9/roles/edpm_telemetry/tasks/deploy_infra.yml#L55

jlarriba:
Just realized that the rebase on main done in Github introduced a second telemetry_node_exporter_image variable.

Fixed it.

roles/edpm_telemetry/defaults/main.yml
```diff
@@ -21,7 +21,7 @@ edpm_telemetry_config_src: /var/lib/openstack/configs/telemetry
 # Directory in the compute node
 edpm_ceilometer_config_dest: /var/lib/openstack/config
 # Image to use for node_exporter
-telemetry_node_exporter_image: quay.io/prometheus/node-exporter:v1.5.0
+edpm_node_exporter_image: quay.io/prometheus/node-exporter:v1.5.0
```
slagle, line 24: Since you're changing the name, can we go ahead and change it to match our guidelines? See docs/source/contributing_roles.rst. It should be `edpm_telemetry_node_exporter_image`.

jlarriba: Done, that also requires that the default variables in the dataplane-operator meet that same guidelines. That has been pushed in this PR: https://github.com/openstack-k8s-operators/dataplane-operator/pull/722

PR #579, opened 2024-02-28
Replace /var/tmp/edpm-ansible with $REMOTE_SOURCE_DIR

slagle:
The variable was not used everywhere it could be used.


fao89 (approved):

fao89:
recheck

PR #580, opened 2024-02-29
Add adoption task to nova role

mrkisaolamb:


roles/edpm_nova/tasks/adoption.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Stop and disable compute services
+  no_log: "{{ edpm_network_config_hide_sensitive_logs | bool }}"
+  tags:
+    - adoption
+    - nova
+  become: true
+  ansible.builtin.shell: |
```
gibizer, line 8: can we use ansible.builtin.systemd instead?

roles/edpm_nova/defaults/main.yml
```diff
@@ -31,3 +31,14 @@ edpm_nova_compute_image: "quay.io/podified-antelope-centos9/openstack-nova-compu
 
 # certs
 edpm_nova_tls_certs_enabled: "{{ edpm_tls_certs_enabled | default(False) }}"
+# default list example
+edpm_nova_compute_services_to_stop:
```
gibizer, line 35: would use a name that signals that these are old tripleo services. What about `edpm_old_tripleo_compute_sevices`

roles/edpm_nova/defaults/main.yml
```diff
@@ -31,3 +31,14 @@ edpm_nova_compute_image: "quay.io/podified-antelope-centos9/openstack-nova-compu
 
 # certs
 edpm_nova_tls_certs_enabled: "{{ edpm_tls_certs_enabled | default(False) }}"
+# default list example
```
gibizer, line 34: probably copy paste

roles/edpm_nova/tasks/adoption.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Stop and disable compute services
+  no_log: "{{ edpm_network_config_hide_sensitive_logs | bool }}"
```
gibizer, line 3: hm, this variable seems to be intended to the network_config roles. Do we even foresee some sensitive logs here?

gibizer (commented):
Do we want to disable these adoption tasks in the greenfield deployment? If so, we need to somehow disable the adoption tag in greenfield.

gibizer:
I'm wondering how easy would be to create a meaningful molecule test for this task. @SeanMooney do you have some ideas how to simulate the existence of an old tripleo service in molecule?

gibizer:
@mrkisaolamb I'm not sure depends-on works between edpm-ansible and data-plane-adoption repo, but it worth a try to see if we can run this in the adoption CI even before merge

roles/edpm_nova/defaults/main.yml
mrkisaolamb: Done

roles/edpm_nova/tasks/adoption.yml
mrkisaolamb: done, removed

roles/edpm_nova/tasks/adoption.yml
mrkisaolamb: done

roles/edpm_nova/tasks/adoption.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Stop and disable compute services
+  tags:
+    - adoption
+    - nova
+  become: true
+  ansible.builtin.systemd:
+        name: "{{ item }}"
+        state: stopped
+        enabled: false
```
gibizer, line 10: we mask them as well with `masked: yes`

gibizer (commented):
Thanks. I think we need some pidfile deletion as well. Or have you checked, stopping the service removes the pidfiles too?

roles/edpm_nova/tasks/adoption.yml
mrkisaolamb: done

mrkisaolamb:
recheck

roles/edpm_nova/tasks/adoption.yml
bogdando: have you pushed the updated revision yet?

mrkisaolamb: I re-add this shell to simplified checking if service exist from list, task with systemd was failing if service wasn't running

mrkisaolamb:
recheck

roles/edpm_nova/tasks/adoption.yml
```diff
@@ -0,0 +1,30 @@
...
+  tags:
+    - adoption
+    - nova
+  block:
+    - name: Check if nova_virtlogd container exists
+      ansible.builtin.command: podman ps -a --filter name=nova_virtlogd --format \{\{.Names\}\}
+      register: nova_virtlogd_status
+      changed_when: false
+    - name: Stop nova_virtlogd containers not managed by service units
+      ansible.builtin.shell: "podman stop nova_virtlogd '||' true"
```
gibizer, line 14: Does podman return a nonzero exitcode if it succeeds? Do we need to ignore all the exit codes?

gibizer:
> Thanks. I think we need some pidfile deletion as well. Or have you checked, stopping the service removes the pidfiles too?

Did we settled this?

gibizer (approved):
So Kamil checked and the pidfiles are removed by the systemd service files. 
Also the systemd module is not used as based on Kamil it cannot stop already disabled services. 
The rest of my comments can be fixed in a follow up. So I approve this to see if it helps the adoption testing where we see virtlogd issues.

PR #581, opened 2024-02-29
[edpm_nova]Remove cert copy task

gibizer:
The cert management is moved to a separate role by #525 but a rebase
mistake in ca102fc reintroduced the task to the nova role.

### DIFF
--- a/roles/edpm_nova/tasks/configure.yml
+++ b/roles/edpm_nova/tasks/configure.yml
@@ -120,23 +120,6 @@
     owner: nova
     group: nova
 
-- name: Copy TLS files to the compute node
-  tags:
-    - configure
-    - nova
-  become: true
-  loop:
-    - {"src": "{{ edpm_nova_certs_src }}/{{ inventory_hostname }}-tls.crt", "dest": "{{ edpm_nova_certs_dest }}/tls.crt"}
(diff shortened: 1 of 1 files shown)


SeanMooney (approved):
i mentioned this on slack but the nova role will need to have tasks to copy the certs into the correct place and this was a best efferort by ade to try and do that. it didn't actually work end to end and did not have test coverage so it has been broken at some point

we can remove this for now as i will need to reimplemt this from scratch in the near future.

but just for clarity there is an expectation that we will have a task like this in the future in this task file when we add support for tlse for nova

gibizer:
> i mentioned this on slack but the nova role will need to have tasks to copy the certs into the correct place and this was a best efferort by ade to try and do that. it didn't actually work end to end and did not have test coverage so it has been broken at some point
> 
> we can remove this for now as i will need to reimplemt this from scratch in the near future.
> 
> but just for clarity there is an expectation that we will have a task like this in the future in this task file when we add support for tlse for nova

As I wrote in the commit message Ade's intention was to remove this task it just reintroduce by an unrelated commit that was rebased badly. I agree that we need TLS functionality, but that will be done independently

SeanMooney:
actuuly looking at your refenced commit this was likely a rebase artefact although ti does not change the fact that its broken and should be removed and then replaced with a tested working version in a future pr

SeanMooney:
check-rdo baremetal job failed.

PR #586, opened 2024-03-04
[edpm_nova]Remove certificate mounts when TLS enabled

vyzigold:
The feature isn't fully implemented yet and the certificates aren't at the specified location. This means, that the ansible fails when deploying with TLS enabled.

### DIFF
--- a/roles/edpm_nova/templates/nova_compute.json.j2
+++ b/roles/edpm_nova/templates/nova_compute.json.j2
@@ -10,11 +10,6 @@
     },
     "volumes": [
         "/var/lib/openstack/config/nova:/var/lib/kolla/config_files:ro",
-{% if edpm_nova_tls_certs_enabled %}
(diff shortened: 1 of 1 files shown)


gibizer (commented):
it seem to be a logical follow up for #581 but I defer to @SeanMooney to decide if we can drop this

SeanMooney (approved):
we can drop this if we ant to but right now we do not suppro TLS on any EDPM node with libvirt or nova

so we can delete this if we want but form my perspective if you want to enable TLS in the ndoe set then you should not be including the nova or libvirt services


since people seam to be interested in testing invalid configurations :) ill approve this for now and just implement the nova part form scratch. it really does not add much extra work and it will ublock those trying to use tls-e before its ready.

vyzigold:
recheck
```
Failed to download metadata for repo 'kubernetes': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

PR #588, opened 2024-03-04
ADD tls-e support for edpm_nova

SeanMooney:
Nova already connected to libvirt via unix socket so it does not
need the libvirt tls certs for local communcitaiton. nova does
not need to connect to remote libvirt, as a result we do not need
to provide access to the libvirt client cert.
The 02-nova-host-specific.conf.j2 template has been updated to
enable libvirt native tls for live migration when tls is enabled.

molecule test coverage is added for the above changes.

Closes: [OSPRH-5053](https://issues.redhat.com//browse/OSPRH-5053)

### DIFF
--- a/roles/edpm_nova/defaults/main.yml
+++ b/roles/edpm_nova/defaults/main.yml
@@ -31,6 +31,11 @@ edpm_nova_compute_image: "quay.io/podified-antelope-centos9/openstack-nova-compu
 
 # certs
 edpm_nova_tls_certs_enabled: "{{ edpm_tls_certs_enabled | default(False) }}"
+# NOTE(sean-k-mooney): nova will use unix sockets for libvirt and communicate with ovs via tcp
+# so we will not need the libvirt or ovs client certs or ca. nova will communicate other services
+# via there api endpoints and will connect to rabbitmq. To support this we will need to trust
+# the general ca root cert.
+edpm_nova_tls_ca_src_dir: "/var/lib/openstack/cacerts/{{ edpm_service_name | default('nova') }}"
 
 # list of services to stop during adoption
 edpm_old_tripleo_compute_sevices:
--- a/roles/edpm_nova/templates/02-nova-host-specific.conf.j2
+++ b/roles/edpm_nova/templates/02-nova-host-specific.conf.j2
@@ -1,2 +1,7 @@
 [DEFAULT]
 my_ip = {{ ctlplane_ip }}
+{% if edpm_nova_tls_certs_enabled | bool %}
+[libvirt]
+live_migration_with_native_tls = true
+live_migration_scheme = tls
+{% endif %}
--- a/roles/edpm_nova/templates/nova_compute.json.j2
+++ b/roles/edpm_nova/templates/nova_compute.json.j2
@@ -10,10 +10,12 @@
     },
     "volumes": [
         "/var/lib/openstack/config/nova:/var/lib/kolla/config_files:ro",
+{% if edpm_nova_tls_certs_enabled %}
+	"{{ edpm_nova_tls_ca_src_dir }}/tls-ca-bundle.pem:/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem:ro",
+{% endif %}
         "/etc/localtime:/etc/localtime:ro",
         "/lib/modules:/lib/modules:ro",
         "/dev:/dev",
-        "/run/openvswitch:/run/openvswitch",
         "/var/log/containers/nova:/var/log/containers/nova",
         "/var/lib/libvirt:/var/lib/libvirt",
         "/run/libvirt:/run/libvirt:shared",
--- a/roles/edpm_nova/molecule/default/collections.yml
+++ b/roles/edpm_nova/molecule/default/collections.yml
@@ -1,0 +1,4 @@
+
+collections:
+  - name: community.crypto
+    type: galaxy
--- a/roles/edpm_nova/molecule/default/converge.yml
+++ b/roles/edpm_nova/molecule/default/converge.yml
@@ -15,3 +15,5 @@
     - role: osp.edpm.edpm_nova
       vars:
         edpm_nova_config_src: "{{lookup('env', 'MOLECULE_PROJECT_DIRECTORY')}}/molecule/default/test-data"
+        edpm_nova_tls_ca_src_dir: /tmp/pki
+        edpm_nova_tls_certs_enabled: true
--- a/roles/edpm_nova/molecule/default/prepare.yml
+++ b/roles/edpm_nova/molecule/default/prepare.yml
@@ -23,6 +23,8 @@
 
 - name: Setup DUT
   hosts: all
+  vars:
+    edpm_nova_tls_ca_src_dir: /tmp/pki
   pre_tasks:
     - name: set basic user fact
       set_fact:
@@ -107,3 +109,48 @@
         owner: root
         group: root
         mode: '0655'
+
+    - name: create pki temp directory
+      become: true
+      ansible.builtin.file:
+        path: "{{ edpm_nova_tls_ca_src_dir }}"
+        state: directory
+        owner: root
+        group: root
+        mode: 0777
+
+    - name: Create private key (RSA, 4096 bits)
+      community.crypto.openssl_privatekey:
+        path: "{{ edpm_nova_tls_ca_src_dir }}/ca.key"
+
+    - name: Create a certificate signing request for ca
+      community.crypto.openssl_csr_pipe:
+        privatekey_path: "{{ edpm_nova_tls_ca_src_dir }}/ca.key"
+        organization_name: fake-corp
+        basic_constraints:
+          - 'CA:TRUE'
+        subject_alt_name:
+          - "DNS:compute-1.example.com"
+          - "DNS:compute-1"
+          - "DNS:{{ inventory_hostname }}"
+        key_usage:
+          - keyCertSign
+      register: csr
+
+    - name: Create a certificate athority
+      community.crypto.x509_certificate:
+        path: "{{ edpm_nova_tls_ca_src_dir }}/tls-ca-bundle.pem"
+        privatekey_path: "{{ edpm_nova_tls_ca_src_dir }}/ca.key"
+        csr_content: "{{ csr.csr }}"
+        provider: selfsigned
+
+    # FIXME(sean-k-mooney): this is a hack to work around the fact that we dont
+    # currently manage the hostname on the DUT via boostrap or a dedicated role
+    # in the molecule test. This is needed to ensure the hostname is resolvable
+    # when executing virsh commands. remove this when we have a proper solution
+    - name: Ensure hostname is resolvable in /etc/hosts
+      become: true
+      ansible.builtin.lineinfile:
+        path: /etc/hosts
+        line: "{{ ansible_default_ipv4.address | default('127.0.0.1') }} {{ ansible_fqdn }} {{ inventory_hostname }}"
+        state: present
--- a/roles/edpm_nova/molecule/default/verify.yml
+++ b/roles/edpm_nova/molecule/default/verify.yml
@@ -6,6 +6,7 @@
   gather_facts: false
   vars:
     test_helper_dir: "../../../../molecule/test-helpers"
+    edpm_nova_tls_ca_src_dir: /tmp/pki
   tasks:
     - name: ensure expected directories exist
       ansible.builtin.include_tasks: "{{test_helper_dir}}/verify_dir.yaml"
@@ -25,6 +26,9 @@
         - "/var/lib/_nova_secontext"
         - "/var/lib/nova/instances"
         - "/var/log/containers/nova"
+        # NOTE(sean-k-mooney): this directory is normaly created by the edpm_install_cert role
+        # in molecule we create it in prepare so lets just assert it exists
+        - "{{ edpm_nova_tls_ca_src_dir }}"
     - name: ensure systemd services are defined and functional
       ansible.builtin.include_tasks: "{{test_helper_dir}}/verify_systemd_unit.yaml"
       loop:
@@ -55,6 +59,12 @@
       ansible.builtin.assert:
         that:
           - "'10.0.0.3' in host_specific_config.content | b64decode"
+
+    - name: Assert libvirt is configured to use tls for live migration
+      ansible.builtin.assert:
+        that:
+          - "'live_migration_with_native_tls = true' in host_specific_config.content | b64decode"
+          - "'live_migration_scheme = tls' in host_specific_config.content | b64decode"
 
     - name: Check if user exists
       ansible.builtin.getent:
--- a/roles/edpm_nova/molecule/vagrant/collections.yml
+++ b/roles/edpm_nova/molecule/vagrant/collections.yml
@@ -1,0 +1,1 @@
+../default/collections.yml


roles/edpm_nova/templates/nova_compute.json.j2
```diff
@@ -11,14 +11,12 @@
...
-	"/var/lib/openstack/certs/nova/tls.crt:/etc/pki/nova/server-cert.pem:ro",
-	"/var/lib/openstack/certs/nova/tls.key:/etc/pki/nova/private/server-key.pem:ro",
-	"/var/lib/openstack/cacerts/nova/tls-ca-bundle.pem:/etc/pki/CA/cacert.pem:ro",
+	"{{ edpm_nova_tls_ca_src_dir }}/tls-ca-bundle.pem:/etc/pki/CA/cacert.pem:ro",
 {% endif %}
         "/etc/localtime:/etc/localtime:ro",
         "/lib/modules:/lib/modules:ro",
         "/dev:/dev",
-        "/run/openvswitch:/run/openvswitch",
+        "/run/openvswitch:/run/openvswitch:shared",
```
SeanMooney, line 19: fyi the only reason i added shared is for parity with libvirt and incase for some reason the nova_compute container stated before ovs on the host

in that case shared should allow nova to see the ovsdb socket when it gets created on the host.

SeanMooney:
i have not tested this outside of molecule so i will try and create an env with tls enabled.

it would be ideal if podified-multinode-edpm-deployment-crc or a similar tempst zuul job could be used to validate this instead.

i currently do not have a CRC deploy so it will take me a day or too to test this end to end.
if anyone can test it quickly in the mean time then please update this with your findings.

roles/edpm_nova/templates/02-nova-host-specific.conf.j2
```diff
@@ -1,2 +1,9 @@
 [DEFAULT]
 my_ip = {{ ctlplane_ip }}
+[os_vif_ovs]
+ovsdb_connection="unix:/var/run/openvswitch/db.sock"
+{% if edpm_nova_tls_certs_enabled | bool %}
+[libvirt]
+live_migration_with_native_tls = true
+live_migration_scheme = tls
+{% endif %}
```
gibizer, line 9: OK, we agreed that we don't have information in nova-operator if TLS is enabled in the dataplane or not, so we cannot provide this information via the cell config. If in the future we make every nova compute config a j2 template then we can  move this template logic to the cell config. However we might not want to as that will couple the nova-operator with the edpm ansible vars.

gibizer:
recheck 
tempest failed. It seem that nova-compute deployed successfully based on the ansible logs, but there is 0 nova-compute log is visible

SeanMooney:
This was the error

```
Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.317 2 ERROR ovsdbapp.backend.ovs_idl.idlutils [None req-79ae71e2-835e-4c4a-9bad-800951cff8be efbad3f6e4b14cfda6834d8f64599d1f f21742f9e5b84ad1a4c24d0024a0d988 - - default default] Unable to open stream to unix:/var/run/openvswitch/db.sock│
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif [None req-79ae71e2-835e-4c4a-9bad-800951cff8be efbad3f6e4b14cfda6834d8f64599d1f f21742f9e5b84ad1a4c24d0024a0d988 - - default default] Failed to plug vif VIFOpenVSwitch(active=False,address=fa:16:3e:50:d5:78,bridge_name=┤
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif Traceback (most recent call last):                                                                                                                                                                                         │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif   File "/usr/lib/python3.9/site-packages/os_vif/__init__.py", line 77, in plug                                                                                                                                             │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif     plugin.plug(vif, instance_info)                                                                                                                                                                                        │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif   File "/usr/lib/python3.9/site-packages/vif_plug_ovs/ovs.py", line 347, in plug                                                                                                                                           │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif     self._plug_vif_generic(vif, instance_info)                                                                                                                                                                             │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif   File "/usr/lib/python3.9/site-packages/vif_plug_ovs/ovs.py", line 280, in _plug_vif_generic                                                                                                                              │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif     self.ovsdb.ensure_ovs_bridge(vif.network.bridge,                                                                                                                                                                       ┤
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif   File "/usr/lib/python3.9/site-packages/vif_plug_ovs/ovsdb/ovsdb_lib.py", line 75, in ensure_ovs_bridge                                                                                                                   │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif     return self.ovsdb.add_br(bridge, may_exist=True,                                                                                                                                                                       │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif   File "/usr/lib/python3.9/site-packages/vif_plug_ovs/ovsdb/ovsdb_lib.py", line 41, in ovsdb                                                                                                                               │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif     self._ovsdb = ovsdb_api.get_instance(self)                                                                                                                                                                             │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif   File "/usr/lib/python3.9/site-packages/vif_plug_ovs/ovsdb/api.py", line 28, in get_instance                                                                                                                              │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif     return iface.api_factory(context)                                                                                                                                                                                      │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif   File "/usr/lib/python3.9/site-packages/vif_plug_ovs/ovsdb/impl_idl.py", line 40, in api_factory                                                                                                                          │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif     idl=idl_factory(config),                                                                                                                                                                                               │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif   File "/usr/lib/python3.9/site-packages/vif_plug_ovs/ovsdb/impl_idl.py", line 32, in idl_factory                                                                                                                          ┤
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif     helper = idlutils.get_schema_helper(conn, schema_name)                                                                                                                                                                 │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif   File "/usr/lib/python3.9/site-packages/ovsdbapp/backend/ovs_idl/idlutils.py", line 215, in get_schema_helper                                                                                                             │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif     return create_schema_helper(fetch_schema_json(connection, schema_name))                                                                                                                                                ┤
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif   File "/usr/lib/python3.9/site-packages/ovsdbapp/backend/ovs_idl/idlutils.py", line 204, in fetch_schema_json                                                                                                             │
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif     raise Exception("Could not retrieve schema from %s" % connection)                                                                                                                                                      ┤
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif Exception: Could not retrieve schema from unix:/var/run/openvswitch/db.sock                                                                                                                                                ┤
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.318 2 ERROR os_vif #033[00m                                                                                                                                                                                                                   ┤
│Mar  5 04:17:54 np0004517961 nova_compute[117833]: 2024-03-05 09:17:54.320 2 ERROR nova.virt.libvirt.driver [None req-79ae71e2-835e-4c4a-9bad-800951cff8be efbad3f6e4b14cfda6834d8f64599d1f f21742f9e5b84ad1a4c24d0024a0d988 - - default default] [instance: dd0b7b57-7334-4ffb-b6f0-960a81d2baa7] Failed to start li┤
│M
```

i have been using lnav -q https://logserver.rdoproject.org/88/588/23fe056c41b9d2bba62646efa8f781f7f847a7f8/github-check/podified-multinode-edpm-deployment-crc/78957db/controller/ci-framework-data/logs/192.168.122.100/log/messages

to view them

olliewalsh:
(so not sure how nova-compute could previously have been connecting to it over TLS)

roles/edpm_nova/defaults/main.yml
```diff
@@ -31,6 +31,11 @@ edpm_nova_compute_image: "quay.io/podified-antelope-centos9/openstack-nova-compu
 
 # certs
 edpm_nova_tls_certs_enabled: "{{ edpm_tls_certs_enabled | default(False) }}"
+# NOTE(sean-k-mooney): nova will use unix sockets for libvirt and ovs communication so we
+# will not need the libvirt or ovs client certs or ca. nova will communicate other services
+# via there api endpoints and will connect to rabbitmq. To support this we will need to trust
+# the general ca root cert.
+edpm_nova_tls_ca_src_dir: /var/lib/openstack/cacerts/nova
```
olliewalsh, line 38: on my deployment it's /var/lib/openstack/cacerts/nova-custom (OpenStackDataPlaneService name). @vakwetu FYI

SeanMooney: hum that was not ment ot change based on the dataplane service name
i don't think we have access to that in ansible

roles/edpm_nova/templates/nova_compute.json.j2
```diff
@@ -10,10 +10,13 @@
     },
     "volumes": [
         "/var/lib/openstack/config/nova:/var/lib/kolla/config_files:ro",
+{% if edpm_nova_tls_certs_enabled %}
+	"{{ edpm_nova_tls_ca_src_dir }}/tls-ca-bundle.pem:/etc/pki/CA/cacert.pem:ro",
```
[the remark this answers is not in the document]
SeanMooney: this is the path @vakwetu previously used so i just used the same but i can mount it to 
/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem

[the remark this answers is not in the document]
olliewalsh: ack, fixed that in https://github.com/openstack-k8s-operators/edpm-ansible/pull/591, it should have been the libvirt CA, not the bundle (as it's mTLS)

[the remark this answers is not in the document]
olliewalsh: ... not sure why libvirt uses such a generic path for it's CA cert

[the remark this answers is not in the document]
SeanMooney: ya i asked them and they mentioned it was also probably a poor naming choice. they have libvirt in thename or path for all the rest except that root ca cert.

roles/edpm_nova/defaults/main.yml
vakwetu: the dataplane-operator doesn't know anything about whether the service name is nova or nova-custom or whatever.   It just knows that there is a dataplane service named X and so it puts its ca bundle under /var/lib/openstack/cacerts/X  or certs and keys (and now ca.certs) under /var/lib/openstack/certs/X. 

We can fix this in two ways.  1) We can add something to the service spec -- something like "mountAlias" that tells the dataplane-operator to use a different path instead.  2) we can add an ansible param that specifies the service name

SeanMooney: i tought about both 2 would only work if differnt named service shared the same tls ca

i.e. if two libivrt serivce had the same ca

i dont think you can do that automatically

so instead of a mount alias i think we need to extend the dataplane service with a type field.
then use that to generate the crts and deifne the mount point.

so type=ovs, type=libvirt or type=nova
that woudl map to 
type=nova with name=nova-custom woudl result in the tls certs beign placed in .../nova ignoring the service name "nova-custom"

/var/lib/openstack/cacerts

do we think that is feisable?

vakwetu: So, edpm-nova handles this discrepancy between nova and nova-custom by discovering the config files https://github.com/openstack-k8s-operators/edpm-ansible/blob/f98ed5f09467620a00e2936c27a80120b634a56c/roles/edpm_nova/tasks/configure.yml#L70  and copying them to the right place on the compute node.

install-certs has no such logic.  We just copy the right certs, keys, cacerts for the node from /var/lib/openstack/cacerts in the ansibleEE pod to /var/lib/openstack/cacerts on the compute node.  Same for /var/lib/openstack/certs.

So, we can solve this either by mounting the certs in the ansibleEE container to the right place (/var/lib/openstack/cacerts/nova) to begin with -- by using the mountAlias thing I just mentioned.

Or we can add an ansible param that provides the service-name and then have the edpm-nova role pick things up from the right path.

Which do you guys prefer?

vakwetu: gah - didn't refresh to see your response .. reading ..

vakwetu: @SeanMooney OK - so I think we're considering something like option 1 - where we have a field that redefines where the certs and cacerts are mounted in the ansibleEE pod to begin with.

The difference is that you call the attribute "type" - which might be useful later in other contexts too - maybe to later simplify the configMap mounts for instance.

The implicit assumption of course is that there is only one dataplane service defined with each "type".

Yeah, this is feasible.  I'll start work on it and get something up for review by early next week.

SeanMooney: yes so i briefly ran the type idea past gibi and his reaction is that would allow use to add validating/default webhooks for the standard datapalne services to be able to check for example that if you define a dataplane service of type=nova tha the secrete always have a migration ssh key defined.

it would also allow use to enforce that for any node set we only have one datapalneservice for any given type.

the value of type for the standard service would basically map to the last component of the playbook

https://github.com/openstack-k8s-operators/nova-operator/blob/main/api/v1beta1/nova_webhook.go#L74-L139

so type=nova is related to   playbook: osp.edpm.nova

we don't have to enforce that but if we are wondering what the type would be for the exiting service that's what i would use.

SeanMooney:
i just pushed this before i finish for the week. i have it running locally but the package installation time is very long for some reason so I'm just going to leave it run and let ci run on it over the weekend.

roles/edpm_nova/defaults/main.yml
olliewalsh: I don't think option 1 is correct. If we have 2 libvirt services (not sure what the use-case would be but hypothetically....) then they have independent config. If they happen to be using the same CA that's fine, but they should get individual keys/certs/secrets

olliewalsh: seems to work https://github.com/openstack-k8s-operators/dataplane-operator/commit/dda1a9d0e8d99d616db05a5ff9d9d242ef33f6bf with https://github.com/olliewalsh/edpm-ansible/commit/13b1158ae14ca5cbcd5e327c7d6833906820dc53

SeanMooney: the service name can solve this problem for this specific usecase so i updated the patch to support that.

i think have a type on the dataplane service so we can use that for validation or defaulting is still worth considering but that can be done separately.

@olliewalsh are you going to submit a pr to the dataplane operator for that.

olliewalsh: Included in the OVN dataplane PR https://github.com/openstack-k8s-operators/dataplane-operator/pull/745

SeanMooney:
check-rdo i added a depends on to https://github.com/openstack-k8s-operators/dataplane-operator/pull/754 to enable tls by default and https://github.com/openstack-k8s-operators/dataplane-operator/pull/745 to pass the edpm_service_name.

SeanMooney:
check-rdo

SeanMooney:
check-rdo this should not pass i belive as  podified-multinode-edpm-deployment-crc  now passes in https://github.com/openstack-k8s-operators/dataplane-operator/pull/754

olliewalsh (approved):
/lgtm

SeanMooney:
check-rdo

PR #591, opened 2024-03-05
Libvirt must not trust the CA bundle

olliewalsh:
The CA bundle includes all of the standard root CAs. Libvirt must only trust the specific CA.

Depends-On: https://github.com/openstack-k8s-operators/dataplane-operator/pull/731


gibizer (approved):
looks good

PR #593, opened 2024-03-07
Adoption: fix stopping nova services from tripleo

bogdando:
Match the target container by a strict name.
Filter unrelated nova_virtlogd_wrapper container
and make the nova_virtlogd stopping condition triggering.

Move the stopping logic for nova_virtlogd_wrapper and nova_virtlogd
into the libvirt role as we need to stop these *before* we start
deploying de-containerized libvirt. While the nova role is called
after the libvirt role, so stopping the tripleo libvirt related
happens too late (libvirt has already failed).

Also stop other Nova control plane services, in case of all-in-one
overcloud adoption.

JIRA [OSPRH-2302](https://issues.redhat.com/browse/OSPRH-2302)

### DIFF
--- a/roles/edpm_nova/defaults/main.yml
+++ b/roles/edpm_nova/defaults/main.yml
@@ -32,14 +32,14 @@ edpm_nova_compute_image: "quay.io/podified-antelope-centos9/openstack-nova-compu
 # certs
 edpm_nova_tls_certs_enabled: "{{ edpm_tls_certs_enabled | default(False) }}"
 
-# list of services to stop during adoption
-edpm_old_tripleo_compute_sevices:
+# list of tripleo nova services to stop during EDPM adoption
+edpm_nova_old_tripleo_compute_sevices:
   - tripleo_nova_compute.service
-  - tripleo_nova_libvirt.target
   - tripleo_nova_migration_target.service
-  - tripleo_nova_virtlogd_wrapper.service
-  - tripleo_nova_virtnodedevd.service
-  - tripleo_nova_virtproxyd.service
-  - tripleo_nova_virtqemud.service
-  - tripleo_nova_virtsecretd.service
-  - tripleo_nova_virtstoraged.service
+  # Other control plane services, in case of all-in-one overcloud adoption
+  - tripleo_nova_api_cron.service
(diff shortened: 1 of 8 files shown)


mrkisaolamb (approved):
lgtm

bogdando:
/hold Now I realised that this task should belong to libvirt
we deploy nova after libvirt, and virtlogd makes the latter failing, before the former can fix it

bogdando:
recheck unrelated

roles/edpm_nova/defaults/main.yml
```diff
@@ -32,14 +32,14 @@ edpm_nova_compute_image: "quay.io/podified-antelope-centos9/openstack-nova-compu
...
   - tripleo_nova_compute.service
-  - tripleo_nova_libvirt.target
   - tripleo_nova_migration_target.service
-  - tripleo_nova_virtlogd_wrapper.service
-  - tripleo_nova_virtnodedevd.service
-  - tripleo_nova_virtproxyd.service
-  - tripleo_nova_virtqemud.service
-  - tripleo_nova_virtsecretd.service
-  - tripleo_nova_virtstoraged.service
+  # Other control plane services, in case of all-in-one overcloud adoption
```
SeanMooney, line 39: hum ok yes if you were adopting form standalone this would be needed.

SeanMooney (commented):
recheck 

this looks good. 
it fixes the responsibilities between the roles
the main feedback is there is no molecule testing for the adoption behavior.

i know gibi was looking at adding an adoption seneario so perhaps we could cover that there but ideally we would use tasks in prepare to create a fake systems service to emulate trileo deployed services.

is this something you plan to do in this pr.

roles/edpm_nova/defaults/main.yml
bogdando: there is also all-in-one overcloud managed with Director supported in OSP (which is NOT standalone tripleo)

bogdando:
recheck edpm-ovn

mrkisaolamb:
recheck

SeanMooney (approved):
very well, lets follow up with the molecule tests then and proceed with this to unblock adoption work

PR #594, opened 2024-03-08
OVN TLS support

olliewalsh:
Update to latest edpm TLS logic.
Use the same key/crt names as control plane for consistency.

Trust the CA bundle. Make it world readable to allow it to be mounted directly.

Expects dataplane-operator to pass edpm_service_name if it's not the default.

Jira: [OSPRH-2191](https://issues.redhat.com//browse/OSPRH-2191)

### DIFF
--- a/roles/edpm_install_certs/tasks/copy_ca_certs.yaml
+++ b/roles/edpm_install_certs/tasks/copy_ca_certs.yaml
@@ -37,6 +37,6 @@
       ansible.builtin.copy:
         src: "{{ cacert_src_path }}/tls-ca-bundle.pem"
         dest: "{{ cacert_dest_path }}/tls-ca-bundle.pem"
-        mode: '0600'
+        mode: '0644'
         owner: root
         group: root
--- a/roles/edpm_ovn/defaults/main.yml
+++ b/roles/edpm_ovn/defaults/main.yml
@@ -7,7 +7,7 @@ edpm_ovn_images_download_delay: 5
 # number of retries for download tasks
 edpm_ovn_images_download_retries: 5
 
-edpm_ovn_config_src: /var/lib/openstack/configs/ovn
+edpm_ovn_config_src: "/var/lib/openstack/configs/{{ edpm_service_name | default('ovn') }}"
 edpm_ovn_bridge: br-int
 edpm_ovn_bridge_mappings: ["datacentre:br-ex"]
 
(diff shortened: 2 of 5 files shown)


roles/edpm_install_certs/tasks/copy_ca_certs.yaml
```diff
@@ -37,6 +37,6 @@
       ansible.builtin.copy:
         src: "{{ cacert_src_path }}/tls-ca-bundle.pem"
         dest: "{{ cacert_dest_path }}/tls-ca-bundle.pem"
-        mode: '0600'
+        mode: '0644'
```
SeanMooney, line 40: i assume this is required to allow non privadged applications to use the pem file to validate certs. if that is correct then this make sense to me to modify. its not mentioned in the commit message but that is what i would infer form context.

roles/edpm_ovn/defaults/main.yml
```diff
@@ -7,7 +7,7 @@ edpm_ovn_images_download_delay: 5
 # number of retries for download tasks
 edpm_ovn_images_download_retries: 5
 
-edpm_ovn_config_src: /var/lib/openstack/configs/ovn
+edpm_ovn_config_src: "/var/lib/openstack/configs/{{ edpm_service_name|default('ovn') }}"
```
SeanMooney, line 10: because you have provided a default this technically does not depend on the 
dataplane change to support edpm_service_name

minor nit i would write this as {{ edpm_service_name | default('ovn') }}

roles/edpm_ovn/defaults/main.yml
```diff
@@ -61,8 +61,12 @@ edpm_ovn_controller_common_volumes:
   - /var/lib/kolla/config_files/ovn_controller.json:/var/lib/kolla/config_files/config.json:ro
 
 edpm_ovn_controller_tls_volumes:
-  - /etc/pki/tls/certs/:/etc/pki/tls/certs/
-  - /etc/pki/tls/private/:/etc/pki/tls/private/
+  - "/var/lib/openstack/certs/{{ edpm_service_name|default('ovn') }}/ca.crt:/etc/pki/tls/certs/ovndbca.crt:ro,z"
+  - "/var/lib/openstack/certs/{{ edpm_service_name|default('ovn') }}/tls.crt:/etc/pki/tls/certs/ovndb.crt:ro,z"
+  - "/var/lib/openstack/certs/{{ edpm_service_name|default('ovn') }}/tls.key:/etc/pki/tls/private/ovndb.key:ro,Z"
+  - "/var/lib/openstack/cacerts/{{ edpm_service_name|default('ovn') }}/tls-ca-bundle.pem:/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem:ro,z"
```
SeanMooney, line 67: i assume this is requried for the webhook callback form ovn to the neuron server to enable it to send events to neutron.

SeanMooney (commented):
over all this looks reasonable to me. some clarifying question in line but in general this look good.

Ollie do you plan to provide molecule test coverage for this?
that would be required for GA but could be done as a separate QE task or a sepetate pr.

roles/edpm_install_certs/tasks/copy_ca_certs.yaml
olliewalsh: yes, in the PR comment "Trust the CA bundle. Make it world readable to allow it to be mounted directly."

roles/edpm_ovn/defaults/main.yml
olliewalsh: it's not required for anything specific, the cabundle just should be trusted everywhere

olliewalsh:
> over all this looks reasonable to me. some clarifying question in line but in general this look good.
> 
> Ollie do you plan to provide molecule test coverage for this? that would be required for GA but could be done as a separate QE task or a sepetate pr.

yes, planned to follow up with molecule tests but don't want to block this as I expect it could require some changes to get the nodes configured correctly for name resolution etc...

SeanMooney (approved):
based on olies responces i think we can proceed with this to start unwinding the stack of depend on we need to be able to test this properly.

PR #596, opened 2024-03-12
use canonical_hostname not inventory hostname for certs

SeanMooney:
the tls certs are indexed by the node.HostName
when generating the invetory the inventory_hostname is
computed via strings.Split(node.HostName, ".")[0]
node.HostName is stored in the canonical_hostname field
of the host vars.

This change updates edpm_install_certs to align
to the go code by using canonical_hostname instead.

Depends-On: https://github.com/openstack-k8s-operators/dataplane-operator/pull/778

### DIFF
--- a/roles/edpm_install_certs/tasks/copy_certs_and_keys.yaml
+++ b/roles/edpm_install_certs/tasks/copy_certs_and_keys.yaml
@@ -39,6 +39,8 @@
         - "{{ key_dest_path }}"
 
     - name: Copy cert and key to the compute node
+      vars:
+        basename: "{{ hostvars[inventory_hostname]['canonical_hostname'] }}"
       ansible.builtin.copy:
         src: "{{ item.src }}"
         dest: "{{ item.dest }}"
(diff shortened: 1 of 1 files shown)


SeanMooney:
/hold

this needs to be fixed in go

the canonical hostname is not the same as node.hostname
canonical_hostname appears to be the the control plane network fqdn.

where this breaks down is if you have a node in the nodeset that looks like this
```
    nodes:
      edpm-compute-0:
        ansible:
          ansibleHost: 192.168.122.100
          ansibleUser: ""
        hostName: edpm-compute-0
        networks:
        - defaultRoute: true
          fixedIP: 192.168.122.100
          name: CtlPlane
          subnetName: subnet1
        - name: InternalApi
          subnetName: subnet1
        - name: Storage
          subnetName: subnet1
        - name: Tenant
          subnetName: subnet1
```
which result in this in the inventory

```
   "hosts": {
      "edpm-compute-0": {
        "ansible_host": "192.168.122.100",
        "canonical_hostname": "edpm-compute-0.ctlplane.example.com",
        "ctlplane_cidr": 24,
        "ctlplane_dns_nameservers": [
          "192.168.122.80"
        ],

```

so before we can use the canonical_hostname or inventory hostname we need to change the key in the secrete to not be node.hostname directly as that does not map to either the inventory_hostname or canonical_hostname

vakwetu:
+2

SeanMooney:
since i needed to rerun the ci anyway i just did a rebase in the UI to get updated results. the code is unchanged form the last version

fao89 (approved):

PR #603, opened 2024-03-18
Support removing custom nova-configs

mrkisaolamb:
Fixes: [OSPRH-2369](https://issues.redhat.com/browse/OSPRH-2369)

### DIFF
--- a/roles/edpm_nova/molecule/default/prepare.yml
+++ b/roles/edpm_nova/molecule/default/prepare.yml
@@ -154,3 +154,22 @@
         path: /etc/hosts
         line: "{{ ansible_default_ipv4.address | default('127.0.0.1') }} {{ ansible_fqdn }} {{ inventory_hostname }}"
         state: present
+
+    - name: Create container config dirs for test
+      become: true
+      ansible.builtin.file:
+        path: "{{ item.path }}"
+        state: "directory"
+        setype: "container_file_t"
+        owner: "{{ item.owner | default(ansible_user) }}"
+        group: "{{ item.group | default(ansible_user) }}"
+        mode: "{{ item.mode | default(omit) }}"
+      loop:
+        - {"path": "/var/lib/openstack/config/nova", "mode": "0755"}
+
+    - name: Create a test file to be deleted
+      become: true
+      ansible.builtin.file:
+        path: '/var/lib/openstack/config/nova/test.conf'
+        state: touch
+        mode: 0644
--- a/roles/edpm_nova/tasks/configure.yml
+++ b/roles/edpm_nova/tasks/configure.yml
@@ -1,5 +1,16 @@
 ---
 
+- name: Create sync config dirs
+  tags:
+    - configure
+    - nova
+  ansible.builtin.file:
+    path: "{{ lookup('env', 'HOME') }}/staging"
+    state: "directory"
+    mode: "0766"
+  delegate_to: localhost
+  run_once: true
+
 - name: Create container config dirs
   tags:
     - configure
@@ -39,6 +50,52 @@
     - {"path": "/var/lib/iscsi"}
     - {"path": "/etc/nvme"}
     - {"path": "/run/openvswitch"}
+
+- name: Discover configmaps in {{ edpm_nova_config_src }}
+  ansible.builtin.find:
+    paths: "{{ edpm_nova_config_src }}"
+    file_type: file
+    recurse: true
+  register: edpm_nova_configmaps
+  changed_when: false
+  check_mode: false
+  delegate_to: localhost
+
+- name: Flatten configmaps
+  tags:
+    - configure
+    - nova
+  ansible.builtin.copy:
+    src: "{{ item.path }}"
+    dest: "{{ lookup('env', 'HOME') }}/staging/{{ item.path | basename }}"
+    mode: "0644"
+  loop: "{{ edpm_nova_configmaps.files }}"
+  delegate_to: localhost
+  run_once: true
+
+- name: Configure nova user and group on the host
+  ansible.builtin.import_role:
+    name: edpm_users
+  vars:
+    edpm_users_users:
+      # 42436 is matching with the uid and gid created by kolla in the nova containers
+      - {"name": "nova", "uid": "42436", "gid": "42436", "shell": "/bin/sh", "comment": "nova user", "groups": "libvirt"}
+    edpm_users_extra_dirs: []
+  tags:
+    - edpm_users
+
+- name: Sync nova config files
+  tags:
+    - configure
+    - nova
+  ansible.posix.synchronize:
+    src: "{{ lookup('env', 'HOME') }}/staging/"
+    dest: "{{ edpm_nova_config_dest }}"
+    recursive: true
+    delete: true
+  notify:
+    - Restart nova
+
 - name: Render nova config files
   tags:
     - configure
@@ -66,40 +123,6 @@
     - {"src": "02-nova-host-specific.conf.j2", "dest": "02-nova-host-specific.conf"}
   notify:
     - Restart nova
-
-- name: Discover configmaps in {{ edpm_nova_config_src }}
-  ansible.builtin.find:
-    paths: "{{ edpm_nova_config_src }}"
-    file_type: file
-    recurse: true
-  register: edpm_nova_configmaps
-  changed_when: false
-  check_mode: false
-  delegate_to: localhost
-
-- name: Flatten configmaps into {{ edpm_nova_config_dest }}
-  tags:
-    - configure
-    - nova
-  ansible.builtin.copy:
-    src: "{{ item.path }}"
-    dest: "{{ edpm_nova_config_dest }}/{{ item.path | basename }}"
-    setype: "container_file_t"
-    mode: "0644"
-  loop: "{{ edpm_nova_configmaps.files }}"
-  notify:
-    - Restart nova
-
-- name: Configure nova user and group on the host
-  ansible.builtin.import_role:
-    name: edpm_users
-  vars:
-    edpm_users_users:
-      # 42436 is matching with the uid and gid created by kolla in the nova containers
-      - {"name": "nova", "uid": "42436", "gid": "42436", "shell": "/bin/sh", "comment": "nova user", "groups": "libvirt"}
-    edpm_users_extra_dirs: []
-  tags:
-    - edpm_users
 
 - name: Create .ssh directory for the nova user on the host
   become: true
--- a/roles/edpm_nova/molecule/default/verify.yml
+++ b/roles/edpm_nova/molecule/default/verify.yml
@@ -120,3 +120,13 @@
           - "compute_id.stat.size == 37"
           - "compute_id.stat.mode == '0400'"
           - "compute_id.stat.attr_flags == 'i'"
+
+    - name: Check if after sync old file is removed
+      ansible.builtin.stat:
+        path: '/var/lib/openstack/config/nova/test.conf'
+      register: nova_host_specific_conf
+
+    - name: Assert that /var/lib/openstack/config/nova/test.conf file does not exist
+      ansible.builtin.assert:
+        that:
+          - "not nova_host_specific_conf.stat.exists"


roles/edpm_nova/defaults/main.yml
```diff
@@ -27,6 +27,7 @@ edpm_nova_image_download_retries: 5
 # dest dir is on the target host
 edpm_nova_config_src: /var/lib/openstack/configs
 edpm_nova_config_dest: /var/lib/openstack/config/nova
+edpm_nova_config_sync_dest: /tmp/nova_sync/
```
gibizer, line 30: I guess we need an argspec for this as the rest has it.

gibizer: Alternatively we could use ansible.builtin.tempfile module

roles/edpm_nova/tasks/configure.yml
```diff
@@ -101,6 +102,32 @@
...
+    - nova
+  ansible.posix.synchronize:
+    src: "{{ edpm_nova_config_sync_dest }}"
+    dest: "{{ edpm_nova_config_dest }}"
+    recursive: true
+    delete: true
+    mode: "push"
+    owner: true
+    perms: true
+  delegate_to: "{{ inventory_hostname }}"
```
gibizer, line 118: OK, we need this to run rsync between two directories on the EDPM node as by default rsync runs in the aee pod.

roles/edpm_nova/tasks/configure.yml
```diff
@@ -101,6 +102,32 @@
...
+- name: Sync nova config files
+  become: true
+  tags:
+    - configure
+    - nova
+  ansible.posix.synchronize:
+    src: "{{ edpm_nova_config_sync_dest }}"
+    dest: "{{ edpm_nova_config_dest }}"
+    recursive: true
+    delete: true
```
gibizer, line 114: OK, this is the meat of the patch, we want to make sure that the files that are not provided by the user are deleted

roles/edpm_nova/molecule/default/prepare.yml
```diff
@@ -154,3 +154,12 @@
         path: /etc/hosts
         line: "{{ ansible_default_ipv4.address | default('127.0.0.1') }} {{ ansible_fqdn }} {{ inventory_hostname }}"
         state: present
+
+    - name: Create config_file_for_delet file
```
gibizer, line 158: but you just create a file named 'test'. So I would simply call this: Create a test file to be deleted

mhuin:
recheck

morucci:
recheck

roles/edpm_nova/tasks/configure.yml
SeanMooney: ya so synchonize uses rsync and adding delete with recusitive is basically like passing mirror

with this current config it will copy all files form the source that are newer then on the test and delete any files on the dest not found on the source.

roles/edpm_nova/tasks/configure.yml
SeanMooney: we only need to do this because we need to flatten multiple direcoties into one.

we could technically flatten them in the AEE container and then rsync the result instead.
that might make more sense.

morucci:
recheck

roles/edpm_nova/tasks/configure.yml
```diff
@@ -77,13 +78,13 @@
   check_mode: false
   delegate_to: localhost
 
-- name: Flatten configmaps into {{ edpm_nova_config_dest }}
+- name: Flatten configmaps into {{ edpm_nova_config_sync_dest }}
```
SeanMooney, line 81: so we could create the flattend config map in the AEE container by delegation this to the aee host and then rsyncing the result after we have merged the files.

that would avoid creating the tempory folder on the dest host it would also have the advantage of only doing the flattening once. rather then doing it for every host.

roles/edpm_nova/tasks/configure.yml
bogdando: I agree, this would be a better path to take

roles/edpm_nova/defaults/main.yml
mrkisaolamb: done

roles/edpm_nova/tasks/configure.yml
mrkisaolamb: done, now rsync is between ansible container and edpm_node

roles/edpm_nova/molecule/default/prepare.yml
mrkisaolamb: done

roles/edpm_nova/tasks/configure.yml
mrkisaolamb: done

roles/edpm_nova/tasks/configure.yml
```diff
@@ -77,18 +62,18 @@
   check_mode: false
   delegate_to: localhost
 
-- name: Flatten configmaps into {{ edpm_nova_config_dest }}
+- name: Flatten configmaps into {{ edpm_nova_config_sync_src }}
+  become: true
```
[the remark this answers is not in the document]
bogdando: see also https://ansible.readthedocs.io/projects/lint/rules/no-same-owner/#problematic-code

perhaps we need to use 
```
owner: false
group: false
```
to not change nova for root?

bogdando (changes requested):
not sure about ownership changes for files during flattening and rsyncing around

roles/edpm_nova/vars/main.yml
```diff
@@ -20,3 +20,6 @@
 # intended to be modified.
 
 # All variables within this role should have a prefix of "edpm_nova"
+
+# this is const path where nova custom configs will land
+edpm_nova_config_sync_src: /var/lib/openstack/configs/custom/nova/
```
bogdando, line 25: please use configs/nova, we don't have sosreports coverage for custom/

https://github.com/sosreport/sos/pull/3414/files#diff-5cbe3417b4fe3c187b14e99f4897ccb987ecd6c3bd2b904266b8b932cef22e21R221

roles/edpm_nova/tasks/configure.yml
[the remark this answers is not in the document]
mrkisaolamb: done

roles/edpm_nova/tasks/configure.yml
```diff
@@ -101,6 +85,51 @@
...
+    recursive: true
+    delete: true
+    mode: "push"
+    owner: false
+    group: false
+  notify:
+    - Restart nova
+
+- name: Render nova config files
+  become: true
```
bogdando, line 105: as we sudo it, let;s specify the owner and user?

mrkisaolamb: this should be ok I think

mrkisaolamb:
recheck

bogdando:
looks good to me

mrkisaolamb:
recheck

roles/edpm_nova/molecule/vagrant/molecule.yml
```diff
@@ -5,14 +5,16 @@ driver:
   name: vagrant
   provider:
     name: libvirt
+    type: libvirt
```
SeanMooney, line 8: i think this will set the default but i wold prefer to keep this isn the platform section is there a reason you moved this?

if you have multiple instance and you want to have different value then its prefer to define it in the platform section

roles/edpm_nova/tasks/configure.yml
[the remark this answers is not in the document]
SeanMooney: i mind using become: true less of an issue in that AEE although were not required yes i prefer to avoid it.

i belive you require become here because

/var/lib/openstack/configs/custom/nova/ is not owned by the use in the AEE container  perhaps we should use something like '{{ lookup('ansible.builtin.env', 'HOME') }}/staging'

that would allow you to drop become here.

roles/edpm_nova/tasks/configure.yml
```diff
@@ -101,6 +86,50 @@
...
+  tags:
+    - configure
+    - nova
+  ansible.posix.synchronize:
+    src: "{{ edpm_nova_config_sync_src }}"
+    dest: "{{ edpm_nova_config_dest }}"
+    recursive: true
+    delete: true
+    mode: "push"
```
SeanMooney, line 99: i belive push is the default by the way but that fine.

roles/edpm_nova/tasks/configure.yml
```diff
@@ -101,6 +86,50 @@
   tags:
     - edpm_users
 
+- name: Sync nova config files
+  become: true
```
SeanMooney, line 90: this should not need become as the directory is owned by the ansible user.

roles/edpm_nova/tasks/configure.yml
```diff
@@ -101,6 +86,50 @@
...
+  tags:
+    - configure
+    - nova
+  ansible.posix.synchronize:
+    src: "{{ edpm_nova_config_sync_src }}"
+    dest: "{{ edpm_nova_config_dest }}"
+    recursive: true
+    delete: true
+    mode: "push"
+    owner: true
```
SeanMooney, line 100: we do not want to preserve the owner of the files we copy.
they should be owned by the ansible user

SeanMooney (changes requested):

roles/edpm_nova/molecule/vagrant/molecule.yml
mrkisaolamb: done, it was unintentional change

roles/edpm_nova/tasks/configure.yml
mrkisaolamb: yes you are right, done

roles/edpm_nova/tasks/configure.yml
mrkisaolamb: thanks, done

roles/edpm_nova/molecule/default/prepare.yml
```diff
@@ -154,3 +154,22 @@
         path: /etc/hosts
         line: "{{ ansible_default_ipv4.address | default('127.0.0.1') }} {{ ansible_fqdn }} {{ inventory_hostname }}"
         state: present
+
+    - name: Create container config dirs
```
SeanMooney, line 158: oh ok your doing that to create the test file.
to ensure its deleted

roles/edpm_nova/tasks/configure.yml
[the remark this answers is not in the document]
SeanMooney: so this become true shoudl not be needed now right

"{{ lookup('ansible.builtin.env', 'HOME') }}/staging/ is owned by the current user

roles/edpm_nova/tasks/configure.yml
```diff
@@ -1,5 +1,17 @@
...
 
+- name: Create sync config dirs
+  tags:
+    - configure
+    - nova
+  become: true
+  ansible.builtin.file:
+    path: "{{ lookup('ansible.builtin.env', 'HOME') }}/staging"
+    state: "directory"
+    mode: "0755"
```
SeanMooney, line 11: mode shoudl eirther by 766 or you should set the owner to the current user

but since this is in your home directory you should be able to just remove become:true instaed

SeanMooney (changes requested):
the permissions are still not quite correct but over all the code looks ok

roles/edpm_nova/tasks/configure.yml
mrkisaolamb: thanks, done!

mrkisaolamb:
recheck

mrkisaolamb:
recheck

mrkisaolamb:
recheck

mrkisaolamb:
recheck

mrkisaolamb:
recheck

mrkisaolamb:
recheck

mrkisaolamb:
recheck

mrkisaolamb:
recheck

mrkisaolamb:
Ci should pass after we land https://github.com/openstack-k8s-operators/edpm-ansible/pull/616 and :latest img will contain this change

SeanMooney (approved):
this looks good to me however the zuul jobs still failed
because https://github.com/openstack-k8s-operators/edpm-ansible/pull/616 is not merged yet

TASK [osp.edpm.edpm_nova : Sync nova config files] *****************************
[1;30mtask path: /usr/share/ansible/collections/ansible_collections/osp/edpm/roles/edpm_nova/tasks/configure.yml:87[0m
[0;31mfatal: [compute-0]: FAILED! => {"changed": false, "msg": "Failed to find required executable \"rsync\" in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}[0m

so we can use check-rdo to rerun those once it 616 has merged

roles/edpm_nova/vars/main.yml
```diff
@@ -20,3 +20,5 @@
 # intended to be modified.
 
 # All variables within this role should have a prefix of "edpm_nova"
+
+# this is const path where nova custom configs will land
```
gibizer, line 24: either a variable is missing below or this comment is not needed

roles/edpm_nova/tasks/configure.yml
```diff
@@ -77,18 +61,17 @@
   check_mode: false
   delegate_to: localhost
 
-- name: Flatten configmaps into {{ edpm_nova_config_dest }}
+- name: Flatten configmaps into {{ lookup('env', 'HOME') }}
```
gibizer, line 64: /staging is missing so the name will be missleading

gibizer (commented):
I have two nits

gibizer:
Could you please clean up the commits in the PR. The second one "Install rsync in container for ansible.posix.synchronize" is actually describing something that landed separately and containing changes that are belonging to the first commit.

mrkisaolamb:
recheck

gibizer (approved):
looks good

roles/edpm_nova/tasks/configure.yml
```diff
@@ -101,6 +84,46 @@
...
+  ansible.builtin.template:
+    src: "{{ item.src }}"
+    dest: "{{ edpm_nova_config_dest }}/{{ item.dest }}"
+    setype: "container_file_t"
+    mode: "0644"
+  loop:
+    - {"src": "config.json.j2", "dest": "config.json"}
+    - {"src": "nova-blank.conf", "dest": "nova-blank.conf"}
+    - {"src": "ssh-config", "dest": "ssh-config"}
+    # NOTE(gibi): This is unfortunate as we would like to avoid config
```
SeanMooney, line 112: nit we now also use this to enable tls.
it still does not change any of the sentiment that its unfortunate that this file need to exist and we would like to remove it in the future.
i would hope that eventually we do not need this for tls but we need to work that out

for now I'm fine with not updating the comment to mention that is used for enabling tls-e especially since it already references tls certificates.
that out of scope of this pr anyway as you are not changing this in this pr

SeanMooney (approved):
i agree with gibi this now looks good to me.
we are not eleveating privladges where we should not be
we have optimised the code to delgate the file creation to the AEE pod which will speed up deployment and there is test coverage so +2 all round

PR #587, opened 2024-03-04
Use FQDN as sevice hostname

gibizer:
The greenfield deployment sets up host FQDN via IPAM but did not change
the /etc/hostname to FQDN. This causes that libvirt will see and use
FQDN for the node names, but nova-compute uses the sort hostname
returned by socket.gethostname().

In greenfield it causes that the nova-compute host name is different
form the compute nodename. Nova works with this setup, but it will cause
confusion for the users coming from older versions where both names are
FQDN.

During adoption, when the adopted cloud does not have FQDN in the
/etc/hostname, the compute service host name will change as in 18 we are
not yet hardcoding the FQDN in the nova.conf. This results in failed
adoption.

This PR ensures the [DEFAULT]host is set to the FQDN based on the
canonical_hostname variable for all the EDPM openstack services.

This patch does not handle the adoption case where:
* the old cloud had something other than the FQDN set in the service
  host configuration
* the old cloud had a different FQDN set than what is set in the
canonical_host in the new cloud configuration

There will be a separate PR to solve the adoption case via a new pre
adoption validation role.

Fixes: [OSPRH-5197](https://issues.redhat.com//browse/OSPRH-5197)

PR #604, opened 2024-03-19
Clean edpm_neutron_metadata role from unused variable

slawqo:
Variable 'edpm_neutron_metadata_agent_DEFAULT_host' defined in the edpm_neutron_metadata role wasn't actually used anywhere in the role except argument_specs.yml meta file where it was used as default value for the 'edpm_neutron_metadata_agent_DEFAULT_nova_metadata_host' variable (which was also not correct).
Variable 'edpm_neutron_metadata_agent_DEFAULT_host' is actually not needed anywhere as neutron_ovn_metadata_agent reports hostname on which it is run using hostname defined in the "ovn chassis" (and this is configured by the edpm_ovn role in external_ids in ovsdb).

Because of all of that mentioned above, this small patch removes that not used variable.

### DIFF
--- a/roles/edpm_neutron_dhcp/defaults/main.yml
+++ b/roles/edpm_neutron_dhcp/defaults/main.yml
@@ -71,7 +71,6 @@ edpm_neutron_dhcp_rootwrap_DEFAULT_rlimit_nofile: 1024
 # neutron-dhcp-agent.conf
 # DEFAULT
 edpm_neutron_dhcp_agent_DEFAULT_state_path: '/var/lib/neutron'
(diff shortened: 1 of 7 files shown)


karelyatin (commented):
/lgtm

gibizer (commented):
looks good to me but let the networking folks to approve it.

All these services will get specific service host configuration in #587 anyhow.

karelyatin:
recheck

slawqo:
recheck

SeanMooney (approved):
yep this makes sense

PR #605, opened 2024-03-19
Force molecule colors

fao89:


SeanMooney (approved):
looking at the the prow job the colourisation works as expected

https://github.com/openstack-k8s-operators/edpm-ansible/actions/runs/8347068956/job/22845868709?pr=605#step:7:4044

and looking at the zuul jobs they are unaffected by this change which is expected as this is only modifying the prow jobs so there has been no regerssion. we could explore also making this change in the ci framework but that is out of scope of this pr

based on that I'm approving this change.

PR #609, opened 2024-03-20
[adoption]Add a new role for pre adoption validation 

gibizer:
The new edpm_pre_adoption_validation role is intended to be used as
dataplane service in a separate deployment during adoption that runs
before the main deployment of the adoption process. This role does not
change anything on the EDPM nodes just reads their current (pre-adoption)
configuration and check if the intended new (post-adoption)
configuration will not cause any trouble.

The first check introduced by this commit ensures that the openstack
service hostname will not change during adoption on the EDPM nodes. Such
change would cause adoption failure in a later stage, after the point of
no return, as nova-compute will refuse to start up if it detects the
host rename. Additionally neutron services might fail too / or create
database corruption due to the host rename.

Implements: [OSPRH-5713](https://issues.redhat.com//browse/OSPRH-5713)

### DIFF
--- a/roles/edpm_pre_adoption_validation/molecule/adoption-negative/converge.yml
+++ b/roles/edpm_pre_adoption_validation/molecule/adoption-negative/converge.yml
@@ -1,0 +1,19 @@
+---
+- name: Converge
+  hosts: all
+  tasks:
+    - name: Converge
+      block:
+        - name: "Include role"
+          ansible.builtin.include_role:
+            name: osp.edpm.edpm_pre_adoption_validation
+
+        - name: "Check execution halted"
+          ansible.builtin.fail:
+            msg: "Execution should stop before this task"
+          register: should_not_run
+      rescue:
+        - name: Asset that role failed
+          ansible.builtin.assert:
+            that:
+              - should_not_run is not defined
--- a/roles/edpm_pre_adoption_validation/molecule/adoption-negative/molecule.yml
+++ b/roles/edpm_pre_adoption_validation/molecule/adoption-negative/molecule.yml
@@ -1,0 +1,31 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: podman
+platforms:
+- name: edpm-0
+  command: /sbin/init
+  dockerfile: ../../../../molecule/common/Containerfile.j2
+  image: ${EDPM_ANSIBLE_MOLECULE_IMAGE:-"ubi9/ubi-init"}
+  registry:
+    url: ${EDPM_ANSIBLE_MOLECULE_REGISTRY:-"registry.access.redhat.com"}
+  ulimits:
+  - host
+provisioner:
+  name: ansible
+  inventory:
+    hosts:
+      all:
+        hosts:
+          edpm-0:
+            canonical_hostname: edpm-0.localdomain
+
+verifier:
+  name: ansible
+scenario:
+  test_sequence:
+     - destroy
+     - create
+     - prepare
+     - converge
--- a/roles/edpm_pre_adoption_validation/molecule/greenfield/converge.yml
+++ b/roles/edpm_pre_adoption_validation/molecule/greenfield/converge.yml
@@ -1,0 +1,13 @@
+---
+- name: Converge
+  hosts: all
+  tasks:
+    - name: Converge
+      block:
+        - name: "Include role"
+          ansible.builtin.include_role:
+            name: "osp.edpm.edpm_pre_adoption_validation"
+      rescue:
+        - name: Assert that validation is skipped
+          ansible.builtin.fail:
+            msg: "validation should be skipped on greenfield as no old config files exist"
--- a/roles/edpm_pre_adoption_validation/tasks/main.yml
+++ b/roles/edpm_pre_adoption_validation/tasks/main.yml
@@ -1,0 +1,68 @@
+---
+# Copyright 2023 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Validate that hostname for the EDPM openstack services do not change
+  become: true
+  tags:
+    - adoption
+    - validation
+  block:
+    - name: Check if old nova.conf exists
+      ansible.builtin.stat:
+        path: "{{ edpm_pre_adoption_validation_old_nova_config }}"
+      register: nova_old_config_file
+
+    # NOTE(gibi): we intentionally using python configparser to extract
+    # the config value. Grep is not enough as multiple sections have 'host'
+    # key. Also we cannot rely on crudini as it is not available on the
+    # node being adopted.
+    - name: Compare old nova service host to {{ canonical_hostname }}
+      when: nova_old_config_file.stat.exists
+      ansible.builtin.command:
+        cmd: >
+          python3 -c
+          "import configparser as c;
+          p = c.ConfigParser(strict=False);
+          p.read('{{ edpm_pre_adoption_validation_old_nova_config }}');
+          print(p['DEFAULT']['host'])"
+      register: nova_old_config_output
+      changed_when: false
+      failed_when: nova_old_config_output.stdout != canonical_hostname
+
+    - name: Check if old neutron.conf exists
+      ansible.builtin.stat:
+        path: "{{ edpm_pre_adoption_validation_old_neutron_config }}"
+      register: neutron_old_config_file
+
+    - name: Compare old neutron service host to {{ canonical_hostname }}
+      when: neutron_old_config_file.stat.exists
+      ansible.builtin.command:
+        cmd: >
+          python3 -c
+          "import configparser as c;
+          p = c.ConfigParser(strict=False);
+          p.read('{{ edpm_pre_adoption_validation_old_neutron_config }}');
+          print(p['DEFAULT']['host'])"
+      register: neutron_old_config_output
+      changed_when: false
+      failed_when: neutron_old_config_output.stdout != canonical_hostname
+
+    - name: Print validated information
+      ansible.builtin.debug:
+        msg:
+          - "canonical_hostname: {{ canonical_hostname }}"
+          - "old nova service host config: {{ nova_old_config_output.stdout | default('not found')}}"
+          - "old neutron service host config: {{ neutron_old_config_output.stdout | default('not found')}}"
--- a/.github/workflows/molecule.yaml
+++ b/.github/workflows/molecule.yaml
@@ -45,6 +45,7 @@ jobs:
         - edpm_telemetry
         - edpm_users
         - env_data
+        - edpm_pre_adoption_validation
 
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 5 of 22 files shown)


gibizer:
The usage of the new role can be seen in https://github.com/openstack-k8s-operators/data-plane-adoption/pull/374

gibizer:
recheck

SeanMooney:
check-rdo content provider failed

roles/edpm_pre_adoption_validation/tasks/main.yml
```diff
@@ -0,0 +1,56 @@
...
+      register: nova_old_config_file
+
+    - name: Compare old nova service host to {{ canonical_hostname }}
+      when: nova_old_config_file.stat.exists
+      ansible.builtin.shell: |
+        set -o pipefail
+        egrep -e '^host.*=' {{ edpm_pre_adoption_validation_old_nova_config }} | cut -d '=' -f2 | sed 's/ //g'
+      register: nova_old_config_output
+      changed_when: true
+      failed_when: nova_old_config_output.stdout != canonical_hostname
```
SeanMooney, line 35: nit this shoudl be changed_when: false
https://ansible.readthedocs.io/projects/lint/rules/no-changed-when/

changed_when: true implies we modifed something as a side effect fo the shell command.

failed_when: nova_old_config_output.stdout != canonical_hostname

is a nice way to do that test +1

roles/edpm_pre_adoption_validation/tasks/main.yml
```diff
@@ -0,0 +1,56 @@
...
+      register: neutron_old_config_output
+      changed_when: true
+      failed_when: neutron_old_config_output.stdout != canonical_hostname
+
+    - name: Print validated information
+      ansible.builtin.debug:
+        msg:
+          - "canonical_hostname: {{ canonical_hostname }}"
+          - "old nova service host config: {{ nova_old_config_output.stdout | default('not found')}}"
+          - "old neutron service host config: {{ neutron_old_config_output.stdout | default('not found')}}"
```
SeanMooney, line 56: ok i guess this works
im slightly inclined to say  we could take the config lookup, assert and printing and execat them into a separate task file

then we could include that in a loop but since we are only checking two files I'm fine with keeping this simple and duplicating the check

if we were to do this for  more files i would like to exctart this into reuabel code but for now I'm ok with the approach you have taken.

roles/edpm_pre_adoption_validation/molecule/greenfield/converge.yml
```diff
@@ -0,0 +1,13 @@
...
+  tasks:
+    - name: Converge
+      block:
+        - name: "Include role"
+          ansible.builtin.include_role:
+            name: "osp.edpm.edpm_pre_adoption_validation"
+      rescue:
+        - name: Assert that validation is skipped
+          ansible.builtin.fail:
+            msg: "validation should be skipped on greenfield as no old config files exist"
```
SeanMooney, line 13: +1
yes this is easy to overlook

roles/edpm_pre_adoption_validation/molecule/adoption-negative/converge.yml
```diff
@@ -0,0 +1,19 @@
...
+
+        - name: "Check execution halted"
+          ansible.builtin.fail:
+            msg: "Execution should stop before this task"
+          register: should_not_run
+      rescue:
+        - name: Asset that role failed
+          ansible.builtin.assert:
+            that:
+              - should_not_run is not defined
```
SeanMooney, line 19: ah yes so the way this si working is 

- name: "Check execution halted" should not run if the validation failed

so the rescue block assert that should_not_run is not defied to asset that the validation failed.

its a little indricet but i don't know of a better way to assert that a role failed while also not causing the playbook to fail.

roles/edpm_pre_adoption_validation/molecule/adoption-negative/molecule.yml
```diff
@@ -0,0 +1,32 @@
...
+  ulimits:
+  - host
+provisioner:
+  name: ansible
+  inventory:
+    hosts:
+      all:
+        hosts:
+          edpm-0:
+            canonical_hostname: edpm-0.localdomain
```
SeanMooney, line 23: the test files have oldname.localdomain and here we are setting it to edpm-0.localdomain

so this is what forces the failure.

roles/edpm_pre_adoption_validation/molecule/adoption-negative/molecule.yml
```diff
@@ -0,0 +1,32 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: podman
+platforms:
+- command: /sbin/init
```
SeanMooney, line 7: nit: the convention for molecule is to use - name as the first key in each object in the list. rather then sorting the keys alphabetically.

it does not actually matter it just make multi instance test simpler to read

```
platforms:
  - name: edpm-0
    image: ...
    ...
```

roles/edpm_pre_adoption_validation/molecule/adoption-negative/molecule.yml
```diff
@@ -0,0 +1,32 @@
...
+  dockerfile: ../../../../molecule/common/Containerfile.j2
+  image: ${EDPM_ANSIBLE_MOLECULE_IMAGE:-"ubi9/ubi-init"}
+  name: edpm-0
+  privileged: true
```
SeanMooney, line 11: nit i also dont think this need to be privileged but i could be wrong

SeanMooney (approved):
so i have a couple of meta comments.

gibi will be on PTO so will not be able to respin this until April

with that in mind while i have a few nit inline i don't think any of them warrant holding this but i would like to see what others think.

i have set hold on this temporarily to let other review on Monday and if there are no negative feedback early next week ill let this progress as is and we can  address the nits in a follow-up pr.

SeanMooney:
the content provider failed again but im not going to recheck it until monday

bogdando:
recheck

bogdando:
lgtm

jistr:
lgtm

roles/edpm_pre_adoption_validation/tasks/main.yml
gibizer: yeah, I thought about a loop. It is sad that ansible cannot directly loop on a block.

roles/edpm_pre_adoption_validation/molecule/adoption-negative/converge.yml
gibizer: I agree. I tried but failed to found a more direct pattern.

roles/edpm_pre_adoption_validation/molecule/adoption-negative/molecule.yml
gibizer: yes

roles/edpm_pre_adoption_validation/tasks/main.yml
gibizer: fixed the changed_when clause here and below

roles/edpm_pre_adoption_validation/molecule/adoption-negative/molecule.yml
gibizer: moved the name to the top here and in the other molecule.yml files

roles/edpm_pre_adoption_validation/molecule/adoption-negative/molecule.yml
gibizer: removed privilege here and in the other molecule.yml files

roles/edpm_pre_adoption_validation/tasks/main.yml
```diff
@@ -0,0 +1,56 @@
...
+    - name: Check if old nova.conf exists
+      ansible.builtin.stat:
+        path: "{{ edpm_pre_adoption_validation_old_nova_config }}"
+      register: nova_old_config_file
+
+    - name: Compare old nova service host to {{ canonical_hostname }}
+      when: nova_old_config_file.stat.exists
+      ansible.builtin.shell: |
+        set -o pipefail
+        egrep -e '^host.*=' {{ edpm_pre_adoption_validation_old_nova_config }} | cut -d '=' -f2 | sed 's/ //g'
```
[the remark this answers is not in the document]
gibizer: Done.

gibizer:
recheck
```
ERROR! Error when getting available collection versions for ansible.utils from default (https://galaxy.ansible.com/api/) (HTTP Code: 504, Message: Gateway Time-out Code: Unknown)
```

roles/edpm_pre_adoption_validation/tasks/main.yml
```diff
@@ -0,0 +1,56 @@
...
+    - name: Check if old nova.conf exists
+      ansible.builtin.stat:
+        path: "{{ edpm_pre_adoption_validation_old_nova_config }}"
+      register: nova_old_config_file
+
+    - name: Compare old nova service host to {{ canonical_hostname }}
+      when: nova_old_config_file.stat.exists
+      ansible.builtin.shell: |
+        set -o pipefail
+        crudini --get {{ edpm_pre_adoption_validation_old_nova_config }} DEFAULT host
```
olliewalsh, line 32: nit: `ansible.builtin.command` would be sufficient

gibizer:
recheck retry limit

roles/edpm_pre_adoption_validation/tasks/main.yml
gibizer, line 32: testing showed that we cannot rely on crudini on 17 either as it is not installed on the hypervisor just in the service containers (I was misled by the fact that edpm computes has it on the hypervisor). So grep is out, crudini is out, I hate but I will do this with a python configparser:
```
[root@standalone ~]# python3 -c "import configparser as c; p = c.ConfigParser(); p.read('/var/lib/config-data/puppet-generated/neutron/etc/neutron/neutron.conf'); print(p['DEFAULT']['host'])"
standalone.localdomain
```

gibizer: done

gibizer:
This worked in the adoption job finally https://logserver.rdoproject.org/74/374/aaca4659383d76fa31b0b217f71dcc9fa4d7e79f/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/3fee568/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/pre-adoption-validation-openstack-pre-adoption-openstack-rfbxh/logs/openstackansibleee.log

gibizer:
recheck

gibizer:
/unhold
We saw this working in the adoption job so this is good to go

fao89 (approved):

gibizer:
recheck
```
2024-04-11 07:36:59.590514 | controller | TASK [set_openstack_containers : Get operator environment variable after update output_dir={{ cifmw_set_openstack_containers_basedir }}/artifacts, script=oc set env {{ operator_pod_name }} -n {{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }} --list > operator_env.txt, chdir={{ cifmw_set_openstack_containers_basedir }}/artifacts, creates=operator_env.txt] ***
2024-04-11 07:36:59.590526 | controller | Thursday 11 April 2024  07:36:59 -0400 (0:00:00.030)       0:11:56.349 ********
2024-04-11 07:36:59.927582 | controller | Follow script's output here: /home/zuul/ci-framework-data/logs/ci_script_010_get_operator_environment.log
2024-04-11 07:36:59.927635 | controller | An exception occurred during task execution. To see the full traceback, use -vvv. The error was: NoneType: None
2024-04-11 07:36:59.927643 | controller | fatal: [localhost]: FAILED! => {"changed": true, "msg": "non-zero return code", "rc": 1, "stderr": "", "stderr_lines": [], "stdout": "~/ci-framework-data/artifacts ~/ci-framework-data/artifacts\nError from server (NotFound): pods \"openstack-ansibleee-operator-controller-manager-77b56db5b-cd647\" not found\n", "stdout_lines": ["~/ci-framework-data/artifacts ~/ci-framework-data/artifacts", "Error from server (NotFound): pods \"openstack-ansibleee-operator-controller-manager-77b56db5b-cd647\" not found"]}
```

PR #612, opened 2024-04-03
Fix qemu cert path and permissions

olliewalsh:
Qemu is expecting the ca cert to be in a single dir and root:qemu 0640

https://docs.openstack.org/nova/latest/admin/secure-live-migration-with-qemu-native-tls.html#other-tls-environment-related-checks-on-compute-nodes

Client cert is not valid right now so disable mTLS until this is fixed.

Also remove unnecessary edpm_libvirt_tls_ca_src_dir var.

### DIFF
--- a/roles/edpm_libvirt/tasks/configure.yml
+++ b/roles/edpm_libvirt/tasks/configure.yml
@@ -17,7 +17,8 @@
     - {"path": "/var/lib/edpm-config/firewall", "owner": "root", "group": "root"}
     - {"path": "/etc/pki/libvirt", "owner": "root", "group": "root"}
     - {"path": "/etc/pki/libvirt/private", "owner": "root", "group": "root"}
-    - {"path": "/etc/pki/CA/libvirt", "owner": "root", "group": "root"}
+    - {"path": "/etc/pki/CA", "owner": "root", "group": "root"}
+    - {"path": "/etc/pki/qemu", "owner": "root", "group": "qemu"}
 
 - name: Render libvirt config files
   tags:
@@ -109,7 +110,7 @@
     persistent: true
     state: true
 
-- name: Move TLS files to the right location on the compute node
+- name: Move libvirt TLS files to the right location on the compute node
   tags:
     - configure
     - libvirt
@@ -119,7 +120,7 @@
     - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.key", "dest": "/etc/pki/libvirt/private/serverkey.pem"}
     - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.crt", "dest": "/etc/pki/libvirt/clientcert.pem"}
     - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.key", "dest": "/etc/pki/libvirt/private/clientkey.pem"}
-    - {"src": "{{ edpm_libvirt_tls_ca_src_dir }}/ca.crt", "dest": "/etc/pki/CA/cacert.pem"}
+    - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/ca.crt", "dest": "/etc/pki/CA/cacert.pem"}
   ansible.builtin.copy:
     src: "{{ item.src }}"
     dest: "{{ item.dest }}"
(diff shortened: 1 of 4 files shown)


olliewalsh:
Hold until https://github.com/openstack-k8s-operators/openstack-operator/pull/727 tempest succeeds

roles/edpm_libvirt/tasks/configure.yml
```diff
@@ -17,7 +17,7 @@
     - {"path": "/var/lib/edpm-config/firewall", "owner": "root", "group": "root"}
     - {"path": "/etc/pki/libvirt", "owner": "root", "group": "root"}
     - {"path": "/etc/pki/libvirt/private", "owner": "root", "group": "root"}
-    - {"path": "/etc/pki/CA/libvirt", "owner": "root", "group": "root"}
+    - {"path": "/etc/pki/qemu", "owner": "root", "group": "{{ edpm_nova_libvirt_qemu_group | default('qemu') }}"}
```
olliewalsh, line 20: @gibizer is this correct? Not sure if it should just be hard-coded to qemu

roles/edpm_libvirt/templates/qemu.conf.j2
```diff
@@ -2,7 +2,6 @@ max_files = 32768
 max_processes = 131072
 vnc_tls = 0
 vnc_tls_x509_verify = 0
-default_tls_x509_verify = 1
 nbd_tls = 0
```
olliewalsh, line 5: is this needed for adoption from computes with NBD TLS enabled?

roles/edpm_libvirt/templates/qemu.conf.j2
```diff
@@ -2,7 +2,6 @@ max_files = 32768
 max_processes = 131072
 vnc_tls = 0
```
olliewalsh, line 3: Do we have a jira for this?

roles/edpm_libvirt/tasks/configure.yml
gibizer: Based on the documentaton edpm_nova_libvirt_qemu_group (The vhost user socket directory group name.) the intention of this variable is different. So I would just hardcode qemu here.

roles/edpm_libvirt/tasks/configure.yml
```diff
@@ -128,3 +129,23 @@
...
+    - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.crt", "dest": "/etc/pki/qemu/client-cert.pem"}
+    - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.key", "dest": "/etc/pki/qemu/client-key.pem"}
+    - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/ca.crt", "dest": "/etc/pki/qemu/ca-cert.pem"}
+  ansible.builtin.copy:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    remote_src: true
+    mode: "0640"
+    owner: "root"
+    group: "{{ edpm_nova_libvirt_qemu_group | default('qemu') }}"
```
gibizer, line 150: ditto I would hardcode qemu here

roles/edpm_libvirt/templates/qemu.conf.j2
gibizer: I don't think we have. Could you file one please?

roles/edpm_libvirt/templates/qemu.conf.j2
gibizer: Based on https://bugzilla.redhat.com/show_bug.cgi?id=1879477 it is enabled by default since libvirt 6.10 so I think we can remove it from here. I don't think it effects adoption as I expect that this is enabled there as well.

gibizer: It is also added conditionally below. I think it is OK to remove it from both places as the default is the same.

olliewalsh: was referring to `nbd_tls = 0`

gibizer: sorry my bad. Honestly I'm out of my depth here. I assume if we keep this here then we need to add nbd_tls = 1 below in the conditional block to ensure that nbd uses tls when tls is enabled.

bogdando (commented):
looks good

olliewalsh:
Merged into https://github.com/openstack-k8s-operators/edpm-ansible/pull/626

PR #614, opened 2024-04-08
Fix TLS for ceilometer

vyzigold:
Ceilometer was failing to access the CA certificate, which wasn't correctly mounted. This mounts the CA certificate to the expected location if TLS is enabled.

Jira: https://issues.redhat.com/browse/OSPRH-6177

### DIFF
--- a/roles/edpm_telemetry/templates/ceilometer_agent_compute.json.j2
+++ b/roles/edpm_telemetry/templates/ceilometer_agent_compute.json.j2
@@ -16,10 +16,10 @@
         "/etc/hosts:/etc/hosts:ro",
         "/etc/pki/tls/certs/ca-bundle.trust.crt:/etc/pki/tls/certs/ca-bundle.trust.crt:ro",
         "/etc/localtime:/etc/localtime:ro",
-        "/etc/pki/ca-trust/extracted:/etc/pki/ca-trust/extracted:ro",
         "/etc/pki/ca-trust/source/anchors:/etc/pki/ca-trust/source/anchors:ro",
-        "/dev/log:/dev/log",
-        "/etc/pki/tls/cert.pem:/etc/pki/tls/cert.pem:ro",
-        "/etc/pki/tls/certs/ca-bundle.crt:/etc/pki/tls/certs/ca-bundle.crt:ro"
+{% if edpm_telemetry_tls_certs_enabled %}
+	"{{ edpm_telemetry_cacerts }}/tls-ca-bundle.pem:/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem:ro,z",
+{% endif %}
+        "/dev/log:/dev/log"
     ]
 }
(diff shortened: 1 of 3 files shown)


vakwetu:
+2

roles/edpm_telemetry/templates/ceilometer-agent-ipmi.json.j2
```diff
@@ -20,6 +20,15 @@
         "perm": "0600",
         "optional": true
       },
+{% if edpm_telemetry_tls_certs_enabled %}
+      {
+        "source": "/var/lib/openstack/cacerts/tls-ca-bundle.pem",
+        "dest": "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem",
+        "owner": "ceilometer",
+        "perm": "0600",
```
olliewalsh, line 28: should be world readable

roles/edpm_telemetry/templates/ceilometer-agent-compute.json.j2
```diff
@@ -20,6 +20,15 @@
         "perm": "0600",
         "optional": true
       },
+{% if edpm_telemetry_tls_certs_enabled %}
+      {
+        "source": "/var/lib/openstack/cacerts/tls-ca-bundle.pem",
+        "dest": "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem",
+        "owner": "ceilometer",
+        "perm": "0600",
```
olliewalsh, line 28: should be world readable

roles/edpm_telemetry/templates/ceilometer-agent-compute.json.j2
```diff
@@ -20,6 +20,15 @@
         "perm": "0600",
         "optional": true
       },
+{% if edpm_telemetry_tls_certs_enabled %}
+      {
+        "source": "/var/lib/openstack/cacerts/tls-ca-bundle.pem",
+        "dest": "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem",
```
olliewalsh, line 26: can just mount directly, don't need to use kolla to copy it

olliewalsh (changes requested):

roles/edpm_telemetry/templates/ceilometer-agent-compute.json.j2
vyzigold: Yes, thanks for the suggestion. I mounted it directly in the latest commit.

roles/edpm_telemetry/templates/ceilometer_agent_compute.json.j2
```diff
@@ -16,10 +16,10 @@
         "/etc/hosts:/etc/hosts:ro",
         "/etc/pki/tls/certs/ca-bundle.trust.crt:/etc/pki/tls/certs/ca-bundle.trust.crt:ro",
         "/etc/localtime:/etc/localtime:ro",
-        "/etc/pki/ca-trust/extracted:/etc/pki/ca-trust/extracted:ro",
         "/etc/pki/ca-trust/source/anchors:/etc/pki/ca-trust/source/anchors:ro",
-        "/dev/log:/dev/log",
-        "/etc/pki/tls/cert.pem:/etc/pki/tls/cert.pem:ro",
-        "/etc/pki/tls/certs/ca-bundle.crt:/etc/pki/tls/certs/ca-bundle.crt:ro"
+{% if edpm_telemetry_tls_certs_enabled %}
```
olliewalsh, line 20: technically should always mount this if it exists. Could have a custom TLS CA bundle without having internal TLS enabled (e.g need to connect to an external service using a custom CA). However most edpm roles are currently like this so can follow-up with a PR to fix it everywhere. Also expect we need to change dataplane operator logic for this.

olliewalsh (approved):
/lgtm

PR #616, opened 2024-04-09
Install rsync in container for ansible.posix.synchronize

mrkisaolamb:
Currently ansible container doesn't have rsync installed so we are not able to use ansible.posix.synchronize to sync files from ansibleee container too edpm node
Partial: [OSPRH-2369](https://issues.redhat.com/browse/OSPRH-2369)


bogdando (commented):
lgtm

gibizer (approved):
make sense

openstack_ansibleee/Dockerfile
```diff
@@ -3,6 +3,11 @@ FROM quay.io/ansible/creator-ee:v0.14.1 as builder
 ARG REMOTE_SOURCE=.
 ARG REMOTE_SOURCE_DIR=/var/tmp/edpm-ansible
 
+RUN \
+microdnf install -y \
+# we are installing ansible.posix with rsync, but we aren't installing rsync itself
+rsync
+
```
slagle, line 10: i think you need to move this after the FROM on line 16

PR #619, opened 2024-04-10
[edpm_nova]Do not add immutable flag to compute_id 

gibizer:
The podman version 4.6.0 has a bug that causes that sometimes a
container fails to start if it mounts an file with +i flag set.
This podman version is used in the adoption form state making the
adoption unstable (1 in 8 CI run fails). The greenfield 18 deployment
uses podman 4.9.4 that does not show this error. So to make the adoption
stable the edpm_nova role will not add the +i flag to existing
compute_id files. Still if the role pre-generates the file that means it
is a greenfield run so the flag can be, and will be, set safely.

### DIFF
--- a/roles/edpm_nova/tasks/configure.yml
+++ b/roles/edpm_nova/tasks/configure.yml
@@ -178,5 +178,8 @@
     owner: nova
     group: nova
     mode: "0400"
-    attributes: "+i"
+    # NOTE(gibi): podman 4.6.0 (OSP 17.1) has a race condition that leads to
+    # container startup failure if a mounted file has the attribute set
+    # while podman 4.9.4 (OSP 18) does not show this issue
+    # attributes: "+i"
   when: compute_id.stat.exists


bogdando (approved):
/lgtm

SeanMooney (approved):
once 17.1.3 is out we can likely remove some of this code form the adoption workflow anyway so i agree with removing this for now and evaluating later once that is released how we want to proceed

bogdando:
At which point and where that flag will be set for greenfield?

gibizer:
> At which point and where that flag will be set for greenfield?

https://github.com/openstack-k8s-operators/edpm-ansible/pull/619/files#diff-36ab95b79773d7839c0045b3afd39c2c9acdf90e31693d9a61afc50af8f6ade2R165

fao89:
it seems we got some hiccup with tide

gibizer:
merging this manually as tide seems to be in bad state. all the relevant tests passed.

PR #622, opened 2024-04-10
Check an FQDN is set

slagle:
An FQDN should be set, and a validation is added to check. The
validation can be disabled with the edpm_nodes_validation_check_for_fqdn
variable.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/797
Jira: https://issues.redhat.com/browse/OSPRH-6187
Jira: [OSPRH-16536](https://issues.redhat.com//browse/OSPRH-16536)

### DIFF
--- a/roles/edpm_nodes_validation/defaults/main.yml
+++ b/roles/edpm_nodes_validation/defaults/main.yml
@@ -22,5 +22,7 @@ edpm_nodes_validation_hide_sensitive_logs: true
 edpm_nodes_validation_ping_test_ips: []
 edpm_nodes_validation_validate_controllers_icmp: true
 edpm_nodes_validation_validate_fqdn: false
+edpm_nodes_validation_validate_fqdn_hosts_file: /etc/hosts
+edpm_nodes_validation_check_for_fqdn: true
 edpm_nodes_validation_validate_gateway_icmp: true
 edpm_nodes_validation_ping_test_gateway_ips: []
--- a/roles/edpm_nodes_validation/meta/argument_specs.yml
+++ b/roles/edpm_nodes_validation/meta/argument_specs.yml
@@ -16,10 +16,18 @@ argument_specs:
         description: Attempt to reach controllers with ping.
         type: bool
         default: true
+      edpm_nodes_validation_check_for_fqdn:
+        description: Verify if an FQDN is set. Typically required for TLS-e.
+        type: bool
+        default: true
       edpm_nodes_validation_validate_fqdn:
         description: Verify if hostname matches FQDN from /etc/hosts
         type: bool
         default: false
+      edpm_nodes_validation_validate_fqdn_hosts_file:
+        description: Hosts file to check for verifying that hostname matches FQDN
+        type: str
+        default: /etc/hosts
       edpm_nodes_validation_validate_gateway_icmp:
         description: Attempt to reach gateway with ping.
         type: bool
--- a/roles/edpm_nodes_validation/tasks/main.yml
+++ b/roles/edpm_nodes_validation/tasks/main.yml
@@ -64,11 +64,26 @@
     - edpm_nodes_validation_validate_controllers_icmp|bool
     - edpm_nodes_validation_ping_test_ips | length > 0
 
-- name: Verify the configured FQDN vs /etc/hosts
+- name: Verify a FQDN is set block
+  when:
+    - edpm_nodes_validation_check_for_fqdn|bool
+  block:
+    - name: Run hostname -f
+      ansible.builtin.command: hostname -f
+      register: hostname
+      changed_when: false
+
+    - name: Verify a FQDN is set
+      ansible.builtin.assert:
+        that: hostname.stdout.find(".") != -1
+        fail_msg: "{{ hostname.stdout }} does not contain . and does not appear to be an FQDN."
+        success_msg: "{{ hostname.stdout }} contains . and appears to be an FQDN."
+
+- name: Verify the configured FQDN vs {{ edpm_nodes_validation_validate_fqdn_hosts_file }}
   ansible.builtin.shell: |
     HOSTNAME=$(hostname)
     SHORT_NAME=$(hostname -s)
-    FQDN_FROM_HOSTS=$(awk '$1 !~ /#/ && $3 == "'${SHORT_NAME}'"{print $2}' /etc/hosts)
+    FQDN_FROM_HOSTS=$(awk '$1 !~ /#/ && $3 == "'${SHORT_NAME}'"{print $2}' {{ edpm_nodes_validation_validate_fqdn_hosts_file }})
     if [[ $HOSTNAME != $FQDN_FROM_HOSTS ]]; then
       echo "FAILURE"
       echo -e "System hostname: ${HOSTNAME}\nEntry from /etc/hosts: ${FQDN_FROM_HOSTS}\n"
--- a/roles/edpm_nodes_validation/molecule/default/prepare.yml
+++ b/roles/edpm_nodes_validation/molecule/default/prepare.yml
@@ -26,5 +26,12 @@
       ansible.builtin.dnf:
         name: hostname
         state: present
+    - name: Set a FQDN
+      command: hostname instance.localdomain
+    - name: Add FQDN to /tmp/hosts
+      lineinfile:
+        path: /tmp/hosts
+        line: 127.0.0.1 instance.localdomain instance
+        create: true
   roles:
     - role: osp.edpm.env_data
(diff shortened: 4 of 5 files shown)


roles/edpm_nodes_validation/tasks/main.yml
```diff
@@ -63,11 +63,25 @@
...
-- name: Verify the configured FQDN vs /etc/hosts
+- name: Verify a FQDN is set block
+  when:
+    - edpm_nodes_validation_check_for_fqdn|bool
+  block:
+  - name: Run hostname
+    command: hostname
+    register: hostname
+
+  - name: Verify a FQDN is set
```
rabi, line 74: Setting fqdn in /etc/hostname is not mandatory. Can you elaborate what tls-e issue we're solving with this?

rabi (dismissed):
Probably you missed my comment in the jira. Neither we should enforce fqdn as hostname nor we use /etc/hosts in OSP18.

roles/edpm_nodes_validation/meta/argument_specs.yml
```diff
@@ -16,10 +16,18 @@ argument_specs:
         description: Attempt to reach controllers with ping.
         type: bool
         default: true
+      edpm_nodes_validation_check_for_fqdn:
+        description: Verify if an FQDN is set. Typically required for TLS-e.
```
olliewalsh, line 20: Shouldn't imply this is only required for TLS-e (it really cares) since it will silently break lots of other stuff too

roles/edpm_nodes_validation/tasks/main.yml
olliewalsh: TLS isn't the main concern. The fqdn MUST NOT change as it is the uid for many services.
If hostname is not explicitly the fqdn then AFAIK it can e.g if a DHCP lease could set the domain. Resolving this in the past required intrusive database surgery.

cloud-init sets it AFAICT (runs `hostname-ctl set-hostname <fqdn>`)

roles/edpm_nodes_validation/tasks/main.yml
```diff
@@ -63,11 +63,25 @@
...
+    ansible.builtin.assert:
+      that: hostname.stdout.find(".") != -1
+      fail_msg: "{{ hostname.stdout }} does not contain . and does not appear to be an FQDN."
+      success_msg: "{{ hostname.stdout }} contains . and appears to be an FQDN."
+
+- name: Verify the configured FQDN vs {{ edpm_nodes_validation_validate_fqdn_hosts_file }}
```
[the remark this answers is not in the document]
olliewalsh: Only thing that it might be worth checking is that /etc/hosts does not explicitly set the hostname or fqdn to the loopback address. Stackoverflow told everyone to do this a while back so might be done out of habit

roles/edpm_nodes_validation/tasks/main.yml
rabi: > If hostname is not explicitly the fqdn then AFAIK it can e.g if a DHCP lease could set the domain.

I'm not sure I get the context here. It would be good to know when and how this could happen in the context of EDPM nodes (wrt dhcp lease) and what past issues we're referring to. We discussed the issues from services (ex.nova) pov in https://github.com/openstack-k8s-operators/edpm-ansible/pull/587 and decided not to change /etc/hostname.


> cloud-init sets it AFAICT (runs `hostname-ctl set-hostname <fqdn>`)

We use cloud-init with baremetal provisioning[1] and I don't see it setting fqdn as hostname[2].

[1] https://github.com/openstack-k8s-operators/openstack-baremetal-operator/blob/main/templates/openstackbaremetalset/cloudinit/userdata#L2-L4

[2] https://logserver.rdoproject.org/12/812/e4ca87621bb89d6ab5a56ab8a92195079e9eafbc/github-check/cifmw-crc-podified-edpm-baremetal/c531266/ci-framework-data/logs/192.168.122.100/etc/hostname

roles/edpm_nodes_validation/tasks/main.yml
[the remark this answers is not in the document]
slagle: The task checking /etc/hosts already existed. It could be adoption relies on it, per 12f54de78cd98be81967beeaf4e98911af55f0cd @fao89 can you comment on that?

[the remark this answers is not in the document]
fao89: actually, adotion@main doesn't need that, it was for adoption@dev-preview3
@jistr and @gibizer have more context on why we needed that for DP3

bogdando (commented):
I think this overlaps with https://github.com/openstack-k8s-operators/edpm-ansible/pull/609 ? @gibi WDYT?

gibizer:
> I think this overlaps with #609 ? @gibi WDYT?

I think 609 is more adoption focused and it mainly there to ensure that the hostname the edpm openstack service are configured with are not changing  during adoption. This PR cannot ensure that.

bogdando:
we don't need different checks for greenfield vs adoption.
please redesign the inputs requirements for osdpns instead.
Make explicit hostnames required for each edpm host in each osdpns ansible vars.
Make a **single check** to fail a deployment, when that intput doesn't match a discovered hostname -f of a edpm node

bogdando (changes requested):
please see my comment

slagle:
> we don't need different checks for greenfield vs adoption. please redesign the inputs requirements for osdpns instead. Make explicit hostnames required for each edpm host in each osdpns ansible vars. Make a **single check** to fail a deployment, when that intput doesn't match a discovered hostname -f of a edpm node
> 
> In case of baremetal provisioned nodes, those hostnames inputs may come automatically, by webhooks, or the BM controller itself, based on its "introspected inventory" data outputs

I don't think that would solve anything since `canonical_hostname` is what is used in various service configs.

bogdando:
we need this change a a minimal fix to https://issues.redhat.com/browse/OSPRH-6187 and all of its related issues (a better fix might still be needed). So LGTM

roles/edpm_nodes_validation/tasks/main.yml
```diff
@@ -64,11 +64,26 @@
     - edpm_nodes_validation_validate_controllers_icmp|bool
     - edpm_nodes_validation_ping_test_ips | length > 0
 
-- name: Verify the configured FQDN vs /etc/hosts
+- name: Verify a FQDN is set block
+  when:
+    - edpm_nodes_validation_check_for_fqdn|bool
+  block:
+    - name: Run hostname
+      ansible.builtin.command: hostname
```
rabi, line 72: I think you've to check `hostname -f` here which will return the fqdn .. This will fail when either hostname is not fqdn or there is no search domain in resolv.conf to attach to the shortname and resolv with dns configured.  Both pre-provisioned/baremetal jobs are failing in this PR because of that looks like.

slagle: Done.

slagle (commented):
recheck

```
2025-05-20 15:59:59.700174 | controller | TASK [Copy pre adoption tempest log files from standalone vm] ******************
2025-05-20 15:59:59.700215 | controller | fatal: [nested_vm]: FAILED! => {"changed": false, "cmd": "/usr/bin/rsync --delay-updates -F --compress --archive --rsh='/usr/bin/ssh -S none -i /home/zuul/.ssh/id_rsa -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null' --out-format='<<CHANGED>>%i %n%L' root@192.168.122.100:/home/zuul/ci-framework-data/tests/pre-adoption-tempest/ /home/zuul/logs/standalone/tests", "msg": "Warning: Permanently added '192.168.122.100' (ED25519) to the list of known hosts.\r\nrsync: [sender] change_dir \"/home/zuul/ci-framework-data/tests/pre-adoption-tempest\" failed: No such file or directory (2)\nrsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1839) [Receiver=3.2.5]\nrsync: [Receiver] write error: Broken pipe (32)\n", "rc": 23}
```

roles/edpm_nodes_validation/defaults/main.yml
```diff
@@ -22,5 +22,7 @@ edpm_nodes_validation_hide_sensitive_logs: true
 edpm_nodes_validation_ping_test_ips: []
 edpm_nodes_validation_validate_controllers_icmp: true
 edpm_nodes_validation_validate_fqdn: false
+edpm_nodes_validation_validate_fqdn_hosts_file: /etc/hosts
+edpm_nodes_validation_check_for_fqdn: true
```
rabi, line 26: We probably don't need this var as I don't see a scenario where it would required to be set false.

rabi (approved):
/lgtm

slagle:
recheck

PR #623, opened 2024-04-12
[edpm_update] Added `edpm_update` role and `update` playbook

rebtoor:
The `edpm_update` role is added to allow the update of packages and containers on EDPM nodes.

The related `update.yml` playbook has been added.

Jira: https://issues.redhat.com/browse/OSPRH-6219

### DIFF
--- a/playbooks/update.yml
+++ b/playbooks/update.yml
@@ -1,0 +1,13 @@
+---
+
+- name: EDPM Update
+  hosts: "{{ edpm_override_hosts | default('all', true) }}"
+  strategy: linear
+  any_errors_fatal: "{{ edpm_any_errors_fatal | default(true) }}"
+  max_fail_percentage: "{{ edpm_max_fail_percentage | default(0) }}"
+  tasks:
+    - name: Import edpm_update
+      ansible.builtin.import_role:
+        name: osp.edpm.edpm_update
+      tags:
+        - edpm_update
--- a/roles/edpm_update/meta/argument_specs.yml
+++ b/roles/edpm_update/meta/argument_specs.yml
@@ -17,3 +17,9 @@ argument_specs:
         default:
           - opensvswitch
         description: List of packages to exclude from the update
+      edpm_update_running_services:
+        type: list
+        default: edpm_services
+        description: >
+          Used to select which list of services are running
+          on the machine. The list is provided by dataplane-operator.
--- a/roles/edpm_update/meta/argument_specs.yml
+++ b/roles/edpm_update/meta/argument_specs.yml
@@ -13,3 +13,8 @@ argument_specs:
         type: bool
         default: true
         description: Toggle to enable/disable containers updates
+      edpm_update_exclude_packages:
+        type: list
+        default:
+          - opensvswitch
+        description: List of packages to exclude from the update
--- a/roles/edpm_update/meta/argument_specs.yml
+++ b/roles/edpm_update/meta/argument_specs.yml
@@ -3,7 +3,6 @@ argument_specs:
   # ./roles/edpm_update/tasks/main.yml entry point
   main:
     short_description: The main entry point for the edpm_update role.
-    description: Multiple lines description
     options:
       edpm_update_enable_packages_update:
         type: bool
--- a/roles/edpm_update/meta/argument_specs.yml
+++ b/roles/edpm_update/meta/argument_specs.yml
@@ -1,0 +1,15 @@
+---
+argument_specs:
+  # ./roles/edpm_update/tasks/main.yml entry point
+  main:
+    short_description: The main entry point for the edpm_update role.
+    description: Multiple lines description
+    options:
+      edpm_update_enable_packages_update:
+        type: bool
+        default: true
+        description: Toggle to enable/disable packages updates
+      edpm_update_enable_containers_update:
+        type: bool
+        default: true
+        description: Toggle to enable/disable containers updates
--- a/roles/edpm_update/molecule/default/converge.yml
+++ b/roles/edpm_update/molecule/default/converge.yml
@@ -23,3 +23,4 @@
         name: osp.edpm.edpm_update
       vars:
         edpm_update_enable_containers_update: false
+        edpm_services: []
--- a/roles/edpm_update/molecule/default/converge.yml
+++ b/roles/edpm_update/molecule/default/converge.yml
@@ -1,0 +1,25 @@
+---
+# Copyright 2024 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+- name: Converge
+  hosts: all
+  tasks:
+    - name: "Call edpm_update role"
+      ansible.builtin.include_role:
+        name: osp.edpm.edpm_update
+      vars:
+        edpm_update_enable_containers_update: false
--- a/roles/edpm_update/molecule/default/molecule.yml
+++ b/roles/edpm_update/molecule/default/molecule.yml
@@ -7,7 +7,7 @@ driver:
   name: podman
 platforms:
 - command: /sbin/init
-  dockerfile: ../../../../molecule/common/Containerfile.j2
+  dockerfile: Containerfile.j2
   image: ${EDPM_ANSIBLE_MOLECULE_IMAGE:-"ubi9/ubi-init"}
   name: instance
   privileged: true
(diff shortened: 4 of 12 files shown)


roles/edpm_ovn/tasks/update.yml
slagle, line 1: for all the update.yml, I think this works for the simplest case when the image is using tags, such as `:latest`.

since we plan to use sha256 as well, we'll also need to run the task to render the container definition files since those contain the image. Then, we need to use the edpm_container_standalone role to restart the container since that will write out the updated systemd unit file (which has the hardcoded image).

This effectively what we already have in run.yml for roles such as these (edpm_ovn for example).

Do you think we should just use run.yml  for both initial deploy and updates? Or use a dedicated update.yml, which will look pretty similar to run.yml?

slagle (commented):
Thanks, this is what we discussed and looks good from that perspective.

However, it occurred to me that since openstack-operator will be providing the image url using sha256 instead of tags, that the simplest case of pull the image and restart the container won't work for that, we'll also need to use the tasks from run.yml.

jpodivin:
The update images and restart containers tasks repeat a lot with modest variation. Would it be possible to factor them out into their own role, and invoke that with proper arguments in the respective plays?

roles/edpm_ovn/tasks/update.yml
rebtoor: @slagle imho we can reuse `run.yml` and avoid a new task file, also for the issue raised by @jpodivin below.

roles/edpm_update/meta/argument_specs.yml
```diff
@@ -0,0 +1,15 @@
+---
+argument_specs:
+  # ./roles/edpm_update/tasks/main.yml entry point
+  main:
+    short_description: The main entry point for the edpm_update role.
+    description: Multiple lines description
```
jpodivin, line 6: Please replace the place holder.

roles/edpm_update/molecule/default/converge.yml
```diff
@@ -0,0 +1,25 @@
+---
+# Copyright 2023 Red Hat, Inc.
```
jpodivin, line 2: 2024 - as in other files

jpodivin (changes requested):

playbooks/update.yml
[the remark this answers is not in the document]
slagle: another possibility (or perhaps in addition) is to add an ansible variable for the list of services on the nodeset, 'edpm_services'. We would set the value equal to NodeSet.Spec.Services. Then use 'when' clauses to run the right roles only if the service is on the NodeSet.

roles/edpm_update/molecule/default/converge.yml
rebtoor: correct..i need to update the role template

playbooks/update.yml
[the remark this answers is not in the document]
rebtoor: I'd go with the latter proposal since it can be useful for `download_cache` role/service as well.

roles/edpm_update/meta/argument_specs.yml
slagle: Done.

slagle:
/unhold

https://github.com/openstack-k8s-operators/dataplane-operator/pull/854 is now merged

roles/edpm_update/tasks/packages.yml
```diff
@@ -0,0 +1,10 @@
+---
+
+- name: Apply packages updates
+  ansible.builtin.dnf:  # noqa: package-latest
+    name: "*"
```
mtomaska, line 5: I am just thinking out loud here. Shouldnt this just update OSP related packages? Because this will unconditionally update everything on the system, which might be undesirable by the end user.

bshephar: Na, definitely need to update RHEL here as part of the minor update process.

roles/edpm_update/molecule/default/molecule.yml
```diff
@@ -0,0 +1,38 @@
+---
+dependency:
+  name: galaxy
+  options:
+    role-file: collections.yml
+driver:
+  name: podman
+platforms:
+- command: /sbin/init
+  dockerfile: ../../../../molecule/common/Containerfile.j2
```
mtomaska, line 10: Sorry to be picking on this PR again, but I am using this PR to build my change on top of it.
The [containerfile](https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/molecule/common/Containerfile.j2#L9) will run `dnf upgrade -y` so essentailly the DUT(instance) is already updated before the converge of this happens. So in the end, testing of this update role is not testing much. Shouldnt the prepare state of this setup some DUT which could be updated and then converge edpm_update role and verify?

rebtoor: You're right, i guess that we can make an exception for this role.

roles/edpm_update/tasks/containers.yml
```diff
@@ -0,0 +1,105 @@
...
+  when: '"nova-storage" in edpm_services'
+
+- name: Updates containers for edpm_ovn role
+  ansible.builtin.include_role:
+    name: osp.edpm.edpm_ovn
+    tasks_from: run.yml
+  tags:
+    - edpm_ovn
+    - edpm_update
+  when: '"ovn" in edpm_services'
```
rebtoor, line 27: Should we handle override like we are doing in download_cache? (see last update as for @fao89 suggestion)

https://github.com/openstack-k8s-operators/edpm-ansible/pull/637/files#diff-24c4ac4687e3ee39c2855757e6a8ed75ca0a2b62aec0dc242d8e8322065ae6a8

jpodivin (commented):
LGTM. But let's wait for the conversation to finish.

roles/edpm_update/tasks/packages.yml
rebtoor: Quoting the doc:

```
- autoremove
        If `true', removes all "leaf" packages from the system that were originally installed as dependencies of user-installed packages but which are no longer required by any such package.
        Should be used alone or when `state' is `absent'
        default: 'no'
        type: bool
        added in: version 2.4 of ansible-core
```

If we want to use it (and it could have sense, IMHO), we should create another dnf task.

roles/edpm_update/tasks/containers.yml
slagle: Yes, I'd add the same handling here

roles/edpm_update/tasks/packages.yml
```diff
@@ -0,0 +1,10 @@
+---
+
```
slagle, line 2: Can we add the tasks to check for a running yum/dnf as done in https://github.com/openstack-archive/tripleo-ansible/blob/stable/wallaby/tripleo_ansible/roles/tripleo_packages/tasks/update.yml?

roles/edpm_update/tasks/packages.yml
```diff
@@ -0,0 +1,10 @@
...
+- name: Apply packages updates
```
slagle, line 3: this needs become:true

roles/edpm_update/tasks/containers.yml
```diff
@@ -0,0 +1,105 @@
+---
+
+- name: Login to container registries if needed
```
slagle, line 3: i think all of these also need become:true

slagle: actually you can't use become:true on include_role, so maybe just set it in the update.yml playbook. Either that or we'll need to go through and add it on every task that needs it.

slagle: when I tested this, I used become:true in the playbook, so let's just go with that for now.

rebtoor: Actually, we can!

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/include_role_module.html#parameter-apply
e.g.

```
- name: Login to container registries if needed
  ansible.builtin.include_role:
    name: osp.edpm.edpm_podman
    tasks_from: login.yml
    apply:
      become: true
  tags:
    - edpm_podman
    - edpm_update
```

slagle (approved):

fao89 (approved):

fao89:
recheck

slagle:
recheck

slagle:
recheck

don't think the failure is related

`
2024-05-22 14:10:49.955459 | controller | Wednesday 22 May 2024  14:10:49 -0400 (0:01:52.098)       0:11:55.742 *********
2024-05-22 14:11:43.793441 | controller | FAILED - RETRYING: [localhost]: openstack-operator - Push bundle image (5 retries left).
2024-05-22 14:11:43.793512 | controller | FAILED - RETRYING: [localhost]: openstack-operator - Push bundle image (4 retries left).
2024-05-22 14:11:43.793525 | controller | FAILED - RETRYING: [localhost]: openstack-operator - Push bundle image (3 retries left).
2024-05-22 14:11:43.793535 | controller | FAILED - RETRYING: [localhost]: openstack-operator - Push bundle image (2 retries left).
2024-05-22 14:11:43.793551 | controller | FAILED - RETRYING: [localhost]: openstack-operator - Push bundle image (1 retries left).
2024-05-22 14:11:43.793561 | controller | An exception occurred during task execution. To see the full traceback, use -vvv. The error was: AttributeError: 'NoneType' object has no attribute 'split'
`

PR #628, opened 2024-04-16
Libvirt SASL support

olliewalsh:
Configure libvirt SASL authentication.
Requires a secret with LibvirtPassword key set.

Depends-On: https://github.com/openstack-k8s-operators/dataplane-operator/pull/831

Related: [OSPRH-6172](https://issues.redhat.com//browse/OSPRH-6172)


roles/edpm_libvirt/defaults/main.yml
```diff
@@ -62,7 +62,11 @@ edpm_libvirt_packages:
   # ceph-common is needed for ceph support to provide the ceph client lib
   # for qemu and the ceph udev rules
   - ceph-common
+  # for SASL auth
+  - cyrus-sasl-scram
 edpm_libvirt_ceph_path: /var/lib/openstack/config/ceph
+edpm_libvirt_password_path: /var/lib/openstack/configs/{{ edpm_service_name | default('libvirt') }}/LibvirtPassword
```
SeanMooney, line 68: does this really need to be configurable?

when would we ever set edpm_service_name to something other the libvirt.

unlike nova we do not intend to allow custom libvirt dataplane services as we do not want to allow arbitrary configuration ro libvirt.

roles/edpm_libvirt/defaults/main.yml
```diff
@@ -62,7 +62,11 @@ edpm_libvirt_packages:
...
+edpm_libvirt_sasl_auth_enabled: "{{ edpm_libvirt_password_path is exists | bool }}"
```
SeanMooney, line 69: i do nnot think this shoul dbe configurable separate form the generic tls support

we do not want to support sasl auth without tls as we do not want to support connecting to libvirt vai the network without tls.

so we should just enable sasl auth if edpm_libvirt_tls_certs_enabled is true

roles/edpm_libvirt/tasks/configure.yml
```diff
@@ -1,5 +1,11 @@
 ---
 
+- name: Get libvirt password
+  ansible.builtin.set_fact:
+    sasl_password: "{{ lookup('file', edpm_libvirt_password_path, lstrip=false, rstrip=false) }}"
```
SeanMooney, line 5: we generally try to aovid the use of lookup plugigns as it generally not what you want

hum a lookup plugin always executes on the Ansible controller which is generally not what people want but I guess in this case this is appropriate.

this will execute the lookup on the filesystem of the AEE pod to which we would expect a secret to be mounted at /var/lib/openstack/configs/{{ edpm_service_name | default('libvirt') }}/LibvirtPassword

or /var/lib/openstack/configs/libvirt/LibvirtPassword in realitity since we should never have a custom libvirt service.

roles/edpm_libvirt/tasks/configure.yml
```diff
@@ -27,21 +33,41 @@
...
 
+- name: Configure libvirt sasl credentials
+  tags:
+    - configure
+    - libvirt
+  become: true
+  block:
+    - name: Add libvirt sasl password
+      ansible.builtin.command:
+        cmd: saslpasswd2 -f /etc/libvirt/passwd.db -p -a libvirt -u overcloud migration
```
SeanMooney, line 61: why overcloud? we do not use that term anymore

should it not be nova or livbirt?

roles/edpm_libvirt/tasks/configure.yml
```diff
@@ -27,21 +33,41 @@
...
+        stdin: "{{ sasl_password }}"
```
SeanMooney, line 62: rather than setting it as a fact with no_log you could just look it up here

"{{ lookup('file', edpm_libvirt_password_path, lstrip=false, rstrip=false) }}"

roles/edpm_libvirt/templates/auth.conf
```diff
@@ -0,0 +1,8 @@
+{% if edpm_libvirt_sasl_auth_enabled|bool %}
+[credentials-overcloud]
+authname=migration@overcloud
```
SeanMooney, line 3: again we do not use the term overcloud here so we should use a different value here.

based on the context i would suggest nova or OpenStack.

roles/edpm_libvirt/templates/virtproxyd.conf
```diff
@@ -2,5 +2,10 @@ unix_sock_ro_perms="0444"
 unix_sock_rw_perms="0770"
 auth_unix_ro="none"
 auth_unix_rw="none"
+auth_tls={% if edpm_libvirt_sasl_auth_enabled|bool %}
```
SeanMooney, line 5: as noted above i would prefer to use edpm_libvirt_tls_certs_enabled here
since we don't want to allow using sasl auth indepently of tls.

SeanMooney (changes requested):

roles/edpm_libvirt/defaults/main.yml
olliewalsh: you need the secret too

roles/edpm_libvirt/defaults/main.yml
olliewalsh: you must support custom libvirt services at least to allow a different issuer to be use for the TLS certs

roles/edpm_libvirt/templates/auth.conf
olliewalsh: copied from puppet-tripleo, didn't want to change it in case it's significant for adoption (probably not)

olliewalsh: can address nits such as this in a follow-up

roles/edpm_libvirt/tasks/configure.yml
olliewalsh: also used in the auth.conf template

roles/edpm_libvirt/defaults/main.yml
olliewalsh: > we do not want to support sasl auth without tls as we do not want to support connecting to libvirt vai the network without tls.

Can't do that. With this can have TLS, or TLS+SASL auth, nothing else

roles/edpm_libvirt/templates/auth.conf
olliewalsh: maybe significant actually, if client sends authname to server. Not willing to risk it for now, can return to it once there is CI coverage for adoption with TLS

roles/edpm_libvirt/defaults/main.yml
olliewalsh: ... but doesn't seem to be optional on tripleo so no reason to disable it

roles/edpm_libvirt/templates/virtproxyd.conf
olliewalsh: doesn't need to be conditional, auth_tls will be ignored if TLS isn't enabled

roles/edpm_libvirt/tasks/configure.yml
olliewalsh: see other comment

roles/edpm_libvirt/templates/auth.conf
olliewalsh: .. and while Gibi (IIRC) mentioned the intention is to not support live-migration during adoption, I think it would be a good thing to avoid breaking it

olliewalsh:
recheck

roles/edpm_libvirt/templates/auth.conf
gibizer: There is no way to do VM lifecycle operations during adoption as at the start of the adoption we stop the nova-compute service on each hypervisor.

olliewalsh:
Passing tempest live migration tests in https://github.com/openstack-k8s-operators/openstack-operator/pull/736

roles/edpm_libvirt/defaults/main.yml
SeanMooney: that not something we planned to do

we could but that is beyond the MVP we require
I'm not against you supporting it in the code but I'm not a fan of documenting this
in our product docs.

i do not want to suggest to customers that we intend to allow them to modify the libvirt configuration in general because we don't want to support that.

roles/edpm_libvirt/templates/auth.conf
SeanMooney: yep during adoption this code will run and regenerate all the auth config so there should be no impact on adoption to using a more approate name.

roles/edpm_libvirt/tasks/configure.yml
SeanMooney: hum i see i dont really like haveing passwords set as facts but i guess that is a valid reason

roles/edpm_libvirt/defaults/main.yml
olliewalsh: maybe not have a choice on this if we want to reach feature parity e.g setting the listen address to a specific network

vakwetu:
+2

SeanMooney (approved):

PR #637, opened 2024-04-24
[edpm_download_cache] Download packages and containers images only when service will be deployed on the EDPM node.

rebtoor:


roles/edpm_download_cache/tasks/container_images.yml
```diff
@@ -9,6 +9,7 @@
     - download_cache
 
 - name: Download images for edpm_iscsid role
+  when: '"nova-storage" in edpm_services'
```
fao89, line 12: We should also take into account the `edpm_services_override`,
maybe we can have a variable that defaults to `edpm_services` and gets `edpm_services_override` value if it is defined

rebtoor: @slagle FYI

fao89: I think we can have something like:
```yaml
edpm_running_services: {{ edpm_services_override | default(edpm_services) }}
```
so if you want the "default" services you can get `edpm_services` and if you want what is being executed you can get `edpm_running_services`

jpodivin: That would be a good idea. The overrides are meant to take priority after all.

rebtoor: It makes sense, I've added `edpm_download_cache_running_services` variable.

roles/edpm_download_cache/tasks/packages.yml
```diff
@@ -57,6 +70,7 @@
     - download_cache
 
 - name: Download images for edpm_libvirt role
```
jpodivin, line 72: Not actually an image download. The role is using `dnf` to download packages. [0]

[0]https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_libvirt/tasks/download_cache.yml

jpodivin (changes requested):

roles/edpm_download_cache/tasks/packages.yml
rebtoor: oops, copy/paste gone wrong. Thanks Jiri.

slagle:
recheck

fao89 (approved):

jpodivin (approved):

PR #638, opened 2024-04-25
Wait for nodes to be reachable

rabi:
Sometimes baremetal nodes are not reachable for few mins after they have been provisioned. As bootstrap is always the first play run, this would allow us wait for sometime for the connections to be ready.

### DIFF
--- a/playbooks/bootstrap.yml
+++ b/playbooks/bootstrap.yml
@@ -1,4 +1,12 @@
 ---
+- name: Check remote connections
+  hosts: "{{ edpm_override_hosts | default('all', true) }}"
+  gather_facts: false
+  tasks:
+    - name: Wait for connection
+      ansible.builtin.wait_for_connection:
+        delay: "{{ edpm_wait_for_connection_delay | default(10) }}"
+        timeout: "{{ edpm_wait_for_connection_timeout | default(600) }}"
 
 - name: Bootstrap node
   hosts: "{{ edpm_override_hosts | default('all', true) }}"


rabi:
recheck RETRY_LIMIT

playbooks/bootstrap.yml
```diff
@@ -1,4 +1,12 @@
 ---
+- name: Check remote connections
+  hosts: "{{ edpm_override_hosts | default('all', true) }}"
+  gather_facts: false
+  tasks:
+    - name: Wait 300 seconds, but only start checking after 10 seconds
+      ansible.builtin.wait_for_connection:
+        delay: 10
```
jpodivin, line 8: I believe this should be parametrized, ideally in a way that allow you to skip this check if you want, or maybe set the time to be slightly long or shorter.

jpodivin (approved):

playbooks/bootstrap.yml
```diff
@@ -1,4 +1,12 @@
 ---
+- name: Check remote connections
+  hosts: "{{ edpm_override_hosts | default('all', true) }}"
+  gather_facts: false
+  tasks:
+    - name: Wait 300 seconds, but only start checking after 10 seconds
```
jpodivin, line 6: I believe this should be parametrized, ideally in a way that allow you to skip this check if you want, or maybe set the time to be slightly long or shorter.

jpodivin (changes requested):

playbooks/bootstrap.yml
rabi: Why would you skip it? 10 seconds initial wait is not too much.  Also, we  don't want to wait longer than 5 mins for sure. Everything does not need to be parameterized.

jpodivin: Maybe because I don't want to wait at all. Or maybe I'm on network with bad latency. Adding parameter with default is easy enough. There is no reason to hard code this.

jpodivin (changes requested):

playbooks/bootstrap.yml
rabi: I don't think the reason for requesting change is justified here. I've never seen customers request that to be changed in TripleO. Anyway, I've updated it now.

jpodivin (approved):

rabi:
recheck

bogdando (approved):
makes sense

rabi:
recheck

PR #640, opened 2024-04-26
[libvirt] enable vencrypt when libvirt tls enabled

stuggi:
Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/780

Jira: [OSPRH-6174](https://issues.redhat.com//browse/OSPRH-6174)

### DIFF
--- a/roles/edpm_libvirt/templates/qemu.conf.j2
+++ b/roles/edpm_libvirt/templates/qemu.conf.j2
@@ -1,7 +1,11 @@
 max_files = 32768
 max_processes = 131072
+{% if edpm_libvirt_tls_certs_enabled|bool %}
+vnc_tls = 1
+vnc_tls_x509_verify = 1
+{% else -%}
 vnc_tls = 0
-vnc_tls_x509_verify = 0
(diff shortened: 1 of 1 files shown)


roles/edpm_libvirt/templates/qemu.conf.j2
```diff
@@ -1,7 +1,12 @@
 max_files = 32768
 max_processes = 131072
+{% if edpm_libvirt_tls_certs_enabled|bool %}
+vnc_tls = 1
+vnc_tls_x509_verify = 1
+{% else -%}
 vnc_tls = 0
 vnc_tls_x509_verify = 0
```
olliewalsh, line 8: can omit `vnc_tls_x509_verify`, that's already the default unless we set `default_tls_x509_verify`

olliewalsh: also isn't relevant when vnc_tls = 0

stuggi:
recheck

gibizer (approved):

PR #645, opened 2024-04-29
Add edpm variable for powersave profile

vcandapp:
Tuned powersave profile setting needs additional
edpm variable to set the c-state value


vcandapp:
Deployed the edpm updates and verifed that the powersave profile is getting set and also the max_power_state is updated in cpu-partitioning-powersave-variables.conf via EDPM

Jaganathancse (approved):
/lgtm

Jaganathancse (approved):

PR #646, opened 2024-04-29
Add edpm_reboot_strategy to reboot role

ciecierski:
Add edpm_reboot_strategy to reboot role that defines if reboot should be started or not. There are three modes: auto, never and force. The default is auto.

Add also more generic check for existence of openstack configs for already
deployed nodes, both old tripleo and new edpm node.

### DIFF
--- a/roles/edpm_bootstrap/tasks/fips.yml
+++ b/roles/edpm_bootstrap/tasks/fips.yml
@@ -61,7 +61,7 @@
             mode: "0600"
         - name: Call edpm_reboot role
           vars:
-            edpm_reboot_force_reboot: true
+            edpm_reboot_strategy: force
           ansible.builtin.include_role:
             name: edpm_reboot
 
--- a/roles/edpm_reboot/defaults/main.yaml
+++ b/roles/edpm_reboot/defaults/main.yaml
@@ -17,7 +17,8 @@
 
 # All variables intended for modification should be placed in this file.
 
-edpm_reboot_force_reboot: false
-edpm_reboot_nova_compute_config_dir: /var/lib/openstack/config/nova
+edpm_reboot_strategy: auto
+edpm_reboot_old_tripleo_node_config_dir: /var/lib/config-data
+edpm_reboot_edpm_node_config_dir: /var/lib/openstack/config
 edpm_reboot_timeout_reboot: 3600
 edpm_reboot_post_reboot_delay: 60
--- a/roles/edpm_reboot/tasks/main.yaml
+++ b/roles/edpm_reboot/tasks/main.yaml
@@ -53,10 +53,15 @@
     paths: /var/lib/openstack/reboot_required/
   register: reboot_files
 
-- name: Check if node has a nova.conf config
+- name: Check if old_tripleo_node config dir exists
   ansible.builtin.stat:
-    path: "{{ edpm_reboot_nova_compute_config_dir }}/01-nova.conf"
-  register: nova_conf_check
+    path: "{{ edpm_reboot_old_tripleo_node_config_dir }}"
+  register: old_tripleo_node_conf
+
+- name: Check if edpm_node config dir exists
+  ansible.builtin.stat:
+    path: "{{ edpm_reboot_edpm_node_config_dir }}"
+  register: edpm_node_conf
 
 - name: Set reboot_required fact
   ansible.builtin.set_fact:
@@ -65,16 +70,17 @@
 
 - name: Print message if reboot is required, but is not going to be started
   ansible.builtin.debug:
-    msg: "Reboot is required but was not started. User has to plan the reboot and set edpm_reboot_force_reboot to true"
+    msg: "Reboot is required but was not started. Edpm_reboot_strategy is set to never or this is
+          already deployed machine. Reboot has to be planned. To start reboot set edpm_reboot_strategy
+          to force"
   when:
     - reboot_required|default(false)
-    - nova_conf_check.stat.exists
-    - not edpm_reboot_force_reboot|bool
+    - (edpm_reboot_strategy|lower == "never" or old_tripleo_node_conf.stat.exists or edpm_node_conf.stat.exists)
 
 - name: Run reboot nodes actions
-  when:
-    - reboot_required is defined and reboot_required|bool
-    - edpm_reboot_force_reboot|bool or not nova_conf_check.stat.exists
+  when: (edpm_reboot_strategy|lower == "force") or
+        (reboot_required|default(false) and edpm_reboot_strategy|lower == "auto" and not old_tripleo_node_conf.stat.exists
+         and not edpm_node_conf.stat.exists)
   block:
     - name: Reboot tasks
       ansible.builtin.include_tasks: reboot.yaml
--- a/docs/source/roles/role-edpm_reboot.rst
+++ b/docs/source/roles/role-edpm_reboot.rst
@@ -14,20 +14,22 @@ This Ansible role allows to do the following tasks:
 
 * Reboot EDPM computes.
   During deployment reboot is triggered automatically if required. During
-  post-deployment reconfiguration reboot is not started. User has to plan
-  reboot maintenance window and set `edpm_reboot_force_reboot` flag to true.
+  post-deployment reconfiguration or adoption process reboot is not started.
+  User has to plan reboot maintenance window and set `edpm_reboot_strategy`
+  flag to force.
 
-Here is an example of a playbook to start reboot:
+
+Here is an example of a playbook to  force start reboot:
 
 .. code-block:: YAML
 
-    - name: Check and start reboot of nodes if required
+    - name: Force start reboot of nodes
       block:
-        - name: "Check and start reboot of nodes if required"
+        - name: "Force start reboot of nodes"
           include_role:
             name: edpm_reboot
            vars:
-             edpm_reboot_force_reboot: true
+             edpm_reboot_strategy: force
 
 
 .. include::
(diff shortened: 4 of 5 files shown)


roles/edpm_reboot/tasks/main.yaml
```diff
@@ -65,16 +65,16 @@
 
 - name: Print message if reboot is required, but is not going to be started
   ansible.builtin.debug:
-    msg: "Reboot is required but was not started. User has to plan the reboot and set edpm_reboot_force_reboot to true"
-  when:
-    - reboot_required|default(false)
-    - nova_conf_check.stat.exists
-    - not edpm_reboot_force_reboot|bool
+    msg: "Reboot is required but was not started. Edpm_reboot_strategy is set to never or this is
+          already deployed machine and reboot has to be planned. To start reboot set edpm_reboot_strategy
```
slagle, line 69: the already deployed check only considers the existence of nova.conf, so perhaps that should be the description. either that, or we need to add a way to flag an edpm node as already deployed independent of nova as there could be other node types.

ciecierski: I discussed it with Jiri and he mentioned networker nodes with ovn. I'm going to add check for ovn files on nodes to include networker nodes. Do you know other node types and check for those?

roles/edpm_reboot/defaults/main.yaml
```diff
@@ -17,7 +17,8 @@
 
 # All variables intended for modification should be placed in this file.
 
-edpm_reboot_force_reboot: false
+edpm_reboot_strategy: auto
+edpm_reboot_networker_node_config_dir: /var/lib/config-data/ovn_controller
```
jistr, line 21: IIUC this should match also on legacy compute nodes, which is nice, since the nova.conf approach probably will not match those (as the config path is different in 17.1).

jistr (commented):
LGTM, thanks :+1:

roles/edpm_reboot/tasks/main.yaml
ciecierski: @slagle In last revision I made a change that makes the check independent of nova conf. Please check pull request once again

bogdando (commented):
lgtm

roles/edpm_reboot/tasks/main.yaml
```diff
@@ -65,16 +70,17 @@
...
-    - nova_conf_check.stat.exists
-    - not edpm_reboot_force_reboot|bool
+    - (edpm_reboot_strategy|lower == "never" or old_tripleo_node_conf.stat.exists or edpm_node_conf.stat.exists)
 
 - name: Run reboot nodes actions
-  when:
-    - reboot_required is defined and reboot_required|bool
-    - edpm_reboot_force_reboot|bool or not nova_conf_check.stat.exists
+  when: (edpm_reboot_strategy|lower == "force") or
+        (reboot_required|default(false) and edpm_reboot_strategy|lower == "auto" and not old_tripleo_node_conf.stat.exists
```
gibizer, line 82: does it need to check not just the old but the new nova conf existence?

ciecierski: It does check in next line.83 (and not edpm_node_conf.stat.exists)
Strictly speaking it checks for any openstack conf that may exists in config paths, because there was an ask to make check more node type agnostic.

ciecierski: No problem, thank you for reviewing change.

jistr:
What is Tide waiting for here in terms of approval? Does it need to be approved by @SeanMooney specifically as the comment above suggests? :)

jistr:
Ah i guess Mikolaj or i would have to be on the approvers list.

SeanMooney:
no it just requries approval form anyone in the owners file
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/OWNERS
or OWNERS_ALIASES
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/OWNERS_ALIASES

roles/edpm_bootstrap/tasks/fips.yml
```diff
@@ -61,7 +61,7 @@
             mode: "0600"
         - name: Call edpm_reboot role
           vars:
-            edpm_reboot_force_reboot: true
+            edpm_reboot_strategy: force
```
SeanMooney, line 64: are we sure that a forced rebbot is something well should
enable by default in any of our roles?

roles/edpm_reboot/defaults/main.yaml
```diff
@@ -17,7 +17,8 @@
 
 # All variables intended for modification should be placed in this file.
 
-edpm_reboot_force_reboot: false
-edpm_reboot_nova_compute_config_dir: /var/lib/openstack/config/nova
```
SeanMooney, line 21: im not sure why this was ever using /var/lib/openstack/config/nova

that was obviously a bug because that directory si reserved for use by the edpm_nova role

we will delete anything in that and recreated every time it runs.

so +1 but we may want to add a linint tests for this at some point to ensure other roles to try to modify files in /var/lib/openstack/config/nova

roles/edpm_reboot/tasks/main.yaml
```diff
@@ -53,10 +53,15 @@
     paths: /var/lib/openstack/reboot_required/
   register: reboot_files
 
-- name: Check if node has a nova.conf config
+- name: Check if old_tripleo_node config dir exists
   ansible.builtin.stat:
-    path: "{{ edpm_reboot_nova_compute_config_dir }}/01-nova.conf"
-  register: nova_conf_check
+    path: "{{ edpm_reboot_old_tripleo_node_config_dir }}"
+  register: old_tripleo_node_conf
```
SeanMooney, line 59: ok so this is trying to skip reboot if there is tripleo content.

we don't every delete those config files during adoption so this will mean we will never reboot. compute nodes in auto mode for any adopted nodes.

SeanMooney (changes requested):
so this pr looks mostly ok but i has no test coverage.
ideally this would have  molecule coverage although I'm not entirely sure how to test this properly.
the reason i marked this as request change is two fold.

first is there a generic adoption task planned to clean up tripleo config? by default we likely should be avoiding remoood for edpm nodes with nova but I'm not sure its safe to force the reboot for fpis mode by default.

so we should likely see what other think before merging

roles/edpm_reboot/tasks/main.yaml
SeanMooney: if there is a jira to plan a post adoption task to clean up these stale config i think we can also ignore this point and proceed with the pr

roles/edpm_bootstrap/tasks/fips.yml
ciecierski: It was 'forced'  before I changed it to edpm_reboot_force_reboot and now to  edpm_reboot_strategy: force . If this is wrong, this is existing bug in tripleo too.

roles/edpm_reboot/defaults/main.yaml
ciecierski: well, in first iteration we took care of computes only. Later it was brought to attention that there are other node types we should check and not reboot.

roles/edpm_reboot/tasks/main.yaml
ciecierski: Auto mode is there only for greenfield deployment only, so regardless if cleanup happen(it should happen I agree) adopted nodes won't be rebooted in auto mode anyway(after adoption they will have new configs generated).

ciecierski:
If there are concerns how this change works for adoption I link change that tests the change in pre-adoption phase.
https://github.com/openstack-k8s-operators/data-plane-adoption/pull/459/files

SeanMooney:
So that pr is not depending on this pr so it's not testing it and it's also failing ci which is not very encouraging 

I'll remove the requested change to allow others to approve but I'm still unsure that how reboot are being done are correct

roles/edpm_reboot/tasks/main.yaml
SeanMooney: So that is problematic because after a deployment is adopted it is ment to be identical to a greenfield deployment with regards to future day 2 operation. So post adoption completing auto should work the same for greenfield and adopted node otherwise we have a big.

bogdando: that is correct, it seems that users then would need to change all osdpns ansible vars for that policy to auto and / or use this value for the future osdp deployments, on "day 2", post-adoption. That is a docs impact though, nothing for this code to change

SeanMooney:
the os-reboot will be used for kernel updates or other changes that require a host reboot as part of day two maintenance.

it feels wrong to require force to be used in that case
auto should work asuimg a prior task has marked the node as needing reboot.

the intended workflow is that you would do a separate deployment including the os-reboot service with --limit to reboot subsets of host that you have drained of workloads in batches.

in that workflow we should not need to update the ansible_vars on the nodeset to force.

if we cant support that usecase we cant do minor updates properly.

SeanMooney (approved):

SeanMooney:
i will file a jira bug for the fips issue

ciecierski:
> So that pr is not depending on this pr so it's not testing it and it's also failing ci which is not very encouraging
Yes, the pr wasn't testing it. I made a patch, but there was missing depends-on. I fixed it and new results are https://github.com/openstack-k8s-operators/data-plane-adoption/pull/459#issuecomment-2133562264 
A non-ceph job failed during adoption, it passed pre-adoption validation with reboot.

SeanMooney:
filed [OSPRH-7301](https://issues.redhat.com/browse/OSPRH-7301)

PR #648, opened 2024-04-30
When kernelargs are changed, ensure reboot file exists

bshephar:
When kernelargs are changed, ensure reboot file exists

This change ensures that the `edpm_kernel` file is created under the
`reboot_required` directory when `grub.cfg` has been updated. This
ensures the user is able to determine when a node needs rebooting[1], and
they can do so by including our `edpm_reboot` role via the `reboot.yml`
playbook in a `servicesOverride` on the deployment.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/docs/source/roles/role-edpm_reboot.rst?plain=1#L10-L18

### DIFF
--- a/roles/edpm_kernel/tasks/kernelargs.yml
+++ b/roles/edpm_kernel/tasks/kernelargs.yml
@@ -155,16 +155,13 @@
 - name: Reboot block
   when:
     - reboot_required is defined and reboot_required
+    - grub_file_entry_check is changed
     - not edpm_kernel_defer_reboot|bool
     - not _workload_protection|default(false)|bool
   block:
     - name: Reboot tasks
       ansible.builtin.include_tasks: reboot.yaml
-      when:
-        - grub_file_entry_check is not changed
 
-    - name: Skipping reboot for deployed node
+    - name: Skipping automated reboot for deployed node
       ansible.builtin.debug:  # noqa: no-handler
-        msg: "Reboot is skipped for kernel arg change, user has to plan the reboot with migration and downtime"
-      when:
-        - grub_file_entry_check is changed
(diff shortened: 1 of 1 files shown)


roles/edpm_kernel/tasks/kernelargs.yml
```diff
@@ -143,6 +143,18 @@
...
+      ansible.builtin.file:
+        path: /var/lib/openstack/reboot_required/
+        state: directory
+        mode: '0755'
+    
+    - name: Create file for reboot required
+      ansible.builtin.file:
+        path: /var/lib/openstack/reboot_required/needs_restarting
+        state: touch
+        mode: '0600'
```
slagle, line 156: this is already happening when reboot.yaml is included. I'd check the ansible log and make sure that was included, if it wasn't, maybe it's the conditions on that task.

bshephar: Yeah, I think because we don't call the `reboot.yml` if the `grub.cfg` is changed:
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_kernel/tasks/kernelargs.yml#L161-L164

I think we probably DO want to call that `reboot.yml` in this case. Maybe just removing the `NOT` from the condition would fix the issue I'm seeing with nodes not being rebooted when kernel args are updated.

bshephar: Converting this to draft while I'm testing it, but I'll try with:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/648/files#diff-42b415cea97608f4334dee2bbaf5a43418ae1318c0e1a926b8aa459354a0ca1aR158

PR #649, opened 2024-04-30
Cleanup role for tripleo services

jpodivin:
Just a role with documentation and tests. Actual invocations of the role, along with removals of other cleanup routines, will be included in follow up PRs.

### DIFF
--- a/docs/source/roles/role-edpm_tripleo_cleanup.rst
+++ b/docs/source/roles/role-edpm_tripleo_cleanup.rst
@@ -1,0 +1,13 @@
+================================
+Role - edpm_tripleo_cleanup
+================================
+
+Role stops and disables all systemd units enumerated to it.
+If the role doesn't recieve list of services, it will instead stop and disable
+all units containing string "tripleo" in their name.
+
+This way we can effectivelly prevent any leftovers of original tripleo based
+deployment from interfering with post-adoption setup.
+
+.. include::
+  ../collections/osp/edpm/edpm_tripleo_cleanup_role.rst
--- a/roles/edpm_tripleo_cleanup/defaults/main.yml
+++ b/roles/edpm_tripleo_cleanup/defaults/main.yml
@@ -1,0 +1,23 @@
+---
+# Copyright 2024 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+# All variables intended for modification should be placed in this file.
+
+edpm_old_tripleo_services: []
+
+# Remove unit files
+edpm_remove_tripleo_unit_files: true
--- a/roles/edpm_tripleo_cleanup/meta/argument_specs.yml
+++ b/roles/edpm_tripleo_cleanup/meta/argument_specs.yml
@@ -1,0 +1,35 @@
+---
+# Copyright 2024 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+argument_specs:
+  # roles/edpm_tripleo_cleanup/tasks/main.yml entry point
+  main:
+    short_description: The main entry point for the edpm_tripleo_cleanup role.
+    options:
+      edpm_old_tripleo_services:
+        type: "list"
+        description: |
+          List of tripleo services to be disabled and stopped.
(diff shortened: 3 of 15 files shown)


roles/edpm_tripleo_cleanup/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+
+- name: Stop and disable tripleo services
+  tags:
+    - adoption
+  become: true
+  ansible.builtin.systemd_service:
+    name: "{{ item }}"
+    state: stopped
+    enabled: false
+  loop: "{{ tripleo_services }}"
```
gibizer, line 42: I think we also want to delete the service definition so that nobody can start those services up after the adoption.

docs/source/roles/role-edpm_tripleo_cleanup.rst
```diff
@@ -0,0 +1,13 @@
+================================
+Role - edpm-edpm_tripleo_cleanup
```
gibizer, line 2: nit: edm-edpm_ should be just edpm_

roles/edpm_tripleo_cleanup/tasks/main.yml
jpodivin: I'd rather not delete them completely. After all, let's give our customers some credit and assume they won't be enabling services for now reason.

But how about just masking them. That would make clear that these aren't supposed to run without limiting people too much.

karelyatin: For what case we want to keep those tripleo systemd units to exists after adoption is done and there is no revert possible? If those are not used would be good to cleanup those completely otherwise those references will live forever on the systems.

jpodivin: It's literally just a file that doesn't do anything and can't even be enabled without further manual intervention. 
Once the package that installed them is removed, which is something we should also automate at some point as part of adoptions, the file will go as well. Assuming it was installed with a package, which I believe is our standard approach.

karelyatin: which package installs those? i don't think any package involved in those tripleo* systemd files, if it does those packages should be fixed.

jpodivin: I would assume they belong to one of the many tripleo* packages. It's something RPMs can handle for you https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_systemd

karelyatin: No AFAIK they don't shipped as part of tripleo* pacakges, we run the services in containers and for managing those we create systemd units as part of roles itself https://github.com/openstack-archive/tripleo-ansible/blob/stable/wallaby/tripleo_ansible/roles/tripleo_container_manage/tasks/create.yml#L27

jpodivin:
recheck - new tests

roles/edpm_tripleo_cleanup/defaults/main.yml
```diff
@@ -0,0 +1,23 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+# All variables intended for modification should be placed in this file.
+
+edpm_old_tripleo_services: []
+
+# Remove unit files
+remove_files: false
```
gibizer, line 23: We probably want to prefix this with edpm_ and make a bit more specific to the role. E.g. edpm_remove_tripleo_service_files .

Also I would make the default true as I believe we should not keep unusable clutter around after the cleanup role.

jpodivin: I'd rather not make it a default, for exactly the reasons I've stated. Those files may be managed by their packages, the removal itself is, unlike masking, disabling etc. irreversible, and the space saved is on order of KBs.

At very best we are freeing equivalent of a small text document, at worst we are inconveniencing administrator.

gibizer: I think @karelyatin showed above that these are not part of a package. 

It is not about saving space it is about to remove confusion. At the point of time when the cleanup role runs the node is not managed by tripleo and will never be able to go back to tripleo as we ran the 18 deployment on it. So having tripleo specific files on the node might wrongly suggest that this node is managed by tripleo or can managed by tripleo in the future.

jpodivin: Very well, it's now the default behavior.

gibizer (commented):
looks good to me. 
I rechecked the data-plane-adoption patch using the new playbook https://github.com/openstack-k8s-operators/data-plane-adoption/pull/438 to see if the newer versions of this PR works in a real env.
We are in a beta freeze and I'm not sure we want to land this now. I let others from the dataplane team decide on that.

roles/edpm_tripleo_cleanup/meta/argument_specs.yml
```diff
@@ -0,0 +1,35 @@
...
+        type: "list"
+        description: |
+          List of tripleo services to be disabled and stopped.
+          If the list is empty, all services containing string 'tripleo'
+          in their names will be disabled, stopped and masked.
+        default: []
+      edpm_remove_tripleo_unit_files:
+        type: "bool"
+        description: |
+          Remove unit files after disabling services.
```
marios, line 32: in two minds about this one. where do we draw the cleanup line? It is one thing to stop the service it is another to cleanup files/packages etc. (yes I realise you're not removing packages here).

jpodivin: At this point I don't really care much either way. The cleanup procedure can be adjusted, or left with defaults. One way or another, given the number of files involving tripleo, we are going to miss something. After all, we have been adopting without cleanup up until now. 

What is critical, is that no tripleo processes are running, or can spontaneously start up after adoption. As that could cause damage.

jistr: I think when we start doing operations on the data plane we are already beyond "point of no return" in terms of rollback so i don't think we'd be reinstating the old services. But i think it's nice that there is a variable to control the removal, thanks.

jistr:
This passed in the adoption CI and the beta freeze is over, so IMO "merge and iterate if necessary" is the way :)

PR #650, opened 2024-05-01
Fix grub2-mkconfig kernel args for --update-bls-cmdline

eshulman2:
In order to update the kernel args properly in RHEL 9.X the --update-bls-cmdline is required.Without this flag the /boot/loader/entries/ files are not updated which causes the kernal args in the grub.cfg to be ignored. Added the flag for RHEL9 and up to allow proper cmdline update during boot.

This patch should fix this issue: [EDPM fails to set kernel args in deployed compute in RHEL 9.4](https://issues.redhat.com/browse/OSPRH-6665)

### DIFF
--- a/roles/edpm_kernel/tasks/kernelargs.yml
+++ b/roles/edpm_kernel/tasks/kernelargs.yml
@@ -99,14 +99,29 @@
         - /boot/efi/EFI/centos
         - /boot/efi/EFI/fedora
 
+    - name: Check if grub2-mkconfig has --update-bls-cmdline option
+      ansible.builtin.shell:
+        cmd: grub2-mkconfig --help | grep '\-\-update-bls-cmdline'
+      ignore_errors: true
+      register: check_update_bls_cmdline
+      changed_when: false
+
     - name: Generate grub config
-      ansible.builtin.command: "grub2-mkconfig -o /boot/grub2/grub.cfg"
+      ansible.builtin.command: >-
+        grub2-mkconfig -o /boot/grub2/grub.cfg
+        {{ '--update-bls-cmdline'
+        if check_update_bls_cmdline.rc == 0
+        else '' }}
       register: _grub2_mkconfig
       changed_when: _grub2_mkconfig.rc == 0
       failed_when: _grub2_mkconfig.rc != 0
 
     - name: Generate EFI grub config
-      ansible.builtin.command: "grub2-mkconfig -o {{ item.stat.path }}/grub.cfg"
+      ansible.builtin.command: >-
+        grub2-mkconfig -o {{ item.stat.path }}/grub.cfg
+        {{ '--update-bls-cmdline'
+        if check_update_bls_cmdline.rc == 0
+        else '' }}
       when: item.stat.exists | bool
       loop: "{{ grub_stat.results }}"
       register: _grub2_mkconfig_efi


ekuris-redhat (approved):
Legit

SeanMooney:
this looks like it was changed in rhel 9.3 
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/9.3_release_notes/new-features#new-features-boot-loader
I'm not sure if this behaviour change has also been applied in centos but 
https://forums.centos.org/viewtopic.php?t=80571 implies it might be.

eshulman2:
@SeanMooney should I change the condition to support Centos as well?

roles/edpm_kernel/tasks/kernelargs.yml
```diff
@@ -100,13 +100,13 @@
         - /boot/efi/EFI/fedora
 
     - name: Generate grub config
-      ansible.builtin.command: "grub2-mkconfig -o /boot/grub2/grub.cfg"
+      ansible.builtin.command: "grub2-mkconfig -o /boot/grub2/grub.cfg {{ '--update-bls-cmdline' if ansible_distribution == 'RedHat' and ansible_distribution_major_version | int >= 9 else '' }}"
```
SeanMooney, line 103: the defualt grub config changed to use bls in 9.3 https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/9.3_release_notes/new-features#new-features-boot-loader

but we are still generating the grub config file.

so provided --update-bls-cmdline is valid on 9.2 for adoption then we should be fine 
otherwise we need to also check the ansible_distribution_minor_version >= 3

SeanMooney:
> @SeanMooney should I change the condition to support Centos as well?

we should confirm if [cifmw-crc-podified-edpm-baremetal](https://review.rdoproject.org/zuul/build/654154f53b954c7fb7b436c7480faf36) or [podified-multinode-edpm-deployment-crc](https://review.rdoproject.org/zuul/build/b4bdebc4d0c140e5bee3fb8383bade1e) exersies this code

idealy i think we would want an enhancement to https://review.rdoproject.org/zuul/build/b407811db8be42328fcac06bff2bf2d8

to assert that the bls is being applied but i think

the current check will skip it on centos so we probably should apply the same logic there also.

roles/edpm_kernel/tasks/kernelargs.yml
```diff
@@ -100,13 +100,25 @@
         - /boot/efi/EFI/fedora
 
     - name: Generate grub config
-      ansible.builtin.command: "grub2-mkconfig -o /boot/grub2/grub.cfg"
+      ansible.builtin.command: >-
+        grub2-mkconfig -o /boot/grub2/grub.cfg
+        {{ '--update-bls-cmdline'
+        if ansible_distribution == 'RedHat'
```
SeanMooney, line 106: so the main open is do we want to also do this if centos 9 stream
or only apply this for downstream/rhel builds

for now we could be conservative and only do it for rhel but my inclination is both.

lest see what others think.

in terms of testing it would be nice to update molecule to have a test for this but that could be done later.
in general i would personally prefer to see this executed in at least one fo our jobs before this merges but lets see what others think.

SeanMooney: so the issue is that the existing test is configured to use the local connection and effectivly the delegated driver

https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_kernel/molecule/kernelargs-update/molecule.yml#L3-L10

which mean that test is running on the os provide by the zuul VM as defined by the label in the relevant nodeset

so its currently running on centos and as a result this wont run.

the current test also only assets the content of /etc/default/grub

https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_kernel/molecule/kernelargs-update/verify.yml#L18-L46

so its not actually testing the current functionally properly
(its not asserting that /boot/grub2/grub.cfg has the expected output)
 and has not been extended for the bsl functionality

we should update the test on rhel to assert that 
the kernel args are in the output of  "sudo grubby --info ALL" i think.

with the current check to have that run in ci upstream we would have to update the nodeset for this molecule job too to use rhel 9.3+

SeanMooney: it looks like there is a cloud-rhel-9-3 label https://review.rdoproject.org/zuul/labels

we do not currently have a nodeset we can directly use but we do have the ability to use labels with rhel

https://github.com/openstack-k8s-operators/ci-framework/blob/66277da980a5664413438f32c522739c6ede987a/zuul.d/nodeset.yaml#L46-L63

for example the centos-9-rhel-9-2-crc-extracted-2.30-3xl nodeset.

so what is probably reasonable here would be to either add centos support and use the existing job or file an issue to add rhel 9 based test coverage.

in either case it would be good to consider if the molecule tests could be enhanced to  assert the bsl entreise are configured properly since /boot/grub2/grub.cfg will be ingored on rhel 9.3+

slagle:
> this looks like it was changed in rhel 9.3 https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/9.3_release_notes/new-features#new-features-boot-loader I'm not sure if this behaviour change has also been applied in centos but https://forums.centos.org/viewtopic.php?t=80571 implies it might be.

grun2-mkconfig doesn't have the option in the latest [CentOS-Stream-GenericCloud-x86_64-9-latest.x86_64.qcow2](https://cloud.centos.org/centos/9-stream/x86_64/images/CentOS-Stream-GenericCloud-x86_64-9-latest.x86_64.qcow2)

SeanMooney:
> > this looks like it was changed in rhel 9.3 https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/9.3_release_notes/new-features#new-features-boot-loader I'm not sure if this behaviour change has also been applied in centos but https://forums.centos.org/viewtopic.php?t=80571 implies it might be.
> 
> grun2-mkconfig doesn't have the option in the latest [CentOS-Stream-GenericCloud-x86_64-9-latest.x86_64.qcow2](https://cloud.centos.org/centos/9-stream/x86_64/images/CentOS-Stream-GenericCloud-x86_64-9-latest.x86_64.qcow2)

based on that do we want to proceed with this without testing for now or hold till we can configure a rhel 9.3+ job? or test downstream only? 

in theory, it should never be the case that Rhel has a feature like this without it first being released in Centos so if it is not in the latest c9s repos then someone didn't follow the correct process here.

slagle:
> > > this looks like it was changed in rhel 9.3 https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/9.3_release_notes/new-features#new-features-boot-loader I'm not sure if this behaviour change has also been applied in centos but https://forums.centos.org/viewtopic.php?t=80571 implies it might be.
> > 
> > 
> > grun2-mkconfig doesn't have the option in the latest [CentOS-Stream-GenericCloud-x86_64-9-latest.x86_64.qcow2](https://cloud.centos.org/centos/9-stream/x86_64/images/CentOS-Stream-GenericCloud-x86_64-9-latest.x86_64.qcow2)
> 
> based on that do we want to proceed with this without testing for now or hold till we can configure a rhel 9.3+ job? or test downstream only?
> 
> in theory, it should never be the case that Rhel has a feature like this without it first being released in Centos so if it is not in the latest c9s repos then someone didn't follow the correct process here.

that's what I was wondering myself. CentOS Stream isn't ahead of RHEL in this case.

eshulman2:
recheck

roles/edpm_kernel/tasks/kernelargs.yml
```diff
@@ -100,13 +100,25 @@
...
+        and ansible_distribution_major_version | int >= 9
+        and ansible_distribution_minor_version | int >= 3
```
rabi, line 108: Does  `ansible_distribution_minor_version` var even exist? I can only see the following facts. Also, it's not good to use vars and instead use `ansible_facts['distribution_major_version']` and we should disable `inject_facts_as_vars` which is very expensive.

`       
        "distribution": "RedHat",
        "distribution_file_parsed": true,
        "distribution_file_path": "/etc/redhat-release",
        "distribution_file_search_string": "Red Hat",
        "distribution_file_variety": "RedHat",
        "distribution_major_version": "9",
        "distribution_release": "Plow",
        "distribution_version": "9.4",
`

eshulman2: My bad I was using very old docs. Should be fixed now.

roles/edpm_kernel/tasks/kernelargs.yml
```diff
@@ -100,13 +100,23 @@
...
+        and ansible_facts['distribution_version'] | float >= 9.3
+        else '' }}
       register: _grub2_mkconfig
       changed_when: _grub2_mkconfig.rc == 0
       failed_when: _grub2_mkconfig.rc != 0
 
     - name: Generate EFI grub config
-      ansible.builtin.command: "grub2-mkconfig -o {{ item.stat.path }}/grub.cfg"
+      ansible.builtin.command: >-
+        grub2-mkconfig -o /boot/grub2/grub.cfg
```
Jaganathancse, line 115: @eshulman2  I think grup2-mkconfig command '-o' option should be same grub2-mkconfig -o {{ item.stat.path }}/grub.cfg and only new arg is the additional.

roles/edpm_kernel/tasks/kernelargs.yml
```diff
@@ -100,13 +100,23 @@
         - /boot/efi/EFI/fedora
 
     - name: Generate grub config
-      ansible.builtin.command: "grub2-mkconfig -o /boot/grub2/grub.cfg"
+      ansible.builtin.command: >-
```
rabi, line 103: We could have checked something like below to not make it distro specific, but this looks good to me for the time being.

if [[ $(grub2-mkconfig --help) =~ '--update-bls-cmdline' ]] ...

rabi (commented):
/lgtm

eshulman2:
Tested patch in downstream environment using RHEL9.4.

Jaganathancse (approved):
/lgtm

roles/edpm_kernel/tasks/kernelargs.yml
```diff
@@ -100,13 +100,23 @@
...
+        grub2-mkconfig -o /boot/grub2/grub.cfg
+        {{ '--update-bls-cmdline'
+        if ansible_facts['distribution'] == 'RedHat'
+        and ansible_facts['distribution_version'] | float >= 9.3
```
SeanMooney, line 107: so this should be 

ansible_facts['distribution_version'] is version( '9.3', '>=' )

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_tests.html#comparing-versions

apparently using filter as a test is deprecated and raises a warning from ansibel 2.5

https://cn-ansibledoc.readthedocs.io/zh-cn/latest/user_guide/playbooks_tests.html#test-syntax

SeanMooney:
i have approved the github workflow to allow the remaining test to run.

the current version should work on any OS where its supported.
we are still missing test coverage however.
so I'm not going to approve the pr as is until a @slagle or others agree to address this testing gap as a follow up.

i would suggest reaching out on slack or attending the edpm call on Monday and adding this as a topic.

eshulman2:
> i have approved the github workflow to allow the remaining test to run.
> 
> the current version should work on any OS where its supported. we are still missing test coverage however. so I'm not going to approve the pr as is until a @slagle or others agree to address this testing gap as a follow up.
> 
> i would suggest reaching out on slack or attending the edpm call on Monday and adding this as a topic.

I will be able to work on tests on Sunday but I have a few questions:

1. should I test it on localhost as it is now and on another host from the pool you mentioned?
2. should we reboot to verify or should be using grubby good enough?
3. are there any additional requirements to test it or should it be sufficient?

Jaganathancse:
> i have approved the github workflow to allow the remaining test to run.
> 
> the current version should work on any OS where its supported. we are still missing test coverage however. so I'm not going to approve the pr as is until a @slagle or others agree to address this testing gap as a follow up.
> 
> i would suggest reaching out on slack or attending the edpm call on Monday and adding this as a topic.

@SeanMooney LGTM, @eshulman2 or i will work on test coverage part related to this as new PR. Can we merge this PR and we can work top of this PR since blocking RHEL9.4 testing?

Jaganathancse (approved):

slagle (approved):

PR #651, opened 2024-05-01
Set ovn_match_northd_version to false by default

booxter:
`true` value is needed when executing "fail safe upgrades" only [0]. (These are upgrades where an LTS release is skipped; or when ovn-controllers are updated after northd.) This is not the upgrade scenario that we follow.

This was fixed in puppet modules in [1] but sadly was missed when the code was imported to this repo (several months before the change in puppet modules.)

[0] https://docs.ovn.org/en/latest/intro/install/ovn-upgrades.html#fail-safe-upgrade
[1] https://opendev.org/openstack/puppet-tripleo/commit/68b4a7c87d3af1200d18149365844d40683cb530

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/792


karelyatin:
/hold

Discussing this over slack

booxter:
@karelyatin considering the slack discussion, are we to move forward with this (post beta obviously) or are we going to wait for staged ovn update work from Dan to land?

booxter:
recheck

booxter:
@karelyatin https://github.com/openstack-k8s-operators/openstack-operator/pull/792 landed. Can we now unhold this PR?

PR #653, opened 2024-05-08
Multiple certs

vakwetu:
This accompanies a change to provide for multiple certs to be issued per dataplane service.

We may have to force merge to unlock the dependencies with the dataplane patch.


vakwetu:
recheck

vakwetu:
Note - we may need to waive CI for this to merge as it requires changes on both the dataplane and edpm_ansible sides.
The failing job passes in https://github.com/openstack-k8s-operators/dataplane-operator/pull/875 (the dataplane side with this as a dependency)

PR #656, opened 2024-05-15
Add output of os-net-config file before applying in edpm_network_config

marios:
Adds a debug task to display the contents of the os-net-config config file before this is applied. This can be especially useful if node connectivity is disrupted after the config is applied.

https://issues.redhat.com/browse/OSPRH-5754

### DIFF
--- a/roles/edpm_network_config/tasks/os_net_config.yml
+++ b/roles/edpm_network_config/tasks/os_net_config.yml
@@ -30,6 +30,16 @@
         dest: "{{ nic_config_file }}"
         mode: '0644'
         backup: true
+    - name: Retrieve and output nic_config_file contents for debug before applying
+      when: edpm_network_config_debug|bool
+      block:
+        - name: Retrieve content of nic_config_file before applying
+          ansible.builtin.slurp:
+            path: "{{ nic_config_file }}"
+          register: os_net_config_config
+        - name: Debug print nic_config_file contents
+          ansible.builtin.debug:
+            msg: "{{ os_net_config_config['content'] | b64decode | trim }}"
     - name: Run edpm_os_net_config_module with network_config
       edpm_os_net_config:
         config_file: "{{ nic_config_file }}"


roles/edpm_network_config/tasks/os_net_config.yml
```diff
@@ -30,6 +29,13 @@
         dest: "{{ nic_config_file }}"
         mode: '0644'
         backup: true
+    - name: Debug display contents of nic_config_file before applying
+      ansible.builtin.command: "cat {{ nic_config_file }}"
+      register: os_net_config_config
+      changed_when: false
+    - name: Print to console
```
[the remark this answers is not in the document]
marios: keeping it unless someone feels strongly

roles/edpm_network_config/tasks/os_net_config.yml
```diff
@@ -30,6 +29,13 @@
...
+      ansible.builtin.debug:
+        msg: "{{ os_net_config_config.stdout }}"
```
raukadah, line 38: Looks good to me. I have two suggestions:

* Instead of cat. Can we switch to slurp module https://docs.ansible.com/ansible/latest/collections/ansible/builtin/slurp_module.html instead of cat command?
* Can we add a var so that end user can enable it when they debug output?

roles/edpm_network_config/tasks/os_net_config.yml
```diff
@@ -16,7 +16,6 @@
 #
 # Apply network configuration with os-net-config.
 #
```
rebtoor, line 18: i'd prefer to keep a newline between header and taskfile content.

roles/edpm_network_config/tasks/os_net_config.yml
pablintino: Another +1 to both points.
@marios I think that if you fix these two points, this PR will be an easy merge.

marios: ack sure will update

roles/edpm_network_config/tasks/os_net_config.yml
marios: ack unintentional restoring

roles/edpm_network_config/tasks/os_net_config.yml
```diff
@@ -30,6 +30,16 @@
...
+    - name: Retrieve and output nic_config_file contents for debug before applying
+      when: edpm_network_config_debug|bool
+      block:
+        - name: Retrieve content of nic_config_file before applying
+          ansible.builtin.slurp:
+            path: "{{ nic_config_file }}"
+          register: os_net_config_config
+        - name: Debug print nic_config_file contents
+          ansible.builtin.debug:
+            msg: "{{ os_net_config_config | b64decode | trim }}"
```
pablintino, line 42: `os_net_config_config` is a dict not the b64 content. To decode that content you need:
`os_net_config_config['content'] | b64decode`

marios: yes right updating sorry did that too quickly :/

marios: (actually during the reviews call ;))

roles/edpm_network_config/tasks/os_net_config.yml
```diff
@@ -30,6 +30,16 @@
         dest: "{{ nic_config_file }}"
         mode: '0644'
         backup: true
+    - name: Retrieve and output nic_config_file contents for debug before applying
+      when: edpm_network_config_debug|bool
```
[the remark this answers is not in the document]
marios: we have it already https://github.com/openstack-k8s-operators/edpm-ansible/blob/323be790fcc83880778684bd65f86a904ac144c8/roles/edpm_network_config/defaults/main.yml#L44 

```

edpm_network_config_debug: "{{ (ansible_verbosity | int) >= 2 | bool }}"

```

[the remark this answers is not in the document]
raukadah: Sorry my bad, it is already there!

raukadah (approved):

lewisdenny (approved):

lewisdenny:
Approving due to multiple lgtm and basic change.

PR #664, opened 2024-05-29
ovs: Don't connect to local ovsdb-server via tcp

booxter:
Just do `unix:`. This allows us to avoid TLSing the connection.

WIP, not ready, should list correct socket path and mount it; will also need to clean up Manager records from pre-adoption after adoption is complete; posted for discussion.


booxter:
On second thought, eventually we are going to move from UNIX to TCP in ovn-operator: https://issues.redhat.com/browse/OSPRH-3020 and it probably makes more sense to keep EDP side run the connection via TCP too. Closing it.

PR #665, opened 2024-05-31
Add ovs update into edpm_ovs role

mtomaska:
Openvswitch(ovs) update requires special attention in respect to the rest of the update on a edpm_node.

The OVS update must not stop and/or restart ovs services in order not to disturb the dataplane networking.
If "restart" is required, it must be done via the `systemctl reload` option, which would trigger OVS builtin mechanism that allows reloading OVS services without disturbing already present Openflows on the edpm node.
By default , we do not reload on update but the `edpm_ovs_reload_after_update`  can be  overwritten to reload if desired.

NOTE about molecule testing:
This PR contains the first multi-scenario non-containerized test. Up to now, non-containerized tests contained a single `default` scenario which was run as [a separate Zuul job](https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/zuul.d/jobs.yaml). 
Currently, the ci-framework does not support running multi-scenario molecule tests. There is work is captured in [JIRA-7812](https://issues.redhat.com/browse/OSPRH-7812) to support it. As a stopgap, this PR includes `cleanup.yml` playbook which should cleanup the test instance between each scenario run.

JIRA: [OSPRH-6141](https://issues.redhat.com/browse/OSPRH-6141)


roles/edpm_update/molecule/default/Containerfile.j2
```diff
@@ -5,9 +5,9 @@ FROM {{ item.registry.url }}/{{ item.image }}
 FROM {{ item.image }}
 {% endif %}
 
-RUN curl -o /etc/yum.repos.d/delorean.repo https://trunk.rdoproject.org/centos9-wallaby/current/delorean.repo && \
```
booxter, line 8: please explain this change here. shouldn't we test edpm_update using wallaby repos?

roles/edpm_ovs/tasks/update.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Update openvswitch packages without stopping the service
```
booxter, line 2: is it possible to run just this role without edpm_update? (Thought that was the plan.)

Should there be molecule tests that target this role specifically?

booxter: Please once you add molecule tests for the role, also enable them in `.github/workflows/molecule.yaml`.

roles/edpm_ovs/tasks/update.yml
```diff
@@ -0,0 +1,24 @@
...
+  become: true
+  tags:
+    - update
+    - ovs
+  ansible.builtin.shell: >
+    dnf update -y openvswitch --setopt=tsflags=noscripts
```
booxter, line 8: please add a comment here explaining why we have to `shell` and not use a ansible module for `dnf` or `rpm`.

please add a comment explaining why we have to skip scripts. Please link to the design document in dev-docs repo here, so that a future reader has context to dive into if needed.

please document the assumption that no useful scripts are defined in the rpm spec (and that's why we think it's safe to skip all of them).

roles/edpm_update/molecule/default/prepare.yml
```diff
@@ -17,5 +17,54 @@
 
 - name: Prepare
   hosts: all
-  roles:
```
booxter, line 20: there should be separate molecule scenarios for

1) do nothing (ovs not updated)
2) ovs updated but not reloaded
3) ovs updated and reloaded

I think these tests should belong to edpm_ovs role itself (not to edpm_update). edpm_update role may want to include the tasks from edpm_ovs (maybe), but it should be possible to execute the `edpm_ovs` role independently, I think.

roles/edpm_update/molecule/default/molecule.yml
```diff
@@ -19,6 +19,12 @@ provisioner:
   log: true
   name: ansible
   inventory:
+    host_vars:
+      test_vars:
+        working_directory: /tmp/molecule_test_dir
```
booxter, line 24: please add a comment explaining why you think you have to use a directory to share files across stages.

roles/edpm_update/molecule/default/verify.yml
```diff
@@ -0,0 +1,54 @@
...
+            "active": ["active"] }
+
+    - name: Read stored openvswitch service pids
+      ansible.builtin.slurp:
+        src: "{{test_vars['working_directory']}}/{{test_vars['ovs_pid_file']}}"
+      register: pids_before_update
+
+    - name: Collect openvswitch service pids after update
+      ansible.builtin.shell:
+        pgrep "ovs(db-server|-vswitch)"
```
booxter, line 32: is it guranteed that `procps-ng` is installed?

roles/edpm_update/molecule/default/verify.yml
```diff
@@ -0,0 +1,54 @@
...
+      register: pids_after_update
+
+    - name: Assert openvswitch service pids did not change after the update
+      ansible.builtin.assert:
+        that: pids_before_update.content | b64decode | split('\n') ==
+              pids_after_update.stdout_lines
```
booxter, line 38: is it guaranteed that the order of lines in each `pgrep` run is the same? (I don't see it documented in the man page for the tool.)

roles/edpm_update/molecule/default/verify.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    - name: Read OpenFlows after the update
+      ansible.builtin.shell: |
+        ovs-ofctl dump-flows --no-stats "{{test_vars['test_bridge']}}"
+      register: ovs_flows_after_update
+      failed_when: ovs_flows_after_update.rc != 0
+
+    - name: Assert openvswitch flows remained unchanged after the update
+      ansible.builtin.assert:
+        that: ovs_flows_before_update.content | b64decode | split('\n') | select() ==
```
booxter, line 53: is the order of dump-flows guaranteed? Your dummy flows all belong to the same table and share the same priority. If the order observed is static, it's probably not guaranteed to stay static in the future.

See in the man page for the command

```              
              By default, ovs-ofctl prints flow entries in the same order that the switch
              sends  them,  which  is unlikely to be intuitive or consistent.  Use --sort
              and  --rsort  to  control  display  order.   The   --names/--no-names   and
              --stats/--no-stats options also affect output formatting.  See the descrip‐
              tions of these options, under OPTIONS below, for more information
```

booxter (changes requested):

roles/edpm_ovs/molecule/default/molecule.yml
```diff
@@ -1,24 +0,0 @@
----
```
booxter, line 1: We should have coverage for the "default" case too, where OVS version hasn't changed. (Basically, for a no-op.) It should be safe (=non-disruptive for DP) to run the task. (I assume that `dnf update` will return 0 and then the role won't do anything. But we should make sure it won't fail.)

roles/edpm_ovs/molecule/common/update/verify.yml
```diff
@@ -0,0 +1,68 @@
...
+          ansible.builtin.assert:
+            that: pids_before_update.content | b64decode | split('\n') | sort() !=
+                  pids_after_update.stdout_lines | sort()
+          when: scenario_name == "update_with_reload"
+
+        - name: Assert openvswitch did not reloaded services
+          ansible.builtin.assert:
+            that: pids_before_update.content | b64decode | split('\n') | sort() ==
+                  pids_after_update.stdout_lines | sort()
+          when: scenario_name == "update_without_reload"
```
booxter, line 68: Should you instead check the variable - edpm_ovs_reload_after_update? Relying on scenario name is a bit fragile.

roles/edpm_ovs/tasks/update.yml
```diff
@@ -0,0 +1,32 @@
...
+    - reload
+    - ovs
+  ansible.builtin.systemd_service:
+    name: openvswitch.service
+    daemon_reload: true
+    state: reloaded
+    masked: false
+  when:
+    - edpm_ovs_reload_after_update
+    - '"Nothing to do." not in update_ovs_result.stdout_lines'
```
booxter, line 32: this scenario (Nothing to do.) should be tested. (via default scenario)

roles/edpm_update/tasks/packages.yml
```diff
@@ -1,5 +1,16 @@
 ---
 
+- name: Import edpm_ovs role for independent openvswitch update
+  ansible.builtin.import_role:
+    name: osp.edpm.edpm_ovs
+    tasks_from: update
+
+- name: Ensure openvswitch is excluded from bulk update
```
booxter, line 8: The package was just updated by the previous task that imported `ovs` role. Why do we need to exclude the package here?

booxter (changes requested):

.github/workflows/molecule.yaml
```diff
@@ -37,6 +37,7 @@ jobs:
         - edpm_nftables
         - edpm_nodes_validation
         - edpm_ovn
+        - edpm_ovs
```
booxter, line 40: if you add it here, then 1) make default scenario run in podman; and 2) remove the ovs job from zuul config.

roles/edpm_ovs/molecule/common/update/prepare.yml
```diff
@@ -0,0 +1,80 @@
...
+      vars:
+        edpm_users_users:
+          - {"name": "hugetlbfs", "gid": "42477", "group_only": true}
+        edpm_users_extra_dirs: []
+
+    - name: Select openvswitch test version to install
+      ansible.builtin.set_fact:
+        # Unversioned name will install the latest available version
+        ovs_test_version:
+          "{{'rdo-openvswitch' if scenario_name == 'update_no_op'
```
booxter, line 38: can we define prepare.yml so that it imports this yaml with the `ovs_test_version` set to the "package version" to install instead? References to test scenario names are ugly and fragile.

booxter: Also, I wonder if rdo- prefix in the name of the package is problematic. It probably means that when we decide to run the scenarios against RHOSO repos, this will fail. Not sure it's something we plan to do though.

roles/edpm_update/tasks/packages.yml
booxter: Was this addressed?

.github/workflows/molecule.yaml
mtomaska: yes, working on that

roles/edpm_ovs/molecule/common/update/prepare.yml
mtomaska: > can we define prepare.yml so that it imports this yaml with the `ovs_test_version` set to the "package version" to install instead? References to test scenario names are ugly and fragile.

Sure I can break this out into a separate yaml and just import it here. But that will still make a decision based on the `scenario_name`. Or I am not understanding you correctly? 

> Also, I wonder if rdo- prefix in the name of the package is problematic. It probably means that when we decide to run the scenarios against RHOSO repos, this will fail. Not sure it's something we plan to do though.

Yes this crossed my mind, but I see the whole edpm_ansible branch is written based on RDO repositories and hardcoded "antelope" "wallaby" etc. references  Updating this will be trivial but first we will need to decide how the whole edpm_ansible will run different branches. I hope I am explaining myself correctly.

roles/edpm_update/tasks/packages.yml
mtomaska: I had comments in pending. See my response from 2 days ago

mtomaska: Actually, i see what you are saying now. The previous task should have already updated OVS so even if the `Apply packages updates` runs with OVS it will be no-op since it has been already updated. I still prefer to have this extra guard but I can drop it if you want

booxter:
@mtomaska default scenario failed with https://github.com/openstack-k8s-operators/edpm-ansible/actions/runs/9486911871/job/26196035141?pr=665

```
  fatal: [instance]: FAILED! => {
      "changed": false,
      "invocation": {
          "module_args": {
              "database": "passwd",
              "fail_key": true,
              "key": "openvswitch",
              "service": null,
              "split": null
          }
      },
      "msg": "One or more supplied key could not be found in the database."
  }
```

booxter:
Sorry, the actual failure is probably

```
  failed: [instance] (item=openvswitch) => {
      "ansible_loop_var": "item",
      "changed": false,
      "invocation": {
          "module_args": {
              "daemon_reexec": false,
              "daemon_reload": false,
              "enabled": true,
              "force": null,
              "masked": false,
              "name": "openvswitch.service",
              "no_block": false,
              "scope": "system",
              "state": "started"
          }
      },
      "item": "openvswitch",
      "msg": "Unable to start service openvswitch.service: A dependency job for openvswitch.service failed. See 'journalctl -xe' for details.\n"
```

booxter:
This is being investigated in: https://github.com/openstack-k8s-operators/edpm-ansible/pull/679

roles/edpm_ovs/tasks/update.yml
```diff
@@ -0,0 +1,33 @@
...
```
slawqo, line 33: I know this is nitpicking and You don't need to address it now (if at all), but should we rely on the output message from dnf? What if someone will have different language set in the system?

slawqo (commented):
/lgtm for me, thx and Ihar's comment about molecule can probably be addressed in the follow up if needed.

roles/edpm_ovs/molecule/README.md
```diff
@@ -0,0 +1,56 @@
+# edpm_ovs molecule<h1>
```
jpodivin, line 1: Why the tags? When you want to set a different heading level in md all you need is another `#`.

roles/edpm_ovs/molecule/common/default/prepare.yml
```diff
@@ -0,0 +1,26 @@
+---
+# Copyright 2023 Red Hat, Inc.
```
jpodivin, line 2: s/2023/2024/

jpodivin (changes requested):

roles/edpm_ovs/tasks/update.yml
```diff
@@ -0,0 +1,33 @@
...
+  tags:
+    - update
+    - ovs
+  # Call dnf via the shell module as the ansible dnf module does not
+  # support passing --setopt options. This is required in order to disable any
+  # OVS rpm scriptlets which would stop and disable OVS services during
+  # a standard update. Further info:
+  # https://github.com/openstack-k8s-operators/dev-docs/blob/main/ovs-update.md#rpm-packaging
+  ansible.builtin.shell: >
+    dnf update -y openvswitch --setopt=tsflags=noscripts
```
karelyatin, line 13: should be use the values from "edpm_ovs_packages" variable instead of hardcoding here?

roles/edpm_update/molecule/default/Containerfile.j2
```diff
@@ -6,6 +6,7 @@ FROM {{ item.image }}
 {% endif %}
 
 RUN curl -o /etc/yum.repos.d/delorean.repo https://trunk.rdoproject.org/centos9-wallaby/current/delorean.repo && \
+    curl -o /etc/yum.repos.d/delorean-deps.repo https://trunk.rdoproject.org/centos9-wallaby/delorean-deps.repo && \
```
karelyatin, line 9: why wallaby deps repo needs to be added in Containerfile? no package seems to be using deps repo while building container. Tests installing antelope repos as needed.

roles/edpm_update/tasks/packages.yml
```diff
@@ -1,5 +1,16 @@
 ---
 
+- name: Import edpm_ovs role for independent openvswitch update
+  ansible.builtin.import_role:
+    name: osp.edpm.edpm_ovs
+    tasks_from: update
```
karelyatin, line 6: Tag not needed here and in below task? edpm_update?

roles/edpm_ovs/molecule/common/update/prepare.yml
karelyatin: Agree with Ihar here, Can't we just pass edpm_ovs_packages with required package name in test scenario itself instead of calculating here?

karelyatin (changes requested):
Overall LGTM, just few fixes inline.

roles/edpm_ovs/tasks/update.yml
mtomaska: Good point about different languages. But I can not come up with a better way to detect that `dnf update` did not actually update anything. Return code is always 0, so I can not use that. 
I dont want to be checking OVS version explicitly as that will change over time ... hmm so it must be something not language dependent

roles/edpm_ovs/molecule/common/cleanup_base.yml
```diff
@@ -0,0 +1,16 @@
...
+- name: Delete test directory with its content
+  become: true
+  ansible.builtin.file:
+    path: "{{working_directory}}"
+    state: absent
+
+- name: Uninstall openvswitch from the test instance
+  become: true
+  ansible.builtin.dnf:
+    name: rdo-openvswitch
```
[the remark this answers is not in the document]
mtomaska: Nope, i think this is just oversight on my end. `openvswitch` is better. thanks

roles/edpm_ovs/tasks/update.yml
mtomaska: Hmm , but if you read the description of `edpm_ovs_packages` "it could be used for pinning versions".  So if someone pins a version that could mess things up. I.e. `dnf update -y openvswitch<some_ver> `  Its also a list so order can change. 
I could perhaps define another variable but then it would always just be set to "openvswitch". So I dont see any benefit in that besides making these tasks more DRY. 
I am not going to make this change, let me know if you have further feedback here. thanks

roles/edpm_update/molecule/default/Containerfile.j2
mtomaska: This is how the #623 set this up. @rebtoor maybe be better to answer this.

I added the additional repo because it contains the `centos<>-nfv-ovs` which contains all ovs packages. This is needed so when this test converges it finds latest ovs rpms when `dnf update openvswitch` is called

roles/edpm_ovs/molecule/common/update/prepare.yml
mtomaska: I moved the version to install into each scenario molecule.yml. Under [`provisioner`][`env`]. Thats the best place I could find for it. Let me know

roles/edpm_ovs/tasks/update.yml
mtomaska: Here is a sample output of `update_ovs_result` when an update [happened](https://paste.opendev.org/show/bgbawBNqLRUadJyZ5bbI/) . 
Here is [no update](https://paste.opendev.org/show/bEtLB6oXxuJtTDMsjFE0/) , already up to date.

Some other solutions that are not language dependant I can think of.
 - When an update happens `stdout` will not contain word  `openvswitch`
 -  no update `stdout_lines` list is very short, so get its len()? Naach this seems flaky

booxter: from `dnf` man page:

```
   Check-Update Command
       Command: check-update
       Aliases: check-upgrade

       dnf [options] check-update [--changelogs] [<package-file-spec>...]
          Non-interactively checks if updates of the specified packages are available.  If
          no <package-file-spec> is given, checks whether any updates at all are available
          for your system. DNF exit code will be 100 when there are updates available  and
          a  list  of  the  updates will be printed, 0 if not and 1 if an error occurs. If
          --changelogs option is specified, also changelog delta of packages about  to  be
          updated is printed.
```

booxter: There's also `check_mode` attribute for Ansible DNF module that returns the availability of an update: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/dnf_module.html#attribute-check_mode

mtomaska: Thanks, classic RTFM on my end :). Ended up going with the `check_mode` as I prefer to use ansible modules instead of raw shell command.

roles/edpm_update/molecule/default/Containerfile.j2
karelyatin: yeap delorean.repo makes sense here for tripleo-repos rpm installation But not the deps repo. Also i see tripleo-repos also not used as using repo-setup from source in prepare_base.yml that would setup required repos including centos-nfv-openvswitch repo. So seems can also drop delorean.repo and that can be seperate change. But as part of this patch delorean-deps part can be dropped

roles/edpm_ovs/tasks/update.yml
karelyatin: Yeap i saw that and that's the reason for asking clarification here. So if someone is explicitly pinning "edpm_ovs_packages" i think that means we should be sticking to that version whether it's fresh deploy or adoption or update. Having more variables just for this doesn't looks good. Is this assumption wrong?

roles/edpm_update/molecule/default/Containerfile.j2
mtomaska: Notice, the `Containerfile.j2` is part of the `edpm_update` role molecule test. This role uses a container to test. So this containerfile sets up all repos for the test container. When the `edpm_update` role test runs it will include 
```  
ansible.builtin.import_role:
    name: osp.edpm.edpm_ovs
    tasks_from: update
```
So that is the reason I added the extro depo, in order for the edpm_ovs update to find ovs rpms.

The, `edpm_ovs` is **a separate role**. There, we test directly on the host and thus we use `repo-setup` tool to help setup required repos for testing.

karelyatin: Thanks for sharing that, i didn't noticed it was changing edpm_update role tests

roles/edpm_ovs/tasks/update.yml
booxter: > So if someone pins a version that could mess things up. I.e. dnf update -y openvswitch<some_ver>

Could you please clarify what it will mess then? Won't it just attempt to update a package that is pinned? (so it will do a patch-level update, if present.)

>  Its also a list so order can change.

The list would require some massaging (`" ".join`?) but what does the order have to do here? I don't think `dnf` considers the order of packages in the command line; it has declarative/transactional semantics, no?

It seems like Yatin suggests to just reuse the existing variable; it sounds reasonable.

booxter: btw if you don't reuse the existing variable, as Yatin suggests, then download-cache may miss your packages: https://github.com/openstack-k8s-operators/edpm-ansible/blob/e685241aa5bf3542401b5e1593582825ab9619c7/roles/edpm_ovs/tasks/download_cache.yml#L4

mtomaska: See latest commit which address this

mtomaska:
recheck

roles/edpm_ovs/molecule/update_when_pinned_version_already_present/converge.yml
```diff
@@ -0,0 +1,33 @@
...
+      ansible.builtin.import_role:
+        name: osp.edpm.edpm_ovs
+        tasks_from: update
+      vars:
+        # Setting an explicit reload, but expecting no-op since no update should occur
+        edpm_ovs_reload_after_update: true
+        # ovs_pre_update_version will prepare the test instance with the specified
+        # version of ovs. Setting the same version to update to during converge should
+        # result in no-op since the specific version is already installed.
+        edpm_ovs_packages: [" {{ ovs_pre_update_version }}"]
```
karelyatin, line 33: extra space

roles/edpm_ovs/tasks/update.yml
```diff
@@ -0,0 +1,46 @@
...
+      tags:
+        - update
+        - ovs
+      # Call dnf via the shell module as the ansible dnf module does not
+      # support passing --setopt options. This is required in order to disable any
+      # OVS rpm scriptlets which would stop and disable OVS services during
+      # a standard update. Further info:
+      # https://github.com/openstack-k8s-operators/dev-docs/blob/main/ovs-update.md#rpm-packaging
+      ansible.builtin.shell: >
+        dnf update -y {{ " ".join(edpm_ovs_packages) }} --cacheonly --setopt=tsflags=noscripts
```
karelyatin, line 29: Let's handle empty list for edpm_ovs_packages as that could trigger full dnf update here, can add the condition on block itself above.

roles/edpm_ovs/tasks/update.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Reload openvswitch service
+      tags:
+        - reload
+        - ovs
+      ansible.builtin.systemd_service:
+        name: openvswitch.service
+        daemon_reload: true
+        state: reloaded
+        masked: false
+      when: edpm_ovs_reload_after_update
```
karelyatin, line 46: edpm_ovs_reload_after_update|bool

roles/edpm_ovs/tasks/update.yml
```diff
@@ -0,0 +1,46 @@
...
+  register: ovs_update_check
+
+- name: Perform ovs update
+  when:
+    - ovs_update_check.rc == 0
+    - ovs_update_check.results | length > 0
+  become: true
+  block:
+    - name: Cache set ovs packages
+      ansible.builtin.include_tasks: download_cache.yml
```
karelyatin, line 17: When running this as part of update seems we need to also enable state: latest in https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_ovs/tasks/download_cache.yml#L3 as without that next step seems can run with cache if present?

mtomaska: I tested this before this change and dnf will handle update scenario correctly with the default `state` present.
E.g.
ovs-3.1 installed and ovs3.2 is requested. The cache directory will be updated to ovs3.2
ovs-3.1 installed and ovs3.1 is requested. The cache directory will stay unchanged and no update needed will be returned
ovs-3.1 installed and some ovs2.17 version is requested. Will fail with conflict version non zero rc code will be set. 

I also tried setting `latest` and the result is the same. Looking at the ansible code, they appear to be doing the same operation.
[ensure function](https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/yum.py#L996)
[present](https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/yum.py#L1021)
[latest](https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/yum.py#L1027)

But for consistancy I can allow passing the state variable

karelyatin: > I tested this before this change and dnf will handle update scenario correctly with the default `state` present. E.g. ovs-3.1 installed and ovs3.2 is requested. The cache directory will be updated to ovs3.2 ovs-3.1 installed and ovs3.1 is requested. The cache directory will stay unchanged and no update needed will be returned ovs-3.1 installed and some ovs2.17 version is requested. Will fail with conflict version non zero rc code will be set.
> 
> I also tried setting `latest` and the result is the same. Looking at the ansible code, they appear to be doing the same operation. [ensure function](https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/yum.py#L996) [present](https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/yum.py#L1021) [latest](https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/yum.py#L1027)
> 
It had failed for me with rdo-openvswitch-3.1 installed, update to rdo-openvswitch-3.2 succeeded, followed by rdo-openvswitch failed as download-cache didn't updated the cache, if this scenario works for you somehow i can cross check 
> But for consistancy I can allow passing the state variable

roles/edpm_ovs/tasks/update.yml
mtomaska: This should be implicitly handled by the `Check if ovs update is available`. If you pass in an empty list there, it will result in an empty list of things to update. And because we have `ovs_update_check.results | length > 0` the update block will never run. But I can add an extra guard for this condition.

karelyatin: Thanks, yes you are right, no need for extra check, it's fine i missed that part.

karelyatin (approved):
/lgtm

PR #667, opened 2024-06-03
Replacing shell wrapped python script with a module

jpodivin:
Existing implementation of the role is working fine. But it doesn't align with our current best practice.
This PR replace script with proper module, including error handling, parameter sanity check, documentation etc.
This module will also provide generic way to get contents of any ini file on target machine.

Approach utilizing [ansible.builitin.ini](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/ini_lookup.html) lookup proved workable. However, since the look up requires file to be present in one of pre-set search paths, its use would necessitate further changes to the role.Possibly including provision for more arguments.

When compared with that, a simple custom module seemed preferable.

### DIFF
--- a/plugins/modules/edpm_get_ini_file.py
+++ b/plugins/modules/edpm_get_ini_file.py
@@ -1,0 +1,117 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+
+# Copyright 2024 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import configparser
+
+import yaml
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+ANSIBLE_METADATA = {
+    'metadata_version': '1.0',
+    'status': ['preview'],
+    'supported_by': 'community'
+}
+
+DOCUMENTATION = """
+---
+module: edpm_get_ini_file
+author:
+    - Jiri Podivin <jpodivin@redhat.com>
+version_added: '0.0.1'
+short_description: Retrieve contents of section, key or entire given ini file
+notes: []
+description: |
+    Retrieve contents of given ini file. Either a single value,
+    section or an entire file as a dictionary.
+options:
+    path:
+        description: |
+            Path to ini file.
+        type: str
+        required: true
+    section:
+        description: |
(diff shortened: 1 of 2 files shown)


slagle (commented):
+2

slagle (commented):
I can't say I'm +2 on this any longer. Replacing 4 lines of python with > 100. We don't need a custom module for this. Unless there is another reason. If so, please update the commit message as to why this is needed.

jpodivin:
recheck

rebtoor:
recheck

SeanMooney:
im not sure which best pratics this is meant to be conforming too.

i know we have stated that we should prefer using existing pyton modules instead of shell commands but i don't generally agree we should write and maintain our own modules to replace all usage of command or shell.

plugins/modules/edpm_get_ini_file.py
```diff
@@ -0,0 +1,117 @@
...
+    'supported_by': 'community'
+}
+
+DOCUMENTATION = """
+---
+module: edpm_get_ini_file
+author:
+    - Jiri Podivin <jpodivin@redhat.com>
+version_added: '0.0.1'
+short_description: Retrieve contents of section, key or entire given ini file
```
SeanMooney, line 38: as im sure your aware openstack does not use ini files

we use oslo.config format which has some changes form microsfos ini format and is actually a superset

configparser can mostly parse it but you need to be carful with multi_opt strings as that + variable interpolation are the two main extetions to ini format.

roles/edpm_pre_adoption_validation/tasks/hostname.yml
```diff
@@ -25,49 +25,52 @@
...
-    # the config value. Grep is not enough as multiple sections have 'host'
-    # key. Also we cannot rely on crudini as it is not available on the
-    # node being adopted.
-    - name: Compare old nova service host to {{ canonical_hostname }}
+    - name: Retrieve original hostname
       when: nova_old_config_file.stat.exists
-      ansible.builtin.command:
-        cmd: >
-          python3 -c
-          "import configparser as c;
```
SeanMooney, line 37: we were usign configparser before because the specific files we are interested in is not using the oslo extensions to the ini format so it can be read with any ini parser such as configparser.

SeanMooney:
im kind of neutral on this.
without proper unit testing of the module i don't think this adds much

plugins/modules/edpm_get_ini_file.py
jpodivin: Actually no, I didn't know that. Even the oslo.config documentation[0] states that it's actually using Ini.
Which I didn't have reason to doubt.

[0] https://docs.openstack.org/oslo.config/yoga/configuration/format.html

slagle (commented):
I think it's easier to read and debug without this change.

plugins/modules/edpm_get_ini_file.py
SeanMooney: it using the ini format as a base but its a superset of the format with many extensions that are not part of the standard ini format like variable interpolation, rich typeing fo variables supporting enum, lists, dictionires, multi opts ectra.

basic ini tools can handle the simple case but no OpenStack project that uses oslo.config actually uses ini file.

PR #675, opened 2024-06-12
Re-enable live-migration NBD TLS

olliewalsh:
Closes: [OSPRH-6173](https://issues.redhat.com//browse/OSPRH-6173)

### DIFF
--- a/roles/edpm_libvirt/molecule/default/verify.yml
+++ b/roles/edpm_libvirt/molecule/default/verify.yml
@@ -121,10 +121,6 @@
         that:
           - "'migration_port_min = 61152' in qemu_conf.content | b64decode"
           - "'migration_port_max = 61215' in qemu_conf.content | b64decode"
-    - name: Assert that nbt_tls is disabled in qemu.conf
-      assert:
-        that:
-          - "'nbd_tls = 0' in qemu_conf.content | b64decode"
 
     - name: Load os_enabled_vtpm selinux boolean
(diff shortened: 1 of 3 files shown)


SeanMooney:
we cannot proceed with this until https://gitlab.com/redhat/centos-stream/src/qemu-kvm/-/merge_requests/240 is merged and the release has been done

i was holding of creating this patch until that happened.

downstream we also need to wait for https://issues.redhat.com/browse/RHEL-33754 before we can merged this so we likely can move forward with this patch for a week or two yet until the 9.4 and c9s rpm are released and on the cdn

SeanMooney:
the package is now in the latest compose built today
https://composes.stream.centos.org/production/latest-CentOS-Stream/compose/AppStream/x86_64/os/Packages/qemu-kvm-9.0.0-4.el9.x86_64.rpm

however its not on the cdn/mirror network so we can likely proceed with this tomorrow or monday

SeanMooney:
still missing form  https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/Packages/ but it should be there later today
once qemu-kvm-9.0.0-4.el9.x86_64.rpm is there we can proceed with this.

SeanMooney:
https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/Packages/qemu-kvm-9.0.0-4.el9.x86_64.rpm is now avaiable

PR #677, opened 2024-06-13
add supprot for limited extra bind mount suppport in edpm_nova

SeanMooney:
This change add support for optionally addding a restrict
set of addtional host bind mounts to the nova_compute container

This functionality is intended only for certifed vendor
integrations not as a generic feature.

Closes: [OSPRH-7644](https://issues.redhat.com//browse/OSPRH-7644)

### DIFF
--- a/roles/edpm_nova/templates/nova_compute.json.j2
+++ b/roles/edpm_nova/templates/nova_compute.json.j2
@@ -29,5 +29,8 @@
         "/etc/nvme:/etc/nvme",
         "/var/lib/openstack/config/ceph:/var/lib/kolla/config_files/ceph:ro",
         "/etc/ssh/ssh_known_hosts:/etc/ssh/ssh_known_hosts:ro"
+{%- for bind_mount in edpm_nova_extra_bind_mounts %}
+        ,"{{ bind_mount.src }}:{{ bind_mount.dest }}:{{ bind_mount.options | default('ro') }}"
+{% endfor -%}
     ]
 }
--- a/molecule/test-helpers/verify_mounts.yaml
+++ b/molecule/test-helpers/verify_mounts.yaml
@@ -1,0 +1,14 @@
+- name: verify dir exists
+  become: true
+  block:
+      - name: "Collect container mounts for: {{ item }}"
+        ansible.builtin.shell: |
+          podman inspect {{ item.name }} | jq '.[0].HostConfig.Binds[]'
+        changed_when: false
+        register:
+          container_mounts
+      - name: Assert mount exists {{ item }}
+        ansible.builtin.assert:
+          that:
+            - "'{{ item.src }}:{{ item.dest }}:{{item.options | default('ro')}}' in container_mounts.stdout"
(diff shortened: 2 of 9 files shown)


gibizer (commented):
Looks good to me. This is according our discussion documented in the Jira.

roles/edpm_nova/templates/nova_compute.json.j2
```diff
@@ -29,5 +29,10 @@
         "/etc/nvme:/etc/nvme",
         "/var/lib/openstack/config/ceph:/var/lib/kolla/config_files/ceph:ro",
         "/etc/ssh/ssh_known_hosts:/etc/ssh/ssh_known_hosts:ro"
+{% if edpm_nova_extra_bind_mounts | length > 0 %}
```
ASBishop, line 32: Perhaps a nit, but is this test necessary? Just let L33 do its thing, which I believe will yield the same results even when the list is empty.

roles/edpm_nova/templates/nova_compute.json.j2
```diff
@@ -29,5 +29,10 @@
...
+    {% for bind_mount in edpm_nova_extra_bind_mounts %}
```
ASBishop, line 33: Another nit: I'm a similar situation, I wanted to suppress the j2 statements from inserting blank lines and found I could append/prepend '-' dashes to yield a clean result. Not sure if that's a factor here, but offer it for your consideration.

ASBishop (commented):
Also looking good to me. Left a couple of comments for consideration while we wait for the final CI results.

roles/edpm_nova/templates/nova_compute.json.j2
SeanMooney: i can try removing it 
i am default to the empty list so it may not be

roles/edpm_nova/templates/nova_compute.json.j2
SeanMooney: yep i can add the - to make it trim space

SeanMooney: im going to finish shortly today so ill likely respin this tomororow

mrkisaolamb (commented):
It looks good to me

bogdando (commented):
lgtm

roles/edpm_nova/templates/nova_compute.json.j2
SeanMooney: im testing this locally and ill push up a new verison once im sure this works

roles/edpm_nova/templates/nova_compute.json.j2
SeanMooney: im also testing this
when i have used that in go tempelates it was tricky to get right with lot of nesting but with just this simple templating i think its probably doable
ill include this when updating for the if

gibizer (commented):
Looks good.

ASBishop (commented):
lgtm too, thanks!

gibizer (approved):

PR #683, opened 2024-06-21
Reworking file and directory ownership

jpodivin:
Playbook level becomes are not a good idea. We should only give the minimum set of permissions possible rather than all at once.

I'm going to test how the deployment behaves once I disable these and progressively add becomes to tasks in roles. Eventually I aim to arrive at minimum viable number of tasks with root access. 

Fixes: https://issues.redhat.com/browse/OSPRH-3763

### DIFF
--- a/roles/edpm_nova/tasks/configure.yml
+++ b/roles/edpm_nova/tasks/configure.yml
@@ -16,7 +16,6 @@
         - nova
 
     - name: Assert edpm_nova_extra_bind_mounts source directories exist
-      become: true
       ansible.builtin.assert:
         that:
           - "source_dir_results.results | map(attribute='stat.exists') | list"
--- a/roles/edpm_ovn/tasks/install.yml
+++ b/roles/edpm_ovn/tasks/install.yml
@@ -15,17 +15,21 @@
 # under the License.
 
 - name: Create persistent directories
+  become: true
   ansible.builtin.file:
     path: "{{ item.path }}"
     state: directory
     setype: "container_file_t"
     mode: "{{ item.mode | default(omit) }}"
+    owner: "{{ item.owner | default(ansible_user) | default(ansible_user_id) }}"
+    group: "{{ item.group | default(ansible_user) | default(ansible_user_id) }}"
   loop:
     - {'path': /var/log/containers/openvswitch, 'mode': '0750'}
     - {'path': /var/lib/edpm-config/firewall, 'mode': '0750'}
-    - {'path': /var/lib/openvswitch/ovn}
+    - {'path': /var/lib/openvswitch/ovn, "owner": "openvswitch", "group": "openvswitch"}
 
 - name: Enable virt_sandbox_use_netlink for healthcheck
+  become: true
   ansible.posix.seboolean:
     name: virt_sandbox_use_netlink
     persistent: true
--- a/roles/edpm_telemetry/tasks/configure.yml
+++ b/roles/edpm_telemetry/tasks/configure.yml
@@ -20,9 +20,9 @@
     path: "{{ item.path }}"
     state: directory
     setype: "{{ item.setype | default('container_file_t') }}"
-    owner: "{{ ansible_user }}"
-    group: "{{ ansible_user }}"
-    mode: "{{ item.mode | default('750') }}"
+    owner: "{{ ansible_user | default(ansible_user_id) }}"
+    group: "{{ ansible_user | default(ansible_user_id) }}"
+    mode: "{{ item.mode | default('0750') }}"
     recurse: true
   loop:
     - {"path": "{{ edpm_telemetry_config_dest }}"}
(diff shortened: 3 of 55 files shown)


jpodivin:
recheck

jpodivin:
recheck - no logs retrieved from last test

jpodivin:
Test operator error @lpiwowar may have some info.  
```
2024-06-28T17:19:01Z	ERROR	Reconciler error	{"controller": "tempest", "controllerGroup": "test.openstack.org", "controllerKind": "Tempest", "Tempest": {"name":"tempest-tests","namespace":"openstack"}, "namespace": "openstack", "name": "tempest-tests", "reconcileID": "f7298ece-7e09-44de-aa1a-3018c238fcbf", "error": "PersistentVolumeClaim \"tempest-tests-7b115\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:324
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:265
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.15.0/pkg/internal/controller/controller.go:226
```

jpodivin:
recheck

jpodivin:
recheck

gateway timeout

booxter:
lgtm

roles/edpm_nova/tasks/configure.yml
```diff
@@ -16,7 +16,6 @@
         - nova
 
     - name: Assert edpm_nova_extra_bind_mounts source directories exist
-      become: true
```
SeanMooney, line 19: the become here was becasue these directores may not be owned by the ansible user.

however if we are just checking it exists then that is probably ok.

SeanMooney (commented):
taking a quick look this looks more targeted but i have not had time to review this properly today.

i would not hold on my review speciricly if other agree this is correct but ill see if i can loop back to this later in the week in either case. perhaps tomorrow.

roles/edpm_telemetry/tasks/configure.yml
```diff
@@ -30,6 +30,7 @@
 - name: Render ceilometer config files
   tags:
     - edpm_telemetry
+  become: true
```
SeanMooney, line 33: become should not need to be used here as the directory should be owned by the current user.

roles/edpm_telemetry/tasks/configure.yml
```diff
@@ -67,6 +68,7 @@
   when: custom_ceilometer_conf.stat.exists
 
 - name: Copy generated ceilometer configs
+  become: true
```
SeanMooney, line 71: the same is true here the dest directory should be owned by the current user

and the become below

roles/edpm_ovn_bgp_agent/tasks/configure.yml
```diff
@@ -22,6 +22,7 @@
 - name: Configure OVN BGP agent
   block:
     - name: Render OVN BGP agent config files
+      become: true
```
SeanMooney, line 25: this also imples that the permission on the dest directory are incorrect 
edpm_ovn_bgp_agent_config_basedir should be writeable by the current user

roles/edpm_ovn/tasks/install.yml
```diff
@@ -15,6 +15,7 @@
 # under the License.
 
 - name: Create persistent directories
+  become: true
```
SeanMooney, line 18: ah this is where the incorrect permission are coming form this should be setting the owner to ansibe_user or ansible_user_id preferably with the double defaulting so that both will be checked

roles/edpm_nova/tasks/configure.yml
jpodivin: This is assert on results of previous task. Become is completely useless here.

SeanMooney: oh yes your correct i didnt fully expand the context.

roles/edpm_ovn/tasks/install.yml
jpodivin: Do we have another preferred owner for these? Otherwise there is no need for double defaulting. Just the `ansible_user` and `ansible_user_id` as a fallback.

roles/edpm_telemetry/tasks/configure.yml
jpodivin: With this we can also get rid of become on the last block.

jpodivin:
edpm-ansible-molecule-edpm_libvirt is failing due to unrelated issue addressed in another PR.

SeanMooney (approved):

PR #684, opened 2024-06-26
Disable martian packts logging as default

eshulman2:
Logging martian packets is done via printk. Depending on the default console configuration, it may cause writing on a serial line. Some UART controllers are slow, and it may stall CPUs for up to 200ms. When this logging happens, it causes random latency issues.

Disable logging of martian packets by default. It can be enabled for debugging purposes.


eshulman2:
CC @rjarry

rjarry (approved):
Hi @eshulman2 thanks for the patch. This looks good to me. Logging to the console causes so many issues, it is best to avoid that at all costs.

slagle (approved):

PR #687, opened 2024-06-26
Disable firewalld service in `edpm_nftables` role

rebtoor:
Since we are defining firewall rules with `edpm_nftables` role (using `nftables` service), it's useless and potentially harmful keep firewalld enabled and running on the host.

Fixes: https://issues.redhat.com/browse/OSPRH-7739


bshephar (approved):
Sounds good to me.

PR #688, opened 2024-07-01
Don't use override services to determine update

bshephar:
This change uses the original edpm_services list to determine which services will be updated, rather than inheriting the servicesOverride. This ensures that all services applied on the nodeSet will be updated during the minor update workflow.

Jira: https://issues.redhat.com/browse/OSPRH-8107

### DIFF
--- a/roles/edpm_update/defaults/main.yml
+++ b/roles/edpm_update/defaults/main.yml
@@ -27,4 +27,4 @@ edpm_update_enable_containers_update: true
 edpm_update_exclude_packages:
   - openvswitch
 
-edpm_update_running_services: "{{ edpm_services_override | default(edpm_services) }}"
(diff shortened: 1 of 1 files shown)


roles/edpm_update/defaults/main.yml
```diff
@@ -27,4 +27,4 @@ edpm_update_enable_containers_update: true
 edpm_update_exclude_packages:
   - openvswitch
 
-edpm_update_running_services: "{{ edpm_services_override | default(edpm_services) }}"
+edpm_update_running_services: "{{ edpm_services }}"
```
slagle, line 30: +2 LGTM

fao89 (approved):

jpodivin (approved):

PR #690, opened 2024-07-03
Ensure firewalld service is disabled if present

son-vyas:
configure-os-edpm-deployment-pre-ceph-openstack-edpm pod fails at 'Ensure firewalld service is disabled' task

Fixes:  https://issues.redhat.com/browse/OSPCIX-352

### DIFF
--- a/roles/edpm_nftables/tasks/service-bootstrap.yml
+++ b/roles/edpm_nftables/tasks/service-bootstrap.yml
@@ -17,6 +17,14 @@
 - name: Gather service facts
   ansible.builtin.service_facts:
 
+- name: Ensure firewalld service is disabled, stopped, and masked if it exists
+  ansible.builtin.systemd:
+    name: firewalld
+    enabled: false
+    state: stopped
+    masked: true
+  when: ansible_facts.services["firewalld.service"] is defined
+
 - name: Switch firewall engine
   become: true
   when:
@@ -35,13 +43,6 @@
       when:
         - ansible_facts.services[item] is defined
         - ansible_facts.services[item]["status"] != "not-found"
-
-    - name: Ensure firewalld service is disabled
-      ansible.builtin.systemd:
-        name: firewalld
-        enabled: false
-        state: stopped
-        masked: true
 
     - name: Ensure nftables service is enabled and running
       ansible.builtin.systemd:


roles/edpm_nftables/tasks/service-bootstrap.yml
```diff
@@ -17,6 +17,25 @@
 - name: Gather service facts
   ansible.builtin.service_facts:
 
+- name: Check if firewalld service exists
+  ansible.builtin.command: systemctl status firewalld
```
pablintino, line 21: Direct usages of `ansible.builtin.command` are discouraged if a module/plugin is available for the same purpose.
In this particular case, the information you want to fetch, if `firewalld` is present or not, is already available thanks to https://github.com/openstack-k8s-operators/edpm-ansible/pull/690/files#diff-2c5bbb52f169a24c08c751de07b1e24779125b4e3712c542f5b0cf50f298ab6aR18
Leverage those facts for this purpose like this:

```yaml
    - name: Ensure firewalld service is disabled
      when:
        - ansible_facts.services["firewalld.service"] is defined
      ansible.builtin.systemd: <....>
        
```

roles/edpm_nftables/tasks/service-bootstrap.yml
```diff
@@ -17,6 +17,25 @@
...
+    enabled: false
+    state: stopped
+    masked: true
+  when: firewalld_status.rc == 0
+  register: manage_firewalld
+
+- name: Ensure deployment fails if firewalld management fails
+  fail:
+    msg: "Failed to manage firewalld service."
+  when: manage_firewalld is failed
```
pablintino, line 37: This will continue to fail in hosts that doesn't have firewalld installed, as `systemctl status firewalld` returns a non-zero exit code for unknown services. If the previous comment is considered this task won't be needed.

roles/edpm_nftables/tasks/service-bootstrap.yml
son-vyas: done

roles/edpm_nftables/tasks/service-bootstrap.yml
son-vyas: done

roles/edpm_nftables/tasks/service-bootstrap.yml
```diff
@@ -36,13 +44,6 @@
         - ansible_facts.services[item] is defined
         - ansible_facts.services[item]["status"] != "not-found"
 
-    - name: Ensure firewalld service is disabled
```
pablintino, line 39: I'd prefer to preserve this disable task instead of moving it above to avoid touching firewalld if we are not going to enable nftables.

rlandy:
will submit new change

PR #691, opened 2024-07-03
Disable firewalld only when present

pablintino:
Under some OS firewalld is not present, leading to the disabling task to fail. Avoid disabling it if not present.

### DIFF
--- a/roles/edpm_nftables/tasks/service-bootstrap.yml
+++ b/roles/edpm_nftables/tasks/service-bootstrap.yml
@@ -42,6 +42,8 @@
         enabled: false
         state: stopped
         masked: true
+      when:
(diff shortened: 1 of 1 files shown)


fao89 (approved):
thank you!

slagle (approved):

roles/edpm_nftables/tasks/service-bootstrap.yml
```diff
@@ -42,6 +42,8 @@
         enabled: false
         state: stopped
         masked: true
+      when:
+        - ansible_facts.services["firewalld.service"] is defined
```
karelyatin, line 46: let's also add ansible_facts.services[item]["status"] != "not-found", i recall weird cases without that

PR #694, opened 2024-07-11
Fixes mount issue when using NFS for nova

snecklifter:
https://issues.redhat.com/browse/OSPCIX-194

### DIFF
--- a/roles/edpm_nova/templates/nova_compute.json.j2
+++ b/roles/edpm_nova/templates/nova_compute.json.j2
@@ -21,7 +21,7 @@
         "/var/log/containers/nova:/var/log/containers/nova",
         "/var/lib/libvirt:/var/lib/libvirt",
         "/run/libvirt:/run/libvirt:shared",
-        "/var/lib/nova:/var/lib/nova:shared,z",
+        "/var/lib/nova:/var/lib/nova:shared",
         "/var/lib/iscsi:/var/lib/iscsi:z",
         "/etc/multipath:/etc/multipath:z",
         "/etc/multipath.conf:/etc/multipath.conf:ro",


roles/edpm_nova/templates/nova_compute.json.j2
```diff
@@ -21,7 +21,7 @@
         "/var/log/containers/nova:/var/log/containers/nova",
         "/var/lib/libvirt:/var/lib/libvirt",
         "/run/libvirt:/run/libvirt:shared",
-        "/var/lib/nova:/var/lib/nova:shared,z",
+        "/var/lib/nova:/var/lib/nova:shared",
```
SeanMooney, line 24: for context we were chattign about this a little on slack.

i think we may be able to proceed with this change
as it reverts to the tripleo bevhiaor

https://github.com/openstack-archive/tripleo-heat-templates/blob/stable/wallaby/deployment/nova/nova-compute-container-puppet.yaml#L1431

but its also only required for running nova on nfs which is a post ga feature.

now I'm not sure we want to have to deal with relableing the directory after ga or any issues that might introduce so I'm slightly inclined to merge this before ga to minimise risk.

however we need to ensure there are no other fallout form this first

bogdando: we need to retain this tripleo feature without bringing in a regression after GA.
So please let's merge, then deal to fallouts as it goes

bogdando (approved):
/lgtm

roles/edpm_nova/templates/nova_compute.json.j2
gibizer: We can land this patch as edpm-ansible is also branched for GA already. So this commit will go to main and will be release post GA.
However if we want this in GA then we need:
* a Jira ticket that is an approved blocker of GA
* we need to cherry-pick this fix to the 18.0.0-proposed branch
* after the fix landed in 18.0.0-proposed make sure that the new edpm-ansible dependency is picked up by the upstream / downstream build.

SeanMooney: i agree with gibi.
we can proceed with merging this but if we want it in GA we need to follow the blockers process

this will not change the support status of /var/lib/nova on nfs

that will still be unsupported in GA and will be part fo a follow up release effectively as a tech preview.

but i also agree with bogdan that we don't want to do this after ga if we can avoid it

ill approve this now and file an issue tracker for ga and we can see if we get agreement to treat it as a blocker seperatly.

SeanMooney (approved):
this has passed molcelue and tempst testing

the main concern was "will this regress seliux itneractions with host installed libvirt" 

based on the tempest results

even though they are very limited we are able to boot vms correctly

https://logserver.rdoproject.org/94/694/3e7f3a86ca8795143c86232ee48d67ce9696f7d5/github-check/podified-multinode-edpm-deployment-crc/9c23db2/controller/ci-framework-data/tests/test_operator/tempest-tests/stestr_results.html

based on that there does not appear to be any regression in libvirt functionality so I'm approving this change for main

SeanMooney:
this is the bug for the backport https://issues.redhat.com/browse/OSPRH-8576
if we agree that its a blocker we can proceed with that once apporved

gibizer:
> this is the bug for the backport https://issues.redhat.com/browse/OSPRH-8576 if we agree that its a blocker we can proceed with that once apporved

Thanks @SeanMooney . Btw once the blocker is approved we can use the bot to do the cherry-pick via `/cherry-pick 18.0.0-proposed`

PR #695, opened 2024-07-12
tripleo_cleanup: Fix a task that was always skipped and it shouldn't be

jistr:
Apparently the 'when' condition on a block is not evaluated only before the block begins executing, but it is evaluated for each task separately. That means that if the fact that is used in the 'when' condition changes during the block execution, it affects whether the following tasks in the block will execute or not.

In our case, the "Filter for tripleo services" task set the tripleo_services fact, and due to that the following "Filter services in skip list" task never executed because the "when" condition on the block says to execute only if tripleo_services is empty.

I've pulled the two set_fact tasks into a single one to get around this issue, since i didn't want to pollute the facts list with a temporary fact. (Facts can't be undefined.)

Resolves: [OSPRH-8584](https://issues.redhat.com/browse/OSPRH-8584)

### DIFF
--- a/roles/edpm_tripleo_cleanup/defaults/main.yml
+++ b/roles/edpm_tripleo_cleanup/defaults/main.yml
@@ -24,16 +24,16 @@ edpm_remove_tripleo_unit_files: true
 
 # Keep services listed
 edpm_service_removal_skip_list:
-  - tripleo_swift_account_auditor
-  - tripleo_swift_account_reaper
-  - tripleo_swift_account_replicator
-  - tripleo_swift_account_server
-  - tripleo_swift_container_auditor
-  - tripleo_swift_container_replicator
-  - tripleo_swift_container_server
-  - tripleo_swift_container_updater
-  - tripleo_swift_object_auditor
-  - tripleo_swift_object_expirer
-  - tripleo_swift_object_replicator
-  - tripleo_swift_object_server
-  - tripleo_swift_object_updater
+  - tripleo_swift_account_auditor.service
+  - tripleo_swift_account_reaper.service
+  - tripleo_swift_account_replicator.service
+  - tripleo_swift_account_server.service
+  - tripleo_swift_container_auditor.service
+  - tripleo_swift_container_replicator.service
+  - tripleo_swift_container_server.service
+  - tripleo_swift_container_updater.service
+  - tripleo_swift_object_auditor.service
+  - tripleo_swift_object_expirer.service
+  - tripleo_swift_object_replicator.service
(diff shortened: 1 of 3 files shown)


odyssey4me (approved):

fao89 (approved):

jistr:
Pending testing with testproject - this is only caught by tempest

jpodivin (approved):

jpodivin:
recheck

xek (approved):

jpodivin (commented):
/hold 

There seems to be a problem with swift filter.

marios (commented):
lgtm, can we also include the edpm_tripleo_cleanup role in https://github.com/openstack-k8s-operators/edpm-ansible/blob/10eb44961c04c19aabcc81225ca54f38035a862f/zuul.d/projects.yaml#L24 

could be followup as i think this is on the way to merge

marios:
> lgtm, can we also include the edpm_tripleo_cleanup role in
> 
> https://github.com/openstack-k8s-operators/edpm-ansible/blob/10eb44961c04c19aabcc81225ca54f38035a862f/zuul.d/projects.yaml#L24
> 
> could be followup as i think this is on the way to merge


a i just saw it has the do not merge label still so ... could update here goign to ping jistr now

marios:
> > lgtm, can we also include the edpm_tripleo_cleanup role in
> > https://github.com/openstack-k8s-operators/edpm-ansible/blob/10eb44961c04c19aabcc81225ca54f38035a862f/zuul.d/projects.yaml#L24
> > 
> > could be followup as i think this is on the way to merge
> 
> ah, i just saw it has the do not merge label still so ... could update here goign to ping jistr now


so, actually if we do this, we would get the 'ceph job' which is fine, but it wouldn't catch the issue being fixed here (which was hitting the non ceph job). Still though I think it is worth running the job on that role since it is being consumed by the adoption tests

jpodivin (approved):

jistr:
Forgot to check the logs as we planned... doing so now

jistr:
https://logserver.rdoproject.org/95/695/bc8c70075366ff31bbe868bdbc6fdc39eeafe0f5/github-check/cifmw-data-plane-adoption-osp-17-to-extracted-crc/8e65bc4/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/tripleo-cleanup-tripleo-cleanup-openstack-lnxhv/logs/openstackansibleee.log

Looks like the only stopped Swift service now is `tripleo_swift_proxy.service`, the storage services were correctly skipped.

PR #696, opened 2024-07-12
Change pattern for detecting pre-adopted versus edpm nodes

ciecierski:
Change edpm_reboot_old_tripleo_node_config_dir to check for tripleo files in /var/lib/config-data/puppet-generated instead of /var/lib/config-data. Path /var/lib/config-data is still being used in edpm for storing configs whereas puppet-generated is suppose to be used only in 17/wallaby. This will make sure for greenfield auto mode is back working, but reboot is skipped on tripleo nodes during adoption.

Resolves: [OSPRH-8415](https://issues.redhat.com//browse/OSPRH-8415)

### DIFF
--- a/roles/edpm_reboot/molecule/pre-adoption/converge.yml
+++ b/roles/edpm_reboot/molecule/pre-adoption/converge.yml
@@ -1,0 +1,35 @@
+---
+# Copyright 2024 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+- name: Converge
+  hosts: all
+  gather_facts: true
+  become: true
+  tasks:
+    - name: Create reboot_required directory
+      ansible.builtin.file:
+        path: /var/lib/openstack/reboot_required/
+        state: directory
+        mode: '0755'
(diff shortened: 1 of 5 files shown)


bogdando (changes requested):
In tripleo-ansible there are still roles that rely on /var/lib/config-data/ansible-generated.

Those might conflict with edpm-ansible, and need to be discovered as "old" ones instead. The only option I see - never use /var/lib/config-data/ansible-generated in edpm-ansible

ciecierski:
> In tripleo-ansible there are still roles that rely on /var/lib/config-data/ansible-generated.
> 
> Those might conflict with edpm-ansible, and need to be discovered as "old" ones instead. The only option I see - never use /var/lib/config-data/ansible-generated in edpm-ansible

I switched to **/var/lib/config-data/puppet-generated**. This is not used in edpm for 100%.
For ansible-generated true, still some edpm-ansible roles use it and some /var/lib/edpm-config

roles/edpm_reboot/molecule/pre-adoption/converge.yml
```diff
@@ -0,0 +1,35 @@
+---
+# Copyright 2023 Red Hat, Inc.
```
jpodivin, line 2: s/2023/2024

jpodivin (approved):
I have only an exceedingly minor nitpick. The documentation should include a note about the pattern for future reference. Other than that this has my +2. Great job!

bogdando:
> I switched to **/var/lib/config-data/puppet-generated**. This is not used in edpm for 100%.
> For ansible-generated true, still some edpm-ansible roles use it and some /var/lib/edpm-config

OK, this should work as long there would be no /var/lib/config-data/ansible-generated w/o puppet-generated in the legacy environment, and to my knowledge, that is always true. So indeed, we can ignore the fact that /var/lib/config-data/ansible-generated may be used by osp AND rhoso for configs rendering

bogdando (approved):
/lgtm

jpodivin (approved):

jistr (approved):

ciecierski:
recheck

PR #701, opened 2024-07-17
Stylistic adjustments for modules, plugins and filters

jpodivin:
We have accumulated quite a bit of idiosyncracies in our python code. This doesn't resolve all of them by any means, but it does take care of some of the most egregious ones. Dead code, useless imports, duplication and typos are all included. 

None of these changes should alter the way our code works. I've tried to avoid veering into full on refactor with this. That being said there is one minor change of `osp.edpm.haskey` filter API. However the filter in question is only used by one role, so it shouldn't have any adverse effects. 

Normally I would squash this. But since most of these are rather disjoint issues, I can't help but feel that it would make a mess of blame for whoever will follow. So I'm keeping the commits separate, unless people insist on squashing.

### DIFF
--- a/plugins/action/container_systemd.py
+++ b/plugins/action/container_systemd.py
@@ -15,7 +15,6 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
-from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 import copy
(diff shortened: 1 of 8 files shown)


jpodivin:
recheck

```
The connection to the server oauth-openshift.apps-crc.testing was refused - did you specify the right host or port?
Can not login to the CRC cluster. Sleeping...
Checking login to the cluster 18...
WARNING: Using insecure TLS client config. Setting this option is not supported!

rebtoor:
@jpodivin PR is more than fine for me but i'm testing it downstream just to be extra sure about its impact. As long as that test will pass, i'll give my `lgtm`.

slagle (approved):

PR #707, opened 2024-07-24
Changing cert permissions to comply with changed virt-pki-validate rules

jpodivin:
libvirt 10.5.0 re-implemented virt-pki-validate in c and changed some of the required values
for the certs. this pr disable the invocation of virt-pki-validate temporarily until we can address the warnings  post ga.

### DIFF
--- a/roles/edpm_libvirt/molecule/default/verify.yml
+++ b/roles/edpm_libvirt/molecule/default/verify.yml
@@ -48,14 +48,17 @@
         key: libvirt
       register: libvirt_user
 
-    - name: validate tls with virt-pki-validate
-      become: true
-      ansible.builtin.shell: "virt-pki-validate"
-      register: virt_pki_validate
-    - name: Assert that virt-pki-validate returns no errors
-      assert:
-        that:
-          - "virt_pki_validate.rc == 0"
+    # TODO: Temporarily disabled until we can reconcile new rules
+    #       in the tool with our environment
+    # - name: validate tls with virt-pki-validate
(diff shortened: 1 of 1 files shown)


roles/edpm_libvirt/molecule/default/prepare.yml
```diff
@@ -176,6 +176,7 @@
       community.crypto.openssl_csr_pipe:
         privatekey_path: "{{ edpm_libvirt_tls_cert_src_dir }}/tls.key"
         organization_name: fake-corp
+        common_name: "{{ inventory_hostname }}"
```
SeanMooney, line 179: this shoudl work as i used it in the subject_alt_name previously
to pass the check before they reworte it.

if not we might need to use hostvars or regestert the hostname using the hostname command.

assuming this is happy in ci however lets keep it simple.

SeanMooney: ok this seam to fail agin 

SERVER: Checking cert hostname match                                         : WARN (Certificate /etc/pki/libvirt/servercert.pem owner does not match the hostname np0004809514.novalocal)

so it looks like it wants the fqdn not the hostname

SeanMooney:
check-rdo

jpodivin:
recheck

SeanMooney (approved):
we need to correct the zuul issue but once that is fixed i belive we can proceed with this for ga and correct this gap after

jpodivin:
recheck

PR #711, opened 2024-07-26
Fix mounting issue for hgpages and test

mrkisaolamb:


slagle (approved):

roles/edpm_extra_mounts/molecule/default/verify.yml
```diff
@@ -13,3 +13,15 @@
...
+      register: mount_check
+      ignore_errors: yes
+      with_items:
+        - "/dev/hugepages1G"
+        - "/dev/hugepages2MB"
+    - name: Fail if any directory is not mounted
+      fail:
+        msg: "Directory {{ item.item }} is not mounted"
+      when: item.rc != 0
+      with_items: "{{ mount_check.results }}"
```
rebtoor, line 27:
```suggestion
    - name: Check if directories are mounted
      ansible.builtin.assert:
        that:
          - "{{ item is mount }}"
        fail_msg: "Directory {{ item }} isn't mounted!"
      loop:
        - "/dev/hugepages1G"
        - "/dev/hugepages2MB"
```

A simpler but elegant solution with a single task using [mount](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/mount_test.html) ansible test.

I tried locally with this playbook:
```
---
- hosts: localhost
  vars:
    mountpoints:
      - /dev/pts
      - /dev/shm
      - /dev/nonexistent
  tasks:
    - name: Check if devices are mounted
      ansible.builtin.assert:
        that:
          - "{{ item is mount }}"
        fail_msg: "Directory {{ item }} isn't mounted!"
      loop: "{{ mountpoints }}"
```

and here's the result:

```
TASK [Check if devices are mounted] **********************************************************************************************************************************************************************************************************************************************
[WARNING]: conditional statements should not include jinja2 templating delimiters such as {{ }} or {% %}. Found: {{ item is mount }}
ok: [localhost] => (item=/dev/pts) => {
    "ansible_loop_var": "item",
    "changed": false,
    "item": "/dev/pts",
    "msg": "All assertions passed"
}
ok: [localhost] => (item=/dev/shm) => {
    "ansible_loop_var": "item",
    "changed": false,
    "item": "/dev/shm",
    "msg": "All assertions passed"
}
failed: [localhost] (item=/dev/nonexistent) => {
    "ansible_loop_var": "item",
    "assertion": "{{ item is mount }}",
    "changed": false,
    "evaluated_to": false,
    "item": "/dev/nonexistent",
    "msg": "Directory /dev/nonexistent isn't mounted!"
}
```

rebtoor (changes requested):

rebtoor (approved):

roles/edpm_extra_mounts/molecule/default/verify.yml
rebtoor: ack, i'll try to understand why, for now we can merge this PR as is. Thanks a lot!

PR #712, opened 2024-07-26
Explicitly setting special permissions bits

jpodivin:
Minor alteration of existing code, normalizing the way we specify permissions.

Resolves: https://issues.redhat.com/browse/OSPRH-8981


jpodivin:
recheck

rebtoor (approved):

PR #715, opened 2024-07-29
Restoring virt-pki-validate check in molecule test

jpodivin:
Resolves: [OSPCIX-393](https://issues.redhat.com//browse/OSPCIX-393) 

Permissions for directories are now set according to new virt-pki-validate recommendations. Instead of checking return code of the tool, we are now matching on output. This is because even a single warning changes rc to 1.

### DIFF
--- a/roles/edpm_libvirt/tasks/configure.yml
+++ b/roles/edpm_libvirt/tasks/configure.yml
@@ -22,9 +22,9 @@
   loop:
     - {"path": "/etc/tmpfiles.d/", "owner": "root", "group": "root"}
     - {"path": "/var/lib/edpm-config/firewall", "owner": "root", "group": "root"}
-    - {"path": "/etc/pki/libvirt", "owner": "root", "group": "root"}
-    - {"path": "/etc/pki/libvirt/private", "owner": "root", "group": "root"}
-    - {"path": "/etc/pki/CA", "owner": "root", "group": "root"}
+    - {"path": "/etc/pki/libvirt", "owner": "root", "group": "root", "mode": "0755"}
+    - {"path": "/etc/pki/libvirt/private", "owner": "root", "group": "root", "mode": "0755"}
+    - {"path": "/etc/pki/CA", "owner": "root", "group": "root", "mode": "0755"}
     - {"path": "/etc/pki/qemu", "owner": "root", "group": "qemu"}
 
 - name: Render libvirt config files
@@ -144,7 +144,7 @@
   become: true
   loop:
     - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.crt", "dest": "/etc/pki/libvirt/servercert.pem"}
-    - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.key", "dest": "/etc/pki/libvirt/private/serverkey.pem"}
+    - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.key", "dest": "/etc/pki/libvirt/private/serverkey.pem", "mode": "0600"}
     - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.crt", "dest": "/etc/pki/libvirt/clientcert.pem"}
     - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.key", "dest": "/etc/pki/libvirt/private/clientkey.pem"}
     - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/ca.crt", "dest": "/etc/pki/CA/cacert.pem"}
@@ -152,7 +152,7 @@
     src: "{{ item.src }}"
     dest: "{{ item.dest }}"
     remote_src: true
-    mode: "0644"
+    mode: "{{ item.mode | default('0644') }}"
     owner: "root"
     group: "root"
   when: edpm_libvirt_tls_certs_enabled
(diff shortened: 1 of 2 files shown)


roles/edpm_libvirt/tasks/configure.yml
```diff
@@ -22,9 +22,9 @@
   loop:
     - {"path": "/etc/tmpfiles.d/", "owner": "root", "group": "root"}
     - {"path": "/var/lib/edpm-config/firewall", "owner": "root", "group": "root"}
-    - {"path": "/etc/pki/libvirt", "owner": "root", "group": "root"}
-    - {"path": "/etc/pki/libvirt/private", "owner": "root", "group": "root"}
-    - {"path": "/etc/pki/CA", "owner": "root", "group": "root"}
+    - {"path": "/etc/pki/libvirt", "owner": "root", "group": "root", "mode": "0755"}
+    - {"path": "/etc/pki/libvirt/private", "owner": "root", "group": "root", "mode": "0755"}
+    - {"path": "/etc/pki/CA", "owner": "root", "group": "root", "mode": "0755"}
```
gibizer, line 27: allowing nonprivileged user to see the directory is OK.

roles/edpm_libvirt/tasks/configure.yml
```diff
@@ -144,15 +144,15 @@
   become: true
   loop:
     - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.crt", "dest": "/etc/pki/libvirt/servercert.pem"}
-    - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.key", "dest": "/etc/pki/libvirt/private/serverkey.pem"}
+    - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.key", "dest": "/etc/pki/libvirt/private/serverkey.pem", "mode": "0600"}
```
gibizer, line 147: only root will see the pivate key. Good.

roles/edpm_libvirt/tasks/configure.yml
```diff
@@ -144,15 +144,15 @@
...
     - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.crt", "dest": "/etc/pki/libvirt/clientcert.pem"}
     - {"src": "{{ edpm_libvirt_tls_cert_src_dir }}/tls.key", "dest": "/etc/pki/libvirt/private/clientkey.pem"}
```
gibizer, line 149: do we need to set 600 on this private key too?

roles/edpm_libvirt/tasks/configure.yml
kashyapc: Yeah, indeed. These permissions are newer recommendations.

Just as some background context: the old implementation of `virt-pki-validate` was not correctly validating the permissions that it was recommending, so if run as root, it was not reporting these permission fails.

The newer `virt-pki-validate` fixes that, hence the errors before this change.

roles/edpm_libvirt/tasks/configure.yml
jpodivin: no. That is not virt-pki-validate recommendation.

kashyapc (approved):
Looks good to me, FWIW.  At least from the PoV of implementing the recommendations of `virt-pki-validate`.

roles/edpm_libvirt/tasks/configure.yml
gibizer: OK. I won't block on this. But lets wait for @SeanMooney to approve it.

SeanMooney: access to the host is privileged. so to access this key you would already have access to a privileged security context (i.e. a shell on the host)
I agree that 600 would minimally improve the security posture but I hesitate to deviate from the tested defaults recommended by the libvirt project.

SeanMooney (approved):
This pr aligns the secirty posture with that recommended by the latest version fo the virt-pki-validate tools not that it has been reimplemnted in c with additional checks.

looking at the job output 
https://softwarefactory-project.io/zuul/t/rdoproject.org/build/734fd7607b034334962986c7d18623eb/log/job-output.txt#1452

we do still have the warning for the cert hostname but that is purely an artifact fo how we generate certs in ci and is not a blocker to moving this forward.

we can address that in a followup later if desired once we fully understand how to generate it such that it works in ci and locally via the delegated and vagrat molecule providers.

roles/edpm_libvirt/tasks/configure.yml
```diff
@@ -22,9 +22,9 @@
   loop:
     - {"path": "/etc/tmpfiles.d/", "owner": "root", "group": "root"}
     - {"path": "/var/lib/edpm-config/firewall", "owner": "root", "group": "root"}
-    - {"path": "/etc/pki/libvirt", "owner": "root", "group": "root"}
-    - {"path": "/etc/pki/libvirt/private", "owner": "root", "group": "root"}
-    - {"path": "/etc/pki/CA", "owner": "root", "group": "root"}
+    - {"path": "/etc/pki/libvirt", "owner": "root", "group": "root", "mode": "0755"}
+    - {"path": "/etc/pki/libvirt/private", "owner": "root", "group": "root", "mode": "0755"}
```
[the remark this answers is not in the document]
bogdando: given that we want ``/etc/pki/libvirt/private/serverkey.pem", "mode": "0600"``, this should be 0750?

PR #717, opened 2024-07-30
Variable default defined incorrect

Jaganathancse:
This changes to fix the incorrect variable's
used to define default value.


Jaganathancse:
recheck

Jaganathancse:
Need to analyze more and take this changes during adoption.

PR #719, opened 2024-08-05
ovn-bgp-agent sub-dirs created with non-root user

eduolivares:
The following PR changed the permissions and ownership for directories created from many roles:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/683

This commit removes `become: true` from the task that creates ovn-bgp-agent configuration directories.


eduolivares:
@jpodivin, can you please take a look at this PR?
I have verified it on my BGP downstream environment.
Thanks!

roles/edpm_ovn_bgp_agent/tasks/configure.yml
```diff
@@ -52,13 +52,15 @@
 - name: Add OVS Manager
   block:
     - name: Check if OVS Manager already exists
+      become: true
```
jpodivin, line 55: Does this need to run under root? And how is this relevant to the dir ownership?

roles/edpm_ovn_bgp_agent/tasks/configure.yml
```diff
@@ -52,13 +52,15 @@
...
       ansible.builtin.shell: |
         set -o pipefail
         ovs-vsctl show | grep -q "Manager"
       register: ovs_manager_configured
       failed_when: false
       changed_when: true
     - name: Add OVS Manager if not exists
+      become: true
```
jpodivin, line 63: ditto

roles/edpm_ovn_bgp_agent/tasks/configure.yml
eduolivares: this needs `become: true` like PR 683 did here, right?
https://github.com/openstack-k8s-operators/edpm-ansible/pull/683/files#diff-07e04cba7b71b24ec5e57638feeb63e9b4f1b09e43c56aa73d8e132c64a887f7
All tasks executing ovs-vsctl commands were modified to use `become: true`

jpodivin (commented):
LGTM

jpodivin (approved):

PR #720, opened 2024-08-05
[WIP] Fail edpm_network_config when there are schema errors

vcandapp:
Currently when there are schema errors in os-net-config, the logs are updated and no failure happens.
Updating this behaviour so that user can fix errors and re-run role


plugins/modules/edpm_os_net_config.py
```diff
@@ -207,7 +209,7 @@ def main():
     detailed_exit_codes = args['detailed_exit_codes']
     safe_defaults = args['safe_defaults']
     use_nmsate = args['use_nmstate']
-    return_codes = [0]
+    return_codes = [0,1]
```
Jaganathancse, line 212: i think no need add 1 here. only 0 is for success to check failed or not.

karthiksundaravel (commented):
lgtm

vcandapp:
@karthiksundaravel / @Jaganathancse , updated this PR for better tracking. This PR will have fix only for the no-auto-default. Will created a new PR for "--exit-on-validation-error"

plugins/modules/edpm_os_net_config.py
vcandapp: Have removed this file now

roles/edpm_network_config/tasks/main.yml
```diff
@@ -36,7 +36,25 @@
     - name: Load system-roles.network tasks [nmstate]
       ansible.builtin.include_role:
         name: "{{ lookup('ansible.builtin.env', 'EDPM_SYSTEMROLES', default='fedora.linux_system_roles') + '.network' }}"
-- name: Load edpm_network_config tasks [os-net-config]
-  ansible.builtin.include_tasks:
-    file: network_config.yml
-  when: edpm_network_config_tool == 'os-net-config'
+
+- name: Disable auto-configuration of all interfaces by NetworkManager
+  when: ( edpm_bootstrap_network_service == 'NetworkManager' ) and
```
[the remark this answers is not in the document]
vcandapp: Yes - the other scenario will be to use nmstate tool

slagle (commented):
Can you update the PR description to match the commits? It seems to be about disabling NetworkManager when os-net-config is used. 
Please drop the WIP and the merge commits in the PR as well if it's ready for review.

vcandapp:
This is replaced by https://github.com/openstack-k8s-operators/edpm-ansible/pull/840

PR #723, opened 2024-08-14
Update telemetry role to deploy Kepler

yadneshk:
Depends-On: openstack-k8s-operators/openstack-operator#1034

### DIFF
--- a/.github/workflows/molecule.yaml
+++ b/.github/workflows/molecule.yaml
@@ -42,7 +42,6 @@ jobs:
         - edpm_sshd
         - edpm_ssh_known_hosts
         - edpm_timezone
-        - edpm_telemetry
         - edpm_telemetry_logging
         - edpm_update
         - edpm_users
--- a/roles/edpm_telemetry/defaults/main.yml
+++ b/roles/edpm_telemetry/defaults/main.yml
@@ -42,3 +42,5 @@ edpm_telemetry_image_download_retries: 5
 edpm_telemetry_old_tripleo_compute_sevices:
   - tripleo_ceilometer_agent_compute.service
   - tripleo_ceilometer_agent_ipmi.service
+# Image to use for kepler
+edpm_telemetry_kepler_image: "quay.io/sustainable_computing_io/kepler:release-0.7.11"
--- a/roles/edpm_telemetry/molecule/default/verify.yml
+++ b/roles/edpm_telemetry/molecule/default/verify.yml
@@ -22,13 +22,15 @@
         - "ceilometer_agent_compute"
         - "ceilometer_agent_ipmi"
         - "node_exporter"
+        - "kepler"
 
     - name: ensure systemd services are defined and functional
       ansible.builtin.include_tasks: "{{test_helper_dir}}/verify_systemd_unit.yaml"
       loop:
-        - {"name": "edpm_ceilometer_agent_compute.service"}
-        - {"name": "edpm_ceilometer_agent_ipmi.service"}
-        - {"name": "edpm_node_exporter.service"}
+        - { "name": "edpm_ceilometer_agent_compute.service" }
+        - { "name": "edpm_ceilometer_agent_ipmi.service" }
+        - { "name": "edpm_node_exporter.service" }
+        - { "name": "edpm_kepler.service" }
 
     - name: ensure kolla_set_configs copied the expected files and started the correct binary
       become: true
@@ -39,7 +41,6 @@
         - "Copying /var/lib/openstack/config/polling.yaml to /etc/ceilometer/polling.yaml"
         - "/usr/bin/ceilometer-polling --polling-namespaces compute --logfile /dev/stdout"
         - "Copying /var/lib/openstack/config/ceilometer-host-specific.conf to /etc/ceilometer/ceilometer.conf.d/02-ceilometer-host-specific.conf"
-
 
     - name: ensure kolla_set_configs copied the expected files and started the correct binary
       become: true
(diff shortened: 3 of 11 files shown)


slagle (commented):
we can't install by downloading RPM's from github directly. maybe as an optional thing, but this would be done for every deployment with telemetry, which is currently included by default. as this gets used downstream as well, we can't install RPM's this way.

roles/edpm_telemetry/molecule/default/verify.yml
```diff
@@ -59,6 +61,13 @@
         - "Starting node_exporter"
         - "TLS is enabled"
 
+    - name: ensure that Kepler is running
+      become: true
+      ansible.builtin.shell: |
+        podman logs kepler 2>&1 | grep "{{item}}" > /dev/null
+      loop:
+        - "Started Kepler"
+
```
rebtoor, line 70: Can we just check if the container is running? Do we expect a running container that doesn't work?

roles/edpm_telemetry/defaults/main.yml
```diff
@@ -42,3 +42,7 @@ edpm_telemetry_image_download_retries: 5
 edpm_telemetry_old_tripleo_compute_sevices:
   - tripleo_ceilometer_agent_compute.service
   - tripleo_ceilometer_agent_ipmi.service
+# Release of Kepler to be installed (Browse all releases here https://github.com/sustainable-computing-io/kepler/releases)
+edpm_telemetry_kepler_release: "0.7.11"
+# URL that provides Kepler rpm to install on compute nodes
+edpm_telemetry_kepler_image: "quay.io/sustainable_computing_io/kepler:release-{{ edpm_telemetry_kepler_release }}"
```
slagle, line 48: we can't just run containers from quay.io in every deployment of our product. This would need to be managed how every other container image is managed in edpm-ansible. Unless this discussion has already happened somewhere, we need to hold off on this PR until this is resolved.

slagle (dismissed):

roles/edpm_telemetry/defaults/main.yml
yadneshk: I see nova-compute image is being pulled from quay here https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_nova/defaults/main.yml#L33 

Node exporter follows the same approach for deployment https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_telemetry/defaults/main.yml#L26 and it pulls image from the `prometheus` repository.

Seems almost all containers are being pulled from quay.

slagle: If it's meant to be handled like the other images, then there needs to be a related PR for openstack-operator that adds support the image with a $RELATED_IMAGE* var, and handling in OpenStackVersion. You should be able to review what's there for the node exporter image and copy that same pattern.

yadneshk:
recheck

.github/workflows/molecule.yaml
```diff
@@ -42,7 +42,6 @@ jobs:
         - edpm_sshd
         - edpm_ssh_known_hosts
         - edpm_timezone
-        - edpm_telemetry
```
vyzigold, line 45: This is intentional?

vyzigold: Found a comment on slack about this. It is intentional.

yadneshk: Yes, this would exclude edpm_telemetry role from being testing in github workflows. There's an issue while doing this. While running Kepler container inside the ubi container it reported this error 
```
error removing memlock: failed to set memlock rlimit: operation not permitted
```
So I have moved testing this role into VM by making change in `zuul.d/jobs.yaml`

zuul.d/projects.yaml
```diff
@@ -52,3 +53,4 @@
               - ^roles/edpm_neutron_metadata/*
               - ^roles/edpm_pre_adoption_validation/*
               - ^roles/edpm_tripleo_cleanup/*
+              - ^roles/edpm_telmetry/*
```
jlarriba, line 56: There is a typo here: "edpm_telmetry"

yadneshk:
recheck

Error during connting VM to fip
```
TASK [dataplane_adoption : verify connectivity to the existing test VM instance using Floating IP] ***
fatal: [localhost]: FAILED! => {"changed": true, "cmd": "ping -c4 192.168.122.20\n", "delta": "0:00:13.063327", "end": "2024-09-03 10:53:16.858523", "msg": "non-zero return code", "rc": 1, "start": "2024-09-03 10:53:03.795196", "stderr": "", "stderr_lines": [], "stdout": "PING 192.168.122.20 (192.168.122.20) 56(84) bytes of data.\n\n--- 192.168.122.20 ping statistics ---\n4 packets transmitted, 0 received, 100% packet loss, time 3053ms", "stdout_lines": ["PING 192.168.122.20 (192.168.122.20) 56(84) bytes of data.", "", "--- 192.168.122.20 ping statistics ---", "4 packets transmitted, 0 received, 100% packet loss, time 3053ms"]}

PLAY RECAP *********************************************************************
localhost                  : ok=192  changed=155  unreachable=0    failed=1    skipped=23   rescued=0    ignored=0 
```

csibbitt (approved):

roles/edpm_telemetry/molecule/default/verify.yml
yadneshk: No, we don't expect such a situation. I have removed this test.

jlarriba:
recheck

roles/edpm_telemetry/defaults/main.yml
jlarriba: @slagle https://github.com/openstack-k8s-operators/openstack-operator/pull/1034

jpodivin (approved):

PR #725, opened 2024-08-14
Removing links to RHEL docs

jpodivin:
The links were broken. Furthermore, they were only useful if you were running a RHEL system.

General direction to set up necessary firewall rules is sufficient for most system administrators to follow.


jpodivin:
Cool, then we don't need this. Although I would prefer not use any links at all, since these may get broken as well.

PR #731, opened 2024-08-21
Releasing version constraint to allow versions 2.15.0

jpodivin:
We can safely move on to the 2.15.* as there were no real breaking changes, with possible exception of `callback_whitelist`. [0]

[0] https://github.com/ansible/ansible/blob/stable-2.15/changelogs/CHANGELOG-v2.15.rst#id34


jpodivin:
recheck

jpodivin:
recheck

node registration failure

slagle:
I don't think the rdo CI failures are related to this patch. Merging this one so we can proceed with other PRs

PR #733, opened 2024-08-21
Execute edpm_bootstrap_command based on the hash

fao89:
closes [OSPRH-8627](https://issues.redhat.com//browse/OSPRH-8627)

### DIFF
--- a/roles/edpm_bootstrap/tasks/bootstrap_command.yml
+++ b/roles/edpm_bootstrap/tasks/bootstrap_command.yml
@@ -1,14 +1,13 @@
 ---
 
-- name: Check if bootstrap command has been executed
-  ansible.builtin.stat:
-    path: "{{ edpm_bootstrap_command_done_file }}"
-  register: edpm_bootstrap_command_executed
+- name: Re-read facts
+  ansible.builtin.setup:
+    filter: ansible_local
 
 - name: Bootstrap command
   when:
     - edpm_bootstrap_command != ""
-    - not edpm_bootstrap_command_executed.stat.exists
+    - ansible_local.edpm is undefined or ansible_local.edpm.bootstrap.hash != edpm_bootstrap_command | hash('sha512')
   become: true
   block:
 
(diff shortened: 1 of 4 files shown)


roles/edpm_bootstrap/tasks/bootstrap_command.yml
```diff
@@ -5,10 +5,18 @@
...
+    - edpm_bootstrap_command_executed.stat.exists
+  ansible.builtin.set_fact:
+    edpm_bootstrap_command_current_hash: "{{ edpm_bootstrap_command | password_hash('sha512') }}"
+    edpm_bootstrap_command_previous_hash: "{{ lookup('ansible.builtin.file', edpm_bootstrap_command_done_file) }}"
+
 - name: Bootstrap command
   when:
     - edpm_bootstrap_command != ""
-    - not edpm_bootstrap_command_executed.stat.exists
+    - not edpm_bootstrap_command_executed.stat.exists or edpm_bootstrap_command_previous_hash != edpm_bootstrap_command_current_hash
```
SeanMooney, line 19: i kind of feel like a better approach si just to remove the conditional execution entirely and always run it.

it really does need to  be idempotent anyway so I'm not sure this extra complexity buy use much give you can already filter which services to run when defining a deployment if you want too.

fao89: it makes sense, @rebtoor what do you think about removing this condition?

rebtoor: @fao89 go on with the removal. I agree with @SeanMooney here.

fao89: done!

slagle: I don't think we can do this, as it's going to break the documented usage as you can't re-run `subscription-manager register` without also using `--force`, which actually creates a whole new system registration.

Asking the user to make sure the command is idempotent isn't my preference. I'd prefer "only run if changed" or "only run if bootstrap file doesn't exist".

fao89: I'll change back to the previous approach

slagle (commented):
this LGTM. i'll hold off on approving until we get other feedback.

roles/edpm_bootstrap/tasks/bootstrap_command.yml
bshephar: My only concern here is the `edpm_bootstrap_command_done_file`, which probably doesn't need to be addressed in this specific PR. But if we're relying on a file that we're randomly dropping on the nodes, and that file not existing will result in deployment failures as James has mentioned. Then I think it's only a matter of time before that becomes a headache for someone. 

In my opinion, it would be better to leverage a custom fact that we can reference. Random files dropped in `/` vs files added to the `/etc/ansible/facts.d` directory is the crux of my concern. 

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_vars_facts.html#facts-d-or-local-facts

So if we were to turn the `edpm_bootstrap_command_done_file` into a fact like `edpm_bootstrap_command_completion` with the same content, then reference that fact rather than reading a local file. That feels like a more Ansible consistent approach and less likely to be interfered with by customers who see a random file that they don't understand.

SeanMooney: honestly I thnk using local fact is more dangerous.

keeping state in a directory we own like /var/lib/openstack/... 

i think is safer then assuming that third-party vendor integration wont clear the fact store.

again i do think dropping this detection is better in the long run even if it requires the docs to be updated as it is unsafe to include any command that isn't idempotent in the bootstrap command so we will need to fix the subscription-manager  doc regardless.

fao89:
recheck

fao89:
recheck

fao89:
Can I have a review?

slagle (approved):

PR #735, opened 2024-08-22
Allow setting Zebra graceful_restart time

lmiccini:
https://docs.frrouting.org/en/latest/zebra.html#cmdoption-zebra-K

"If this option is specified, the graceful restart time is TIME seconds. Zebra, when started, will read in routes. Those routes that Zebra identifies that it was the originator of will be swept in TIME seconds. If no time is specified then we will sweep those routes immediately."


jpodivin (changes requested):
Thanks for the PR. Could you add the variable into arguments of the role?

jpodivin (approved):

PR #736, opened 2024-08-26
Enable profiling on molecule jobs

fao89:


zuul.d/jobs.yaml
```diff
@@ -9,65 +9,91 @@
...
+      ANSIBLE_VERBOSITY: 2
 
 # non-containerized role jobs
 - job:
     name: edpm-ansible-molecule-edpm_bootstrap
     parent: edpm-ansible-molecule-base
     vars:
       TEST_RUN: edpm_bootstrap
+      ANSIBLE_CALLBACKS_ENABLED: profile_tasks
+      ANSIBLE_VERBOSITY: 2
```
jpodivin, line 22: Aren't the vars inherited from the parent job? Or am I reading this wrong.

https://zuul-ci.org/docs/zuul/latest/config/job.html

PR #743, opened 2024-08-30
Change action for config with `restart` in container_manage to `created`

cubeek:
Don't start container if restarted is set

If the container_manage is called with `restart` set, it means the
systemd handles the restart. In such case, we do not want containers to
be started because they will be started later by the systemd.

Resolves: [OSPRH-9812](https://issues.redhat.com//browse/OSPRH-9812)

### DIFF
--- a/plugins/modules/edpm_container_manage.py
+++ b/plugins/modules/edpm_container_manage.py
@@ -317,6 +317,7 @@ class EdpmContainerManage:
             # NOTE(mwhahaha): converation from edpm format to podman as
             # systemd handles this restart config
             opts['restart'] = False
+            opts['state'] = 'created'
         if 'stop_grace_period' in opts:
             opts['stop_timeout'] = opts.pop('stop_grace_period')
 


slagle (commented):
LGTM. the containers are started by the container_systemd task that follows this task, so everything will be started by systemd

cubeek:
I'm going to remove the WIP tag as this seems to be good to go.

plugins/modules/edpm_container_manage.py
```diff
@@ -283,7 +283,7 @@ def exec_container(self, name, config):
     def manage_container(self, name, config):
         opts = {
             'name': name,
-            'state': "started",
+            'state': "created",
```
rabi, line 286: Maybe  we should add  opts['state'] = "created" after L311 instead (i.e when systemd handles the restart) unless all containers in edpm are expected to be managed by systemd?

rabi: This would also make 'start_order` for containers irrelevant, though it seems we concurently manage every container individually with `podman_container_manage` and `start_order` we've in the configs is kind of useless.

cubeek: You mean to override to `created` if the config comes with `restart`?

rabi: Yes,  that is used to filter services for creating systemd units. https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_container_manage/tasks/create.yml#L45

cubeek: I'm not sure I understand. How I read the linked code is that it filters out configs from all_containers_hash based on restart=[some specific values]. But the all_containers_hash map comes from `container_config_data` module: https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_container_manage/tasks/main.yml#L30

As for the `start_order` - it confuses me again how is this related. The goal of this patch is to not start the containers after they are created because systemd will start them through its service files.

rabi: > But the all_containers_hash map comes from container_config_data module

Thats right.. But, let me try to explain it better...

step 1:  `all_containers_hash` comes from `container_config_data` module
step 2:   `edpm_container_manage` module transforms that a container config and calls `PodmanManager` to create or start the containers.
step 3: `container_systemd` action plugin called to create the systemd units by fitering the containers that has restart=something.

What I'm suggesting is that rather than assuming that all containers would require systemd units and hence should only be created (state=created) by PodmanManager, we should only set `state=created` only for those configs that has restart=<always, unless-stopped> as those are the only ones where we would create systemd units and let PodmanManager start the other containers with (state=started).

Hope that calrifies.

rabi: > As for the start_order - it confuses me again how is this related

`start_order` is relevant as container configs are given to PodmanManager sorted by start_order . If you start all of bunch of containers with systemd (and not by PodmanManager)  there would be no order. However, start_order is not important atm (all configs have start_order 0) and I had removed them from all service container configs, as we start all services as standalone services individually one after the other and there is not order expected.

cubeek: I'm still missing why would we need to override. The state is a parameter that can be passed from the caller. This patch just changes it to `created` by default but it still can be overridden by passing "state"="restarted" by the caller and it will be updated on L299. Those who do not want to manage container by the systemd can still pass "restarted". By setting the state after the ops.update() call we would take this functionality away, no?

rabi: > This patch just changes it to created by default in edpm-ansible

We're changing the default and not only for cases where we expect services to be managed by systemd. The default in podman module is `started`[1] and normally services don't specify `state` in config and this the behavior from `tripleo-ansibe`.

> By setting the state after the ops.update() call we would take this functionality away, no?

When we want services to be managed by systemd, passing state='started/stopped' to this module has no meaning and for cleanup (state=absent), there is a separate role 'edpm_container_rm` that does not use this module[2].

If we still want to change this default, we should document that "By default containers are to be managed by systemd and have to provide restart=<>".Also, users have to use `state=started` when using this module if they don't want containers to be managed by systemd".  If by chance user  does not set `restart=<>` and `state=started`, the container would just be created and neither podman nor systemd would start it.

[1] https://github.com/containers/ansible-podman-collections/blob/master/plugins/module_utils/podman/podman_container_lib.py#L22
[2] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_container_rm/tasks/edpm_podman_container_rm.yml#L87-L89

rabi (commented):
/lgtm

rabi:
The commit message would need a change and also there is a merge commit in the PR.

cubeek:
> The commit message would need a change and also there is a merge commit in the PR.

Not sure if you meant really the commit message or the text in the PR. I copied the message from the commit message to be the same.

fao89:
recheck

PR #747, opened 2024-09-09
Check for AnsibleUnicode or str for edpm_kernel_args

slagle:
The type of edpm_kernel_args could be a str, so both types need to be
checked, otherwise the whole block is skipped.

Jira: [OSPRH-10007](https://issues.redhat.com//browse/OSPRH-10007)


bshephar (approved):
Nice find. lgtm

PR #749, opened 2024-09-10
Fix ansible roles to always mount the cacerts

vakwetu:
If cacerts are present, they should be mounted
even if tls_enabled is not set as they could contain third party certs.

Fixes edpm_neutron_ovn, edpm_ovn, neutron_sriov,
neutron_metadata, edpm_ovn_bgp_agent, edpm_neutron_dhcp, telemetry

Jira: https://issues.redhat.com//browse/OSPRH-9451

### DIFF
--- a/roles/edpm_neutron_metadata/defaults/main.yml
+++ b/roles/edpm_neutron_metadata/defaults/main.yml
@@ -57,9 +57,11 @@ edpm_neutron_metadata_agent_ovn_ovsdb_probe_interval: '60000'
 # tls
 edpm_neutron_metadata_agent_tls_enabled: "{{ edpm_tls_certs_enabled | default(False) }}"
 
+edpm_neutron_metadata_tls_cacert_bundle_src: "/var/lib/openstack/cacerts/{{ edpm_neutron_metadata_service_name }}/tls-ca-bundle.pem"
+edpm_neutron_metadata_tls_cacert_bundle_dest: "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
+edpm_neutron_metadata_tls_cacert_volumes: []
+
 edpm_neutron_metadata_tls_volumes:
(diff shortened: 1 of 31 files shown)


vakwetu:
Note: still needs some work to account for cases where cacerts are not present.  Testing in progress.

slagle (commented):
this is looking good to me, but yea, need to handle when the certs don't exist, according to the molecule failures

roles/edpm_neutron_metadata/defaults/main.yml
```diff
@@ -57,9 +57,11 @@ edpm_neutron_metadata_agent_ovn_ovsdb_probe_interval: '60000'
 # tls
 edpm_neutron_metadata_agent_tls_enabled: "{{ edpm_tls_certs_enabled | default(False) }}"
 
+edpm_neutron_metadata_tls_cacert_bundle_src: "/var/lib/openstack/cacerts/{{ edpm_neutron_metadata_service_name }}/tls-ca-bundle.pem"
+edpm_neutron_metadata_tls_cacert_bundle_dest: "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
+edpm_neutron_metadata_tls_cacert_volumes: []
```
jpodivin, line 62: Default in argspecs and in defaults/main.yml must match. You can add note that it's set in as part of the configuration.

jpodivin (commented):
Good idea, the failures in tests have to be resolved, but other than that it's fine indeed.

vakwetu:
recheck

karelyatin:
This broke update job[1], for instance cacert no longer mounted in ovn_metadata container since this PR. non update job works fine.

[1] https://issues.redhat.com/browse/OSPRH-10450
[2] https://issues.redhat.com/browse/OSPRH-10451

PR #751, opened 2024-09-16
Remove some debug vars

rabi:
They're not used.


rabi:
recheck

jpodivin (approved):
These are essentially leftovers from the old template. I have no problem with the removal, at one point we thought they may be used for something again, but since we didn't bother to implement anything in the mean time, I think it's safe to say that it wasn't essential.

slagle (approved):
recheck

rabi:
recheck

PR #757, opened 2024-09-23
Gather ssh key ansible facts for edpm-ssh-known-hosts

rabi:
This was missing from the PR which changed gather_facts default to false.


gibizer (approved):
yepp this role depends on those ansible facts

PR #758, opened 2024-09-23
Add multinode tempest job to edpm-ansible

rabi:
Without it, we miss testing some of the roles like edpm-ssh-known-hosts.

Let's see if adding a parent from openstack-operator works. Else we have to copy the full definition.

jira: https://issues.redhat.com/browse/OSPRH-10307


rabi:
recheck

bshephar (approved):

PR #759, opened 2024-09-23
Add pre-adoption check for machined

bogdando:
Fail pre-adoption check, if systemd-container package is missing. The package must be pre-installed to not affect running workloads during EDPM adoption.

Related molecule fix: use validations enablement feature flags
to properly isolate testing cases.

JIRA: [OSPRH-10318](https://issues.redhat.com/browse/OSPRH-10318)

### DIFF
--- a/roles/edpm_pre_adoption_validation/tasks/machined.yml
+++ b/roles/edpm_pre_adoption_validation/tasks/machined.yml
@@ -1,0 +1,37 @@
+---
+# Copyright 2024 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+# NOTE(gibi): We need this trick as these tasks depend on the
+# edpm_tuned_profile variable. If that var is not provided in the inventory
+# then it is defaulted by the edpm_tuned role to a non trivial default value.
+- name: Check systemd-container package present or not
+  ansible.builtin.package:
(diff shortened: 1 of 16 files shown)


bogdando:
fixed molecule issues!

bogdando:
recheck [openstack-k8s-operators-content-provider](https://softwarefactory-project.io/zuul/t/rdoproject.org/build/dea1c2bbf57e459092d9be54d6d3276c)

fao89 (commented):
/approve

bogdando:
PTAL

roles/edpm_pre_adoption_validation/tasks/machined.yml
```diff
@@ -0,0 +1,37 @@
...
+  ansible.builtin.assert:
+    that:
+      - is_systemd_container_package.changed
+    success_msg: Found installed systemd-container package
+    fail_msg: |
+      You must have the systemd-container package installed.
+      Do not install it manually as that stops VM instances.
+      Complete a minor update of the source cloud and reboot
+      your Nova Compute hypervisors ony by one. To reboot
+      a hypervisor, firstly live-migrate its VM instances.
```
gibizer, line 37: Do we also need to make sure that the compute was restarted after the systemd-containerd is installed?

gibizer: I know this is not relevant in our CI where we install this package before deploying the test VM in 17.1 but for a real life adoption having a compute with VMs that has the package installed but never restarted could cause a problem.

gibizer: Or do we assume that it is only gets installed during a reboot?

bogdando: no, we don't need reboot nodes after that package installed. This is said here to reboot only because this is the way how we deliver that package via minor updates in tripleo

gibizer: So if we see that the package is installed then can we assume that it was installed during a reboot in the last minor upgrade in 17.1? If so then I'm OK with it.

bogdando: yes, correct

bogdando:
could we merge that one, folks?

gibizer (approved):
looks good.

PR #763, opened 2024-09-25
Ensure fact cache is cleared before running frr molecule

bshephar:
This change ensures fact_cache is cleared prior to converging the FRR molecule job. This ensures we are testing our role in conditions that we can expect to represent a real deployment.

Jira: https://issues.redhat.com/browse/OSPRH-10353
Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/764

### DIFF
--- a/roles/edpm_frr/molecule/default/converge.yml
+++ b/roles/edpm_frr/molecule/default/converge.yml
@@ -17,6 +17,7 @@
 - name: Converge
   hosts: all
   become: true
+  gather_facts: false
   vars:
     edpm_download_cache_podman_auth_file: "" # Override since the file doesn't exist
   pre_tasks:
--- a/roles/edpm_ovn_bgp_agent/molecule/default/converge.yml
+++ b/roles/edpm_ovn_bgp_agent/molecule/default/converge.yml
@@ -16,7 +16,7 @@
 
 - name: Converge
   hosts: all
-  gather_facts: true
+  gather_facts: false
   become: true
   roles:
     - role: "edpm_ovn_bgp_agent"


roles/edpm_frr/meta/argument_specs.yml
```diff
@@ -143,7 +143,7 @@ argument_specs:
         description: ''
         type: str
       edpm_frr_hostname:
-        default: '{{ ansible_facts[''hostname''] | default("")}}'
+        default: '{{ ansible_facts[''hostname''] }}'
```
bshephar, line 146: This is just to see if we can simulate the failure using molecule. Assuming this does indeed fail in the same way outlined in this bug: https://issues.redhat.com/browse/OSPRH-10338

Then we can revert this specific part of the change

bshephar: Indeed, this did manage to reproduce the failure:
https://softwarefactory-project.io/zuul/t/rdoproject.org/build/745dcfd57bf548a1a43d61aa7d899476

bshephar:
retest

roles/edpm_frr/molecule/default/converge.yml
```diff
@@ -30,6 +31,10 @@
       ansible.builtin.include_role:
         name: "osp.edpm.edpm_frr"
         tasks_from: "download_cache.yml"
+
```
rabi, line 34: We probably need to remove it from https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci/playbooks/molecule-test.yml#L4 and  add `gather_facts: false` in molecule prepare/converge.yaml for all roles rather than collecting and then cleaning them. That way all roles would be tested without default `gather_facts: true` I guess.

rabi:
recheck

roles/edpm_frr/molecule/default/converge.yml
bshephar: I looked at disabling it in `test_deps`:
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/molecule/common/test_deps/tasks/main.yml#L17-L41

Seemed like there was more thought required before doing that though. Since `test_deps` gather them, I think everything else will already have a fact cache available because they all call `test_deps` first.

But maybe we just need to disable it everywhere and then fix whatever breaks.

bshephar: Ah, right. Yeah I see, it's specifically excluding `min`.

bshephar: Ok, so I'll drop the meta task to clear facts and we can set a Depends-On for your PR in CIFMW:
https://github.com/openstack-k8s-operators/ci-framework/pull/2383

rabi: I've changed it in all roles with https://github.com/openstack-k8s-operators/edpm-ansible/pull/767/files.

bshephar: Ah cool, we can close this one then.

PR #769, opened 2024-10-01
Ensure ssh_known_hosts file exists

fao89:
closes [OSPRH-8412](https://issues.redhat.com//browse/OSPRH-8412)

### DIFF
--- a/roles/edpm_nova/tasks/main.yml
+++ b/roles/edpm_nova/tasks/main.yml
@@ -14,6 +14,13 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Ensure ssh_known_hosts file exists
+  become: true
+  ansible.builtin.file:
+    path: /etc/ssh/ssh_known_hosts
+    state: touch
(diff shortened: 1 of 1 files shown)


roles/edpm_nova/tasks/main.yml
```diff
@@ -14,6 +14,18 @@
...
+    - name: Check for ssh_known_hosts file
+      ansible.builtin.stat:
+        path: /etc/ssh/ssh_known_hosts
+      register: _ssh_known_hosts
+
+    - name: Fail if ssh_known_hosts file is missing
+      ansible.builtin.fail:
+        msg: "Nova depends on ssh-known-hosts OpenStackDataPlaneService"
+      when: not _ssh_known_hosts.stat.exists | bool
+
```
slagle, line 28: couldn't we just touch the file (as root) if it does't exist? I guess there might be some valid reasons to omit using the ssh-known-hosts service.

slagle (approved):

PR #771, opened 2024-10-01
Fix CAcert mounting

vakwetu:
The logic to determine whether to mount a cacert bundle was mistakenly placed in configure.yml instead of run.yaml where the templates that use this logic are generated. (run.yaml and run_ovn.yaml)  This breaks updates which execute run.yaml instead of configure.yaml

This patch moves the logic to the right place.
Resolves: https://issues.redhat.com/browse/OSPRH-10450


karelyatin (commented):
LGTM Thanks this should work for update, but any idea why configure.yml and others not running during update? as we could miss some configuration during udpate

bshephar (approved):

PR #776, opened 2024-10-07
Fix creating huge pages

SeanMooney:
hugepages shoudn't have mounted path, only name

Conflicts:
    roles/edpm_extra_mounts/defaults/main.yml
    roles/edpm_extra_mounts/molecule/default/converge.yml

conflcits were resolved by accpeting the incoming version

Closes: [OSPRH-10215](https://issues.redhat.com//browse/OSPRH-10215)
(cherry picked from commit ca0fe3ced6ebb837ad0d0b0994857d986c2c9cd4)

### DIFF
--- a/roles/edpm_extra_mounts/defaults/main.yml
+++ b/roles/edpm_extra_mounts/defaults/main.yml
@@ -23,5 +23,5 @@ edpm_extra_mounts_hide_sensitive_logs: true
 
 edpm_extra_mounts: []
 edpm_default_mounts: [
-  {name: 'hugepages1G', path: '/dev/hugepages1G', opts: 'pagesize=1G', fstype: 'hugetlbfs'},
-  {name: 'hugepages2M', path: /dev/hugepages2M, opts: 'pagesize=2M', fstype: 'hugetlbfs'}]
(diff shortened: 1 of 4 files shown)


gibizer:
Do we need to cherry pick this now that 18.0.2 closed? I assume 18.0.3 is the same as FP1 and that its content will be from main not from 18.0.0-propose (or 18.0.0-proposed will be reset to main)

SeanMooney:
> Do we need to cherry pick this now that 18.0.2 closed? I assume 18.0.3 is the same as FP1 and that its content will be from main not from 18.0.0-propose (or main will be merged into 18.0.0-proposed)

this should have been cherry-picked for 18.0.2 because it was merged before the code freeze for it on main but we forgot to do that at that time. I am actually debating if this is worth escalating as an expectation for 18.0.2 which was the main reason I created this cherry pick but strictly speaking this is not required.

18.0.3 and FR1 may not be the same the feature release, FR1 can be deferred at most once if required which would align with 18.0.4 depending on the state of verification.
While the intent is to be time-based fi there are infra issues then that is what we would do.

PR #779, opened 2024-10-09
Healthchecks resubmit

paramite:
This PR adds health checks for containerized services.

Related: [OBSDA-574](https://issues.redhat.com/browse/OBSDA-574)
Related: [OSPRH-1052](https://issues.redhat.com/browse/OSPRH-1052)
Closes: [OSPRH-4118](https://issues.redhat.com/browse/OSPRH-4118)

PR #784, opened 2024-10-14
Use edpm_container_manage role to deploy node_exporter

yadneshk:


yadneshk:
This is done in https://github.com/openstack-k8s-operators/edpm-ansible/pull/779

PR #788, opened 2024-10-16
Replace creator-ee with ansible-dev-tools image

fao89:
As [quay.io/ansible/creator-ee](https://quay.io/repository/ansible/creator-ee?tab=tags&tag=latest)  is no longer maintained: https://github.com/ansible/creator-ee

closes [OSPRH-10743](https://issues.redhat.com//browse/OSPRH-10743)

### DIFF
--- a/openstack_ansibleee/Dockerfile
+++ b/openstack_ansibleee/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/ansible/creator-ee:v0.21.0 as builder
+FROM ghcr.io/ansible/community-ansible-dev-tools:v24.10.0 as builder
 
 ARG REMOTE_SOURCE=.
 ARG REMOTE_SOURCE_DIR=/var/tmp/edpm-ansible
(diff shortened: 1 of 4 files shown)


openstack_ansibleee/Dockerfile
```diff
@@ -18,10 +18,15 @@ COPY --from=builder /usr/share/ansible /usr/share/ansible
...
+RUN chmod +x /opt/builder/bin/edpm_entrypoint
 ENV EDPM_SYSTEMROLES='fedora.linux_system_roles'
+RUN chmod ug+rw /etc/passwd
+RUN mkdir -p /runner && chgrp 0 /runner && chmod -R ug+rwx /runner
 WORKDIR /runner
-ENTRYPOINT ["edpm_entrypoint"]
+LABEL ansible-execution-environment=true
+USER 1000
+ENTRYPOINT ["/opt/builder/bin/edpm_entrypoint", "dumb-init"]
+CMD ["bash"]
```
slagle, line 32: is the CMD needed?

openstack_ansibleee/Dockerfile
```diff
@@ -18,10 +18,15 @@ COPY --from=builder /usr/share/ansible /usr/share/ansible
...
+RUN sed -i '1d' /opt/builder/bin/entrypoint
+RUN cat /opt/builder/bin/entrypoint >> /opt/builder/bin/edpm_entrypoint
+RUN chmod +x /opt/builder/bin/edpm_entrypoint
 ENV EDPM_SYSTEMROLES='fedora.linux_system_roles'
+RUN chmod ug+rw /etc/passwd
+RUN mkdir -p /runner && chgrp 0 /runner && chmod -R ug+rwx /runner
 WORKDIR /runner
-ENTRYPOINT ["edpm_entrypoint"]
+LABEL ansible-execution-environment=true
+USER 1000
```
slagle, line 30: this ends up adding my local user to the container image. that doesn't seem right.

openstack_ansibleee/Dockerfile
fao89: I just tried to follow the ansible-builder output: https://gist.github.com/fao89/64b2f35d6d5edfb8c96afe525c817c7e

fao89: but I don't think it is needed

slagle: with an empty execution-environment.yaml file, I get different results than that gist. The base image used is actually, quay.io/ansible/ansible-runner:latest.

I wonder if we could that as our base image instead of ghcr.io/ansible/community-ansible-dev-tools:latest

fao89: that image seems to be 2 years old:https://quay.io/repository/ansible/ansible-runner?tab=tags&tag=latest
my execution-environment.yaml was based on https://ansible.readthedocs.io/projects/dev-tools/container/#usage  (without the python requirements and package manager)

fao89: I started this: https://github.com/ansible/ansible-builder/pull/715 so we can see what ansible-builder people have to say about the image

fao89: From https://github.com/ansible/ansible-builder/pull/715#issuecomment-2427624093 it seems quay.io/ansible/ansible-runner is there to keep compatibility with older versions

slagle: ok, then I think we can proceed with this.
In the longer term, we might want to look at migrating the image build to use ansible-builder directly instead of maintaining our own method with the Containerfile.

slagle (approved):

openstack_ansibleee/Dockerfile
slagle: I filed https://issues.redhat.com/browse/OSPRH-11080 to track that work

fao89: we can follow ansible-dev-tools image example, as it is built with ansible-builder: https://github.com/ansible/ansible-dev-tools/blob/main/execution-environment.yml

PR #793, opened 2024-10-18
Do not store power monitoring config files with telemetry

yadneshk:
Since, ipmi agent is deployed by power monitoring role it's configuration should not be stored with telemetry services. This is necessary since both require different polling.yaml files.


fao89 (commented):
/approve

PR #799, opened 2024-11-06
Don't use rescue with any_errors_fatal: true

rabi:
This seems broken with ansible 2.17.x. Also the current logic of running certain tasks for failed hosts won't work with `any_errors_fatal: true` as it would run them for all nodes.

Jira: https://issues.redhat.com/browse/OSPRH-11264


fao89 (commented):
should we do the same here? https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_frr/tasks/configure.yml#L34

rabi:
> should we do the same here? https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_frr/tasks/configure.yml#L34

Hmm.. that's using rescue block for a custom failure message. Could be written differently, but yeah I'll also fix that.

PR #805, opened 2024-11-11
Use ubi9 image for aee

rabi:
The current image is a base fedora image that installs packages using pip and prone to be broken with latest changes in packages.

This uses ubi9-minimal as the base image and also reduces the image size by ~2/3rd (455 MB vs 167MB).


rabi:
Not sure if we can create an image based on `ee-minimal-rhel9` and push it to quay.

fao89:
we can use ansible-builder:
https://ansible.readthedocs.io/projects/builder/en/latest/index.html#choosing-a-base-image
this is how the current base image is built:
https://github.com/ansible/ansible-dev-tools/blob/main/execution-environment.yml

openstack_ansibleee/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/ansible/community-ansible-dev-tools:v24.10.0 as builder
+FROM registry.redhat.io/ansible-automation-platform-24/ee-minimal-rhel9:1.0 as builder
```
slagle, line 1: This requires a login, so I don't think we can use this upstream.

slagle (commented):
+2
I like this, and the size reduction is nice. Given there's no ee-minimal image available upstream on quay.io, using ubi9 seems pretty reasonable.

I'll leave it to @fao89 to approve, just to make sure he doesn't have any concerns.

fao89 (approved):

fao89:
I think OSPRH-11080 was covered by this PR

rabi:
> I think [OSPRH-11080](https://issues.redhat.com//browse/OSPRH-11080) was covered by this PR

I think the intent of that jira is to write an `execution-environment.yml ` that can be used with `ansible-builder`.  `ansible-builder` generates a `Containerfile` from `execution-environment.yml` (which we're now doing manually) and runs podman build with it. If we want to just maintain an `execution-environment.yml ` , we can try that as well, but I had found `ansible-builder` to be buggy in some of my tests (sometimes the build command returns without generating the image).

PR #809, opened 2024-11-13
Add nw connectivity task to all playbooks

stuggi:
When using bgp where pods using a network-attachment-definition it takes some seconds to advertise it on the fabric. As a result roles which don't have a the connectivity task check will fail.

A rerun of the ansible jobs will not fix this because every time the deployment job pod gets created it will receive a new ip from the ctlplane pool.

There were discussions on a more generic way on slack, for now lets just add the check to all the playbooks which is already there on the bootstrap playbook:
https://github.com/openstack-k8s-operators/edpm-ansible/blob/6615889c5d708df2accb7dafec5aa099946292b8/playbooks/bootstrap.yml#L8

Jira: https://issues.redhat.com/browse/OSPRH-8680

### DIFF
--- a/playbooks/ceph_client.yml
+++ b/playbooks/ceph_client.yml
@@ -1,5 +1,6 @@
 ---
-
+- name: Check remote connections
+  ansible.builtin.import_playbook: connection.yml
 - name: Configure EDPM as client of Ceph
   hosts: "{{ edpm_override_hosts | default('all', true) }}"
   strategy: linear
--- a/playbooks/configure_network.yml
+++ b/playbooks/configure_network.yml
@@ -1,5 +1,6 @@
 ---
-
+- name: Check remote connections
+  ansible.builtin.import_playbook: connection.yml
 - name: Deploy EDPM Network
   hosts: "{{ edpm_override_hosts | default('all', true) }}"
   strategy: linear
--- a/playbooks/bootstrap.yml
+++ b/playbooks/bootstrap.yml
@@ -1,13 +1,6 @@
 ---
 - name: Check remote connections
-  hosts: "{{ edpm_override_hosts | default('all', true) }}"
-  strategy: linear
-  gather_facts: "{{ gather_facts | default(false) }}"
-  tasks:
-    - name: Wait for connection
-      ansible.builtin.wait_for_connection:
-        delay: "{{ edpm_wait_for_connection_delay | default(10) }}"
-        timeout: "{{ edpm_wait_for_connection_timeout | default(600) }}"
+  ansible.builtin.import_playbook: connection.yml
 - name: Bootstrap node
   hosts: "{{ edpm_override_hosts | default('all', true) }}"
   strategy: linear
--- a/playbooks/ceph_hci_pre.yml
+++ b/playbooks/ceph_hci_pre.yml
@@ -1,5 +1,6 @@
 ---
-
+- name: Check remote connections
+  ansible.builtin.import_playbook: connection.yml
 - name: Prepare EDPM to Host Ceph
   hosts: "{{ edpm_override_hosts | default('all', true) }}"
   strategy: linear
--- a/playbooks/configure_os.yml
+++ b/playbooks/configure_os.yml
@@ -1,5 +1,6 @@
 ---
-
+- name: Check remote connections
+  ansible.builtin.import_playbook: connection.yml
 - name: Deploy EDPM Operating System Configure
   hosts: "{{ edpm_override_hosts | default('all', true) }}"
   strategy: linear
--- a/playbooks/configure_ovs_dpdk.yml
+++ b/playbooks/configure_ovs_dpdk.yml
@@ -1,5 +1,6 @@
 ---
-
+- name: Check remote connections
+  ansible.builtin.import_playbook: connection.yml
 - name: Configure OvS DPDK
   hosts: "{{ edpm_override_hosts | default('all', true) }}"
   strategy: linear
--- a/playbooks/connection.yml
+++ b/playbooks/connection.yml
@@ -1,0 +1,9 @@
+---
+- name: Check remote connections
+  hosts: "{{ edpm_override_hosts | default('all', true) }}"
+  gather_facts: false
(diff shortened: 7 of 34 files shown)


rabi:
> There were discussions on a more generic way on slack, for now lets just add the check to all the playbooks which is already there on the bootstrap playbook:

As mentioned in the slack thread, we also have to ensure that all custom services have that task, and it has to be documented. If we know it's only for few seconds, adding `sleep x` [1] or some relevant check to https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/openstack_ansibleee/edpm_entrypoint.sh would possibly do the job rather than changing all roles?

[1] https://github.com/openstack-k8s-operators/edpm-ansible/compare/main...rabi:edpm-ansible:wait_for_network?expand=1

playbooks/configure_network.yml
```diff
@@ -1,5 +1,13 @@
 ---
-
+- name: Check remote connections
+  hosts: "{{ edpm_override_hosts | default('all', true) }}"
+  strategy: linear
+  gather_facts: "{{ gather_facts | default(false) }}"
+  tasks:
```
rabi, line 6: If decide to go this route, then we also should add an ansible var to run the task only for bgp deployments.

playbooks/ceph_client.yml
```diff
@@ -1,5 +1,13 @@
...
-
+- name: Check remote connections
+  hosts: "{{ edpm_override_hosts | default('all', true) }}"
+  strategy: linear
+  gather_facts: "{{ gather_facts | default(false) }}"
+  tasks:
+    - name: Wait for connection
+      ansible.builtin.wait_for_connection:
+        delay: "{{ edpm_wait_for_connection_delay | default(10) }}"
+        timeout: "{{ edpm_wait_for_connection_timeout | default(600) }}"
```
slagle, line 10: can we put this task into a common role? I'm thinking edpm_nodes_validation, in a separate tasks file that can be included.

stuggi: yes, I think that makes sense

playbooks/ceph_client.yml
```diff
@@ -1,12 +1,15 @@
 ---
-
 - name: Configure EDPM as client of Ceph
   hosts: "{{ edpm_override_hosts | default('all', true) }}"
   strategy: linear
   gather_facts: "{{ gather_facts | default(false) }}"
   any_errors_fatal: "{{ edpm_any_errors_fatal | default(true) }}"
   max_fail_percentage: "{{ edpm_max_fail_percentage | default(0) }}"
   tasks:
+    - name: Check remote connections
```
rabi, line 9: This won't work when `gather_facts` is true, as it would try gathering facts and fail before calling the task to `wait_for_connection`. That's why it was added as a separate play in [1] with `gather_facts: false`. We regressed that later though.

We either  create a separate play (with `gather_facts: false`) like earlier or change it to false in every playbook (not honor user provided `gather_facts` ansible var at all).

Also, what're we going to do for custom services? Are we not expecting people to use them in bgp deployments? Are we going to document that?

[1] https://github.com/openstack-k8s-operators/edpm-ansible/commit/07a081594bcaff4edc85230663b00f1142351777

stuggi: thanks for the pointer on the gather_facts. I'll update it. 

re custom services, from my pov I think its ok to document it. @slagle whats your take on this?

stuggi: we tried to just pass in the `ANSIBLE_TIMEOUT` env var via the nodeset, but it does not work. Not sure if the `No route to host`  error results in the timeout to be handled as we need it.

```
$ oc logs download-cache-edpm-deployment-r2-compute-nodes-r2-fqvgp                                                                                                                        
Identity added: /runner/artifacts/download-cache-edpm-deployment-r2-compute-nodes-r2/ssh_key_data (cifmw_reproducer_key)                                                                                              
[WARNING]: Invalid characters were found in group names but not replaced, use
-vvvv to see details

PLAY [EDPM Download cache] *****************************************************

TASK [osp.edpm.edpm_download_cache : Validating arguments against arg spec 'main' - The main entry point for the edpm_download_cache role.] ***                                                                       
ok: [r2-compute-ib21ttu9-0]

TASK [osp.edpm.edpm_download_cache : Install role requirements] ****************
included: /usr/share/ansible/collections/ansible_collections/osp/edpm/roles/edpm_download_cache/tasks/install.yml for r2-compute-ib21ttu9-0                                                                           

TASK [Execute bootstrap command] ***********************************************
included: osp.edpm.edpm_bootstrap for r2-compute-ib21ttu9-0

TASK [osp.edpm.edpm_bootstrap : Re-read facts] *********************************
fatal: [r2-compute-ib21ttu9-0]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: ssh: connect to host r2-compute-ib21ttu9-0 port 22: No route to host", "unreachable": true}         

NO MORE HOSTS LEFT *************************************************************

NO MORE HOSTS LEFT *************************************************************

PLAY RECAP *********************************************************************
r2-compute-ib21ttu9-0      : ok=3    changed=0    unreachable=1    failed=0    skipped=0    rescued=0    ignored=0   
```

stuggi: @slagle you have other idea for a ansible configuration to test instead of what we tried as mentioned above?

slagle: ` ANSIBLE_SSH_RETRIES` and `ANSIBLE_SSH_TIMEOUT` ( i believe that's the same as `ANSIBLE_TIMEOUT`, but it may be worth trying)

stuggi: tests with setting ANSIBLE_SSH_RETRIES and ANSIBLE_SSH_TIMEOUT were successful. I'll close this pr. Thanks a lot @slagle

stuggi:
tests with setting ANSIBLE_SSH_RETRIES and ANSIBLE_SSH_TIMEOUT were successful. I'll close this pr. Thanks a lot @slagle

PR #814, opened 2024-11-15
Escalate privs for NetworkManager-ovs install

bshephar:
When installing packages, we need to escalate privileges. This change adds become: true to the block that is installing NetworkManager-ovs for use with nmstate.

Jira: https://issues.redhat.com/browse/OSPRH-11663


leducflorian (approved):
LGTM

slagle (approved):

fao89 (approved):

PR #825, opened 2024-11-22
DNM Pin edpm-ansible collections to downstream versions

rabi:


slagle (commented):
I think we might want this actually. Using a consistent version across upstream/downstream could prevent some incompatible changes from going in. Should we revive the PR?

rabi:
> I think we might want this actually. Using a consistent version across upstream/downstream could prevent some incompatible changes from going in. Should we revive the PR?

Sorry I missed replying to this comment earlier. Probably I mentioned it earlier, unless we can pin podman (and other packages), we can't pin  ansible collections like `containers.podman` as older ansible collections tend to cause regressions with newer podman. We have had so many instances in the past.

PR #813, opened 2024-11-14
Initial bootc support

slagle:
- **Add edpm_bootc role**
- **Update edpm_bootstrap for bootc**

Jira: [OSPRH-11433](https://issues.redhat.com//browse/OSPRH-11433)
Jira: [OSPRH-11505](https://issues.redhat.com//browse/OSPRH-11505)

PR #830, opened 2024-11-27
Support bootc

bshephar:
This PR adds a number of changes to roles in order to facilitate the use of image mode RHEL.

### DIFF
--- a/playbooks/bootstrap.yml
+++ b/playbooks/bootstrap.yml
@@ -42,6 +42,7 @@
         name: osp.edpm.edpm_kernel
       tags:
         - edpm_kernel
+      when: not bootc
     - name: Import edpm_tuned
       ansible.builtin.import_role:
         name: osp.edpm.edpm_tuned
@@ -53,9 +54,11 @@
         tasks_from: kernelargs.yml
       tags:
         - edpm_kernel
+      when: not bootc
     - name: Configure KSM for kernel
       ansible.builtin.import_role:
         name: osp.edpm.edpm_kernel
         tasks_from: ksm.yml
       tags:
         - edpm_kernel
+      when: not bootc
--- a/playbooks/bootstrap.yml
+++ b/playbooks/bootstrap.yml
@@ -42,7 +42,7 @@
         name: osp.edpm.edpm_kernel
       tags:
         - edpm_kernel
-      when: not bootc
+      when: not ansible_local.bootc
     - name: Import edpm_tuned
       ansible.builtin.import_role:
         name: osp.edpm.edpm_tuned
@@ -54,11 +54,11 @@
         tasks_from: kernelargs.yml
       tags:
         - edpm_kernel
-      when: not bootc
+      when: not ansible_local.bootc
     - name: Configure KSM for kernel
       ansible.builtin.import_role:
         name: osp.edpm.edpm_kernel
         tasks_from: ksm.yml
       tags:
         - edpm_kernel
-      when: not bootc
+      when: not ansible_local.bootc
--- a/roles/edpm_container_manage/tasks/shutdown.yml
+++ b/roles/edpm_container_manage/tasks/shutdown.yml
@@ -27,7 +27,7 @@
       loop:
         - 'edpm-container-shutdown'
         - 'edpm-start-podman-container'
-      when: not edpm_use_bootc
+      when: not ansible_local.bootc
 
     - name: "Create systemd preset dir"
       ansible.builtin.file:
--- a/roles/edpm_container_manage/tasks/shutdown.yml
+++ b/roles/edpm_container_manage/tasks/shutdown.yml
@@ -27,17 +27,25 @@
       loop:
         - 'edpm-container-shutdown'
         - 'edpm-start-podman-container'
-    - name: "Create /usr/lib/systemd/system/edpm-container-shutdown.service"
+      when: not edpm_use_bootc
+
+    - name: "Create systemd preset dir"
+      ansible.builtin.file:
+        state: "directory"
+        path: "/etc/systemd/system-preset"
+        mode: 0644
+
+    - name: "Create /etc/systemd/system/edpm-container-shutdown.service"
       ansible.builtin.copy:
         src: "{{ role_path }}/files/edpm-container-shutdown-service"
-        dest: "/usr/lib/systemd/system/edpm-container-shutdown.service"
+        dest: "/etc/systemd/system/edpm-container-shutdown.service"
         mode: '0644'
         owner: root
         group: root
-    - name: "Create /usr/lib/systemd/system-preset/91-edpm-container-shutdown.preset"
+    - name: "Create /etc/systemd/system-preset/91-edpm-container-shutdown.preset"
       ansible.builtin.copy:
         src: "{{ role_path }}/files/91-edpm-container-shutdown-preset"
-        dest: "/usr/lib/systemd/system-preset/91-edpm-container-shutdown.preset"
+        dest: "/etc/systemd/system-preset/91-edpm-container-shutdown.preset"
         mode: '0644'
         owner: root
         group: root
@@ -47,17 +55,17 @@
         state: started
         enabled: true
         daemon_reload: true
-    - name: "Create /usr/lib/systemd/system/netns-placeholder.service"
+    - name: "Create /etc/systemd/system/netns-placeholder.service"
       ansible.builtin.copy:
         src: "{{ role_path }}/files/netns-placeholder-service"
-        dest: "/usr/lib/systemd/system/netns-placeholder.service"
+        dest: "/etc/systemd/system/netns-placeholder.service"
         mode: '0644'
         owner: root
         group: root
-    - name: "Create /usr/lib/systemd/system-preset/91-netns-placeholder.preset"
+    - name: "Create /etc/systemd/system-preset/91-netns-placeholder.preset"
       ansible.builtin.copy:
         src: "{{ role_path }}/files/91-netns-placeholder-preset"
-        dest: "/usr/lib/systemd/system-preset/91-netns-placeholder.preset"
+        dest: "/etc/systemd/system-preset/91-netns-placeholder.preset"
         mode: '0644'
         owner: root
         group: root
--- a/roles/edpm_kernel/defaults/main.yml
+++ b/roles/edpm_kernel/defaults/main.yml
@@ -36,4 +36,3 @@ edpm_nova_compute_config_dir: /var/lib/config-data/ansible-generated/nova_libvir
 
 # KSM control
 edpm_kernel_enable_ksm: false
-edpm_use_bootc: false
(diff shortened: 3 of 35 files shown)


roles/edpm_kernel/defaults/main.yml
```diff
@@ -36,3 +36,4 @@ edpm_nova_compute_config_dir: /var/lib/config-data/ansible-generated/nova_libvir
 
 # KSM control
 edpm_kernel_enable_ksm: false
+edpm_use_bootc: false
```
bshephar, line 39: We probably need a better way to implement this globally. But at least for testing purposes, this is what I've used to get something that deploys.

roles/edpm_logrotate_crond/tasks/configure.yml
```diff
@@ -17,9 +17,15 @@
...
+    - name: Create /var/lib/openstack/cron
+      ansible.builtin.file:
+        path: "/var/lib/openstack/cron"
+        mode: 0644
+        state: directory
+
     - name: Push script
       ansible.builtin.copy:
-        dest: /usr/local/sbin/containers-tmpwatch
+        dest: /var/lib/openstack/cron/containers-tmpwatch
```
bshephar, line 28: `/usr` is immutable with bootc deployments. So I've proposed doing this in two different ways. 1, we bake the scripts into the container file:
https://github.com/openstack-k8s-operators/install_yamls/pull/950/files#diff-f8fb9af5355b45b9ca8936bf0d721c6f0e37e13b637f5598e2be19995dea23e7R45-R46

And 2. Which is this method of writing to `/var/lib/openstack`. I personally prefer doing it this way if we can agree on a common place for any scripts that we want to use. That saves us baking things into images and then trying to keep them in sync. Better imo to have them in edpm-ansible for now.

roles/edpm_ovn/tasks/run.yml
```diff
@@ -33,11 +33,7 @@
...
 
 - name: Run ovn_controller container
-  ansible.builtin.include_role:
-    name: osp.edpm.edpm_container_standalone
-  vars:
-    edpm_container_standalone_service: ovn_controller
-    edpm_container_standalone_container_defs:
-      ovn_controller: "{{ lookup('template', 'ovn_controller.yaml.j2') | from_yaml }}"
-    edpm_container_standalone_kolla_config_files:
-      ovn_controller: "{{ lookup('template', 'kolla_ovn_controller.yaml.j2') | from_yaml }}"
```
bshephar, line 43: All of this needs to stay in order to support both deployment methodologies. It can just be conditional like:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/830/files#diff-34e3323585e197e806d463771e3b5132716048c41818b1318fecb2c0d8e36cd6R45

roles/edpm_ovn/tasks/run.yml
```diff
@@ -43,9 +43,52 @@
...
+    ansible.builtin.set_fact:
+      ovn_controller_pod_spec: "{{ ovn_controller_pod_spec.content | b64decode | from_yaml }}"
+
+  - name: Update fields in the YAML data
+    ansible.builtin.set_fact:
+      updated_pod_spec: >-
+        {{
+          ovn_controller_pod_spec | combine({
+            'spec': {
+              'containers': ovn_controller_pod_spec.spec.containers | zip_longest([], [{'image': edpm_ovn_controller_agent_image}]) | map('combine') | list,
```
slagle, line 63: I believe that if we were to customize the image like this then by definition the container is no longer "logically bound". Instead, it would be considered a "floating" container per https://containers.github.io/bootc/logically-bound-images.html#comparison-with-default-podman-systemd-units

and also:

> There is no mechanism to inject arbitrary arguments to the podman pull (or equivalent) invocation used by bootc.

which seems to apply additional mounts or other options passed to podman pull are not possible.

The dynamically-injected ConfigMaps[1][2] may provide some customization, but that is still not likely for the app container image itself, b/c once that is changed to some other image, then that no longer fits into how logically bound images should be managed with the lifecycle of the base bootc image itself.

Point being...if we choose to allow the ability to podman run any arbitrary image at runtime, then these really aren't logically bound images at all, but are considered "floating".

The question becomes, should we adopt logically bound images, and require our end users to be building new bootc images before deploying EDPM nodes, depending on if they need to customize any of the container images? We could ship a bootc image that had all the images logically bound, but if a user wanted to run a different one (from a partner, etc) they they would need to rebuild that image.

I do like the quadlet/systemd design, and I think we can still adopt that.

[1] https://containers.github.io/bootc/building/guidance.html?highlight=configmap#configuration
[2] https://github.com/containers/bootc/issues/22

roles/edpm_kernel/defaults/main.yml
slagle: This is why I added the edpm_bootc role in https://github.com/openstack-k8s-operators/edpm-ansible/pull/813 so that we had a way to do it consistently across anywhere that needs it.

slagle: Also, I created a new bootc branch for edpm-ansible: https://github.com/openstack-k8s-operators/edpm-ansible/tree/bootc

Can you propose this PR to the bootc branch instead?

I'll be reverting https://github.com/openstack-k8s-operators/edpm-ansible/pull/813 from main until we are ready to merge all bootc support into main.

slagle:
I created a new bootc branch for edpm-ansible: https://github.com/openstack-k8s-operators/edpm-ansible/tree/bootc

Can you propose this PR to the bootc branch instead?

I reverted https://github.com/openstack-k8s-operators/edpm-ansible/pull/813 from main in https://github.com/openstack-k8s-operators/edpm-ansible/pull/844

roles/edpm_kernel/defaults/main.yml
bshephar: Branch thing is done. But that role would need to be called from each and every playbook to detect and set the `bootc` variable right? I guess we can just add it as a ansibleVar and avoid calling the role each and every time we start a new service.

bshephar (commented):
The entire handling of kernel args needs to be reconsidered for bootc:
https://containers.github.io/bootc/building/kernel-arguments.html

steveb:
> The entire handling of kernel args needs to be reconsidered for bootc: https://containers.github.io/bootc/building/kernel-arguments.html

that kargs.d mechanism looks interesting. I reckon if we're doing anything for the demo, ansible should write to a /usr/lib/bootc/kargs.d/00-edpm.toml and reboot. Also we need to ensure the generated qcow2 has console= arguments removed

roles/edpm_kernel/defaults/main.yml
bshephar: So, we can use a custom fact. For example:
```
cat /etc/ansible/facts.d/bootc.fact
#!/usr/bin/env bash

is_bootc() {
  BOOTC_STATUS=$(sudo bootc status --json | jq .status.type)
  if [[ "$BOOTC_STATUS" == \"bootcHost\" ]]; then
     BOOTC_SYSTEM="true"
  else
     BOOTC_SYSTEM="false"
  fi
}

is_bootc

echo ${BOOTC_SYSTEM}
```

This is good from the perspective of not needing the user to manually define that they are using a bootc system, plus it works for our non bootc systems:
```
[m3@osp-df-3 bootc]$ ansible -i inv.yaml all -m setup -a "filter=ansible_local"
edpm-compute-1 | SUCCESS => {
    "ansible_facts": {
        "ansible_local": {
            "bootc": true
        },
        "discovered_interpreter_python": "/usr/bin/python3"
    },
    "changed": false
}
```

So you could combine them in the same NodeSet if you wanted to. But, the down side of this approach is that we need to gather facts from each service. We have thus far tried to limit the amount of fact gathering required, so this approach my not be what we want to do without some more granular control of which facts are being gathered in each service. At the moment, we just define a variable for `gather_facts`. If that variable is true, then we gather all facts. Obviously, that becomes necessary if we want to allow individual task executions that require facts, but when we want to _just_ gather local facts, then gathering all of them introduces non-trivial time to our executions of each service.

Offering it as a potential solution that we can debate. The alternative is that we require either bootc or non-bootc nodes in each NodeSet.

roles/edpm_container_manage/tasks/shutdown.yml
```diff
@@ -27,17 +27,25 @@
       loop:
         - 'edpm-container-shutdown'
         - 'edpm-start-podman-container'
-    - name: "Create /usr/lib/systemd/system/edpm-container-shutdown.service"
+      when: not edpm_use_bootc
```
slagle, line 30: I don't see `edpm_use_bootc` set anywhere. Is this a leftover from an earlier revision?

roles/edpm_logrotate_crond/tasks/configure.yml
slagle, line 1: seems like the `Insert cronjob in root crontab` task requires the cronie rpm. We might need to add that to the image build.

slagle: https://github.com/openstack-k8s-operators/edpm-image-builder/pull/43

playbooks/bootstrap.yml
```diff
@@ -42,6 +42,7 @@
         name: osp.edpm.edpm_kernel
       tags:
         - edpm_kernel
+      when: not ansible_local.bootc
```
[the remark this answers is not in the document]
bshephar: We need to figure out what we want this to look like still. I have just set `gather_facts: true` in my deployment Ansible vars.

Maybe we would need to change the default for it to `gather_subset: local` at a minimum. It's not ideal that we would gather all facts for every service, but at the moment, that's what my `gather_facts: true` is doing until I come up with something better.

roles/edpm_container_manage/tasks/shutdown.yml
bshephar: Yeah, before the custom fact. I'll fix this

roles/edpm_logrotate_crond/tasks/run.yml
```diff
@@ -27,3 +27,20 @@
...
+    ansible.builtin.copy:
+      src: files/logrotate_crond.yaml
+      dest: /var/lib/kolla/config_files/logrotate_crond.json
+      mode: 0644
+
+  - name: Manage logrotate crond containers
+    ansible.builtin.systemd_service:
+      name: edpm-compute@logrotate_crond
+      enabled: true
+      state: started
```
slagle, line 46: I think we need to separate out this initial bootc support from the logically bound containers PR, https://github.com/openstack-k8s-operators/edpm-image-builder/pull/39

This PR has a strong dep on the logically bound PR, and that complicates things. Let's just get a base bootc working with how we manage containers presently. We can move to logically bound and all the quadlet/systemd stuff as a next step.

bshephar: Yeah, ok. Let's decouple them. I'll just submit a new PR to `edpm-image-builder` to change the Quadlet files over to using `.container` instead of `.kube`. Then a new one here to work with those instead of the `.kube` files.

bshephar:
So, we merged a subset of this in the interest of slimming down the number of required changes. Let's backlog this particular piece of work in favor of getting something demo worthy. We'll collect feedback, and then we can loop back on the logically bound images work if we feel it's still worth pursuing.

PR #831, opened 2024-11-30
Use PAM for sshd Motd

bshephar:
This change switches the method we use to display the motd. Previously, we used PrintMotd in the sshd config. RHEL9 configures mod_motd.so in PAM, and as such the motd will be duplicated when using PrintMotd. To avoid this, we can configure the motd via PAM.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2329414

### DIFF
--- a/roles/edpm_sshd/tasks/configure.yml
+++ b/roles/edpm_sshd/tasks/configure.yml
@@ -74,10 +74,15 @@
          {% if edpm_sshd_banner_enabled %}
          {% set _ = edpm_sshd_server_options.__setitem__('Banner', '/etc/issue') %}
          {% endif %}
-         {% if edpm_sshd_motd_enabled %}
-         {% set _ = edpm_sshd_server_options.__setitem__('PrintMotd', 'yes') %}
-         {% endif %}
          {{ edpm_sshd_server_options }}
+
+    - name: Set sshd motd when enabled
+      ansible.builtin.lineinfile:
+        path: /etc/pam.d/sshd
+        regexp: "^session.*optional.*pam_motd.so"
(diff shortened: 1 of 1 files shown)


bshephar:
recheck

roles/edpm_sshd/tasks/configure.yml
```diff
@@ -74,11 +74,16 @@
...
-         {% if edpm_sshd_motd_enabled %}
-         {% set _ = edpm_sshd_server_options.__setitem__('PrintMotd', 'yes') %}
-         {% endif %}
          {{ edpm_sshd_server_options }}
 
+    - name: Set sshd motd when enabled
+      ansible.builtin.lineinfile:
+        path: /etc/pam.d/sshd
+        regexp: "^session.*optional.*pam_motd.so"
+        line: "session    optional     pam_motd.so motd=/etc/motd"
```
slagle, line 83: LGTM, i already have the line in this file in my RHEL 9.2 edpm node, but without the `motd=/etc/motd`, that is the default, so that part is likely not needed. But, I'm fine either way.

slagle (approved):

roles/edpm_sshd/tasks/configure.yml
bshephar: Yeah, so the issue this solves is that `PrintMotd` defaults to `yes` in `sshd_config`. So if we don't explicitly set it to `no` then the motd is duplicated.

Explicitly telling pam the exact file to use results in the single message being printed instead of the duplication. But, I did some more digging to make sure I was answering the questions on the BZ accurately, and I actually think that this might be a better solution to the problem:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/838

PR #832, opened 2024-12-02
Update the doc content for edpm_network_config

vcandapp:


docs/source/roles/role-edpm_network_config.rst
```diff
@@ -1,10 +1,33 @@
-=============================
+============================
 Role - edpm_network_config
-=============================
+============================
 
-.. note::
-  When the `edpm_network_config_tool` is set to `'os-net-config'`, the `ctlplane_gateway_ip` and `ctlplane_ip`
-  variables must be set on the host.
```
slagle, line 7: is this no  longer true?

docs/source/roles/role-edpm_network_config.rst
```diff
@@ -1,10 +1,33 @@
...
+~~~~~
 
-.. include::
-  ../collections/osp/edpm/edpm_network_config_role.rst
+This Ansible role does the following tasks:
+
+* Read the configured edpm_network_config_tool
+  The following choices can be used to configure the host network:
+  - nmstate, i.e based on systemroles.network
+  - os-net-config, i.e based on custom tasks
```
slagle, line 13: should we say which is the default?

docs/source/roles/role-edpm_network_config.rst
```diff
@@ -1,10 +1,33 @@
...
+
+* For the latter, prepare to running os-net-config tool in the host
+  It creates necessary folder/files for rendering the network template
+  and the NIC mapping (optional) in the host
+  It also checks for presence of required RPMs
```
slagle, line 18: I suggest rephrasing to:

```
For os-net-config, the role prepares the host by:
  - creating necessary folders and files for rendering network templates and NIC mappings (optional)
  - Checks for the presence of required RPMS
```

vcandapp: Updated

docs/source/roles/role-edpm_network_config.rst
vcandapp: Done

docs/source/roles/role-edpm_network_config.rst
vcandapp: True - but it will be applicable to both "tools" I guess.
Also, this param was originally part of undercloud.conf - now it is defined as part of network definition. For rhoso, still not sure if user will use same control plane for ocp/edpm nodes. Removed it to avoid confusion.

slagle: that makes sense, thank you

slagle (approved):

PR #834, opened 2024-12-03
Healthcheck for Kepler

yadneshk:
Since Kepler container image doesn't have either "ss" or "netstat" commands, check it's health by making curl requests to port 8888.


yadneshk:
recheck

slagle (approved):

roles/edpm_telemetry_power_monitoring/files/healthchecks/exporter/healthcheck
```diff
@@ -15,4 +15,24 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
-ss -lnp | grep $1 || netstat -plnt | grep -i listen | grep $1
```
paramite, line 18: You cannot just delete script which is used for other exporters. You need to create new script file and configure it's usage accordingly at https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_telemetry/defaults/main.yml#L45

paramite (changes requested):

PR #836, opened 2024-12-04
Fix blocked update because of health checks deployment 

paramite:
- Disables podman_exporter deployment by default

    We need to disable podman_exporter deployment by default until we get
    the exporter image downstream.

    Resolves: OSPRH-12089

- Moves health check script deployment to run.yml

    Minor updates are broken due to the fact that health check scripts
    are deployed in install.yml task lists while the update role is using
    run.yml task lists

    Resolves: OSPRH-12088


roles/edpm_iscsid/tasks/run.yml
```diff
@@ -52,7 +56,6 @@
     state: restarted
 
 - name: Remove iscsid container restart sentinel file
-  become: true
```
slagle, line 55: this is created with `become: true` in the handler. why drop it here?

roles/edpm_telemetry/defaults/main.yml
```diff
@@ -49,3 +49,7 @@ edpm_telemetry_healthcheck_sources:
 #  kepler: exporter
 # If telemetry services should have health checks enabled
 edpm_telemetry_healthcheck: true
+# List of ceilometer agents to be stopped during EDPM adoption
+edpm_telemetry_enabled_exporters:
+  - ceilometer_agent_compute
+  - node_exporter
```
slagle, line 55: need EOF markers

roles/edpm_iscsid/tasks/run.yml
paramite: No reason, I must have removed it accidentally.

roles/edpm_telemetry/defaults/main.yml
paramite: ack

slagle (approved):

PR #840, opened 2024-12-06
Prevent NM from automatic (DHCP) configuration on ethernet devices

vcandapp:
When edpm_network_config role uses os-net-config to perform host network configuration, duplicate DHCP requests from NM are causing connectivity issue. Eg.

https://issues.redhat.com/browse/OSPRH-9142

This fix prevents NM from configure DHCP by default

This PR replaces an older [PR](https://github.com/openstack-k8s-operators/edpm-ansible/pull/720)

### DIFF
--- a/roles/edpm_network_config/tasks/main.yml
+++ b/roles/edpm_network_config/tasks/main.yml
@@ -46,7 +46,25 @@
     - name: Load system-roles.network tasks [nmstate]
       ansible.builtin.include_role:
         name: "{{ lookup('ansible.builtin.env', 'EDPM_SYSTEMROLES', default='fedora.linux_system_roles') + '.network' }}"
-- name: Load edpm_network_config tasks [os-net-config]
-  ansible.builtin.include_tasks:
-    file: network_config.yml
+
+- name: Disable auto-configuration of all interfaces by NetworkManager
   when: edpm_network_config_tool == 'os-net-config'
+  become: true
+  block:
+    - name: Set 'no-auto-default' in /etc/NetworkManager/NetworkManager.conf
+      community.general.ini_file:
+        path: /etc/NetworkManager/NetworkManager.conf
+        state: present
+        mode: "0644"
+        no_extra_spaces: true
+        section: main
+        option: no-auto-default
+        value: "*"
+        backup: true
+    - name: Restart NetworkManager
+      ansible.builtin.systemd:
+        name: NetworkManager
+        state: restarted
+    - name: Load edpm_network_config tasks for os-net-config tool
+      ansible.builtin.include_tasks:
+        file: network_config.yml
--- a/playbooks/configure_network.yml
+++ b/playbooks/configure_network.yml
@@ -12,7 +12,7 @@
         name: osp.edpm.edpm_ovs
       tags:
         - edpm_ovs
-    # Sets up network using OVS
+    # Sets up EDPM network using os-net-config
     - name: Import edpm_network_config
       ansible.builtin.import_role:
         name: osp.edpm.edpm_network_config
--- a/roles/edpm_network_config/molecule/default/converge.yml
+++ b/roles/edpm_network_config/molecule/default/converge.yml
@@ -33,5 +33,6 @@
               next_hop: 192.168.180.1
     edpm_network_config_manage_service: false
     edpm_network_config_hide_sensitive_logs: false
+    edpm_network_config_nmstate: false
   roles:
     - role: "osp.edpm.edpm_network_config"
--- a/roles/edpm_network_config/molecule/default/prepare.yml
+++ b/roles/edpm_network_config/molecule/default/prepare.yml
@@ -22,6 +22,7 @@
       test_deps_setup_edpm: true
       test_deps_extra_packages:
         - iproute
+        - NetworkManager
 - name: Prepare
   hosts: all
   gather_facts: false
@@ -34,7 +35,7 @@
         name: network-scripts
         state: present
       when:
-        - ansible_facts['distribution_major_version'] is version('8', '==')
+        - ansible_facts['distribution_major_version'] is version('8', '>=')
     - name: Create a dummy network interface
       ansible.builtin.command: "ip link add dummy0 type dummy"
       register: ip_command_output


vcandapp:
recheck

slagle (approved):

bshephar (approved):

bshephar:
recheck

vcandapp:
recheck

roles/edpm_network_config/tasks/main.yml
```diff
@@ -46,7 +46,26 @@
     - name: Load system-roles.network tasks [nmstate]
       ansible.builtin.include_role:
         name: "{{ lookup('ansible.builtin.env', 'EDPM_SYSTEMROLES', default='fedora.linux_system_roles') + '.network' }}"
-- name: Load edpm_network_config tasks [os-net-config]
-  ansible.builtin.include_tasks:
-    file: network_config.yml
-  when: edpm_network_config_tool == 'os-net-config'
+
+- name: Disable auto-configuration of all interfaces by NetworkManager
+  when: ( edpm_bootstrap_network_service == 'NetworkManager' ) and
```
rabi, line 51: Where is `edpm_bootstrap_network_service`  var set or default defined? This var does not exist by default. Also the code does not look correct to me. When  `edpm_bootstrap_network_service` is anything other than 'NetworkManager` os-net-config won't run?

rabi (changes requested):

roles/edpm_network_config/tasks/main.yml
bshephar: That variable appears to be set in the bootstrap role:
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_bootstrap/defaults/main.yml#L62

So, I assumed they were using it from there.

The same conditional check is also used in the Bootstrap role by the looks of it:
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_bootstrap/tasks/bootstrap.yml#L73

I'm not sure of the background for this variable though. Just an observation in response to this comment. Maybe we can take this opportunity to evaluate our use of the variable to ensure it still makes sense.

vcandapp: Also the code does not look correct to me. When `edpm_bootstrap_network_service` is anything other than 'NetworkManager` os-net-config won't run?

- os-net-config with ifcfg scripts doesn't have any dependency on NetworkManager.
- But, the reason for adding this check was, if os-net-config is used for deployment and bootstrap used "NetworkManager" during cloud-init, we need to disable the auto dhcp

vcandapp: > > So, I assumed they were using it from there.
> 
> That's a role var and it won't be availale for other roles or other playbook runs. Also, every service is run as a separate playbook, so that won't be available in this role. 

@rabi , will referring like this fix it?
- https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_derive_pci_device_spec/defaults/main.yml#L26
(neutron_sriov_agent default is referred in derive_pci role)

rabi: > @rabi , will referring like this fix it?

I'm not sure what you mean by that. This has to have a role default like in bootstrap role or  set as a fact or ansible var when the playbook is called.

slagle: The `edpm_bootstrap_network_service` variable doesn't look like it's used correctly even in edpm_bootstrap, as it defaults to NetworkManager, which will then be enabled after installing the `network-scripts` package.

Honestly, this commit looks pretty suspect that changed the default from `network` to `NetworkManager`:

https://github.com/openstack-k8s-operators/edpm-ansible/commit/31b1b0297d8c6f4121bcb9682008647666d3900c

I don't see a compelling reason why we want to enable `NetworkManager` after installing `network-scripts`.

Perhaps we should track this issue down from that angle, b/c it's really the fact that `NetworkManager` is enabled that is the issue.

vcandapp: The above commits seems to be related to [Only use the network service for EL 7](https://github.com/openstack-archive/tripleo-ansible/commit/964fa4ddef95bcaa184e81d0565a521aeb7c3975) , since NM is by default enabled >=rhel8

Maybe for now I can safely assume "edpm_bootstrap_network_service" is NetworkManager and remove the check for CI to pass

vcandapp: This is the [patch](https://code.engineering.redhat.com/gerrit/c/tripleo-ansible/+/453443) we need to fix for edpm also

rabi: > Maybe for now I can safely assume "edpm_bootstrap_network_service" is NetworkManager and remove the check for CI to pass

Can you answer the question @slagle asked? Should we not stop/disable NetworkManager when using network-scripts. I understand that we've been keeping it enabled and running since TripleO days and doing these hacks to prevent it doing it's things, but do we really need it running.

vcandapp: @slagle / @rabi, there are few reasons for not disabling NM as of now:
- os-net-config can use nmstate provider in future, which will eventually need NM
- Disabling/stopping NM might need a wider review and QE scope
- The issue we try to fix is the flood of DHCP requests (esp. for SR-IOV VFs) during node provision, which leads to timeout 
- SR-IOV VFs are treated as individual i/f by NM and DHCP requests are sent out for all on their creation

cc @karthiksundaravel DanSneddon for further comments

karthiksundaravel: The network-scripts has a variable NM_CONTROLLED, which could be used to enable or disable NetworkManager for that interface. All we need now is to prevent NM from configuring the devices which are not configured or yet to be configured by network scripts. Given this requirement, setting no-auto-default=* does the job well.  This configuration should hold good irrespective of the os-net-config provider (ifcfg or nmstate) as well.

vcandapp:
recheck

dsneddon:
As @vcandapp mentioned NM will cause duplicate DHCP requests, so this fix will prevent NM from automatically creating DHCP requests when they are not needed. This will improve the behavior that we see and ensure that NM is operating in a way that does not conflict with os-net-config. It is understandable that NM would create DHCP enabled configurations by default for interfaces for normal use cases, but for cloud we really need for it to get out of the way and let the automation work as designed.

roles/edpm_network_config/tasks/main.yml
dsneddon: Yes, I agree with the comments from @karthiksundaravel and we already have seen this causing issues in the field. The only way to be sure that things behave as expected is for auto-default to be turned off in the configuration as it isn't needed for this use case anyway (cloud-init will do what is required, but NM doesn't need to step in and create configurations for unconfigured interfaces).

vcandapp:
recheck

rabi (commented):
+2 I'm fine with the current change.

rabi:
+2

vcandapp:
receck

slagle (approved):

PR #843, opened 2024-12-09
Document nmstate as experimental and require an override to use it

slagle:
"edpm_network_config_tool: nmstate" is experimental and officially
unsupported. This change will require users to explicitly opt-in to
using nmstate directly instead of using os-net-config+nmstate. It adds a
task that will fail unless edpm_network_config_tool_nmstate_override is
set to true (defaults to false).

Jira: https://issues.redhat.com/browse/OSPRH-12316

### DIFF
--- a/roles/edpm_network_config/defaults/main.yml
+++ b/roles/edpm_network_config/defaults/main.yml
@@ -22,10 +22,13 @@
 # This variable let the use choose which tool will be used to configure networking on hosts
 # Accepted values are:
 # * os-net-config (default) -> the os-net-config will be used
-# * nmstate -> the network role from linux-system-roles/rhel-system-roles will be used to
+# * nmstate (experimental, unsupported) -> the network role from linux-system-roles/rhel-system-roles will be used to
 #   configure networking on hosts. In particular, the nmstate tool should cover all the
 #   supported scenario. Refer to nmstate.io for configuration snippets.
 edpm_network_config_tool: os-net-config
+
+# Must be set to true to allow usage of nmstate as edpm_network_config_tool
+edpm_network_config_tool_nmstate_override: false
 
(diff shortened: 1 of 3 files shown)


bshephar (commented):
Do you think more things like this are likely to come up in the future?

I wonder if a better approach might be to have a dictionary called something like `edpm_tech_preview`. Then allow users to enable tech preview features be appending them to the list?

Obviously, this isn't required if we're only doing it for `nmstate`. But I do wonder how many more of these things will come up in the future and how cumbersome it will become to individually manage them on a per role basis?

Similar to feature gates I guess:
https://docs.openshift.com/container-platform/4.17/nodes/clusters/nodes-cluster-enabling-features.html

https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

Could be overkill for our requirements, but putting it out there for consideration.

slagle:
> Do you think more things like this are likely to come up in the future?
> 
> I wonder if a better approach might be to have a dictionary called something like `edpm_tech_preview`. Then allow users to enable tech preview features be appending them to the list?
> 
> Obviously, this isn't required if we're only doing it for `nmstate`. But I do wonder how many more of these things will come up in the future and how cumbersome it will become to individually manage them on a per role basis?
> 
> Similar to feature gates I guess: https://docs.openshift.com/container-platform/4.17/nodes/clusters/nodes-cluster-enabling-features.html
> 
> https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
> 
> Could be overkill for our requirements, but putting it out there for consideration.

Yes, I could see adding a field like this on the NodeSet, which would set an ansible var to enable certain features. I think it would consider it separately from this PR

bshephar (approved):

vcandapp:
ltgm

PR #846, opened 2024-12-13
Stop and disable iscsi-starter.service

knoha-rh:
When an EDPM node rebooted which runs a VM with iscsi-backed volume, 
it leaves node records under /var/lib/iscsi/nodes. 
While the records exist, iscsi-starter.service runs iscsi.service. 
Then it starts iscsid.service also.

To prevent this issue, we need to stop and disable iscsi-starter.service.

JIRA:[OSPRH-12372](https://issues.redhat.com/browse/OSPRH-12372)

### DIFF
--- a/roles/edpm_iscsid/tasks/install.yml
+++ b/roles/edpm_iscsid/tasks/install.yml
@@ -46,12 +46,15 @@
     - name: Gather services facts
       ansible.builtin.service_facts:
 
-    - name: Stop iscsi.service
+    - name: Stop and disable iscsi.service and iscsi-starter.service
       ansible.builtin.systemd:
-        name: iscsi.service
+        name: "{{ item }}"
         state: stopped
         enabled: false
       when:
-        - ansible_facts.services["iscsi.service"] is defined
-        - ansible_facts.services["iscsi.service"]["status"] != "not-found"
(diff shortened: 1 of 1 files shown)


roles/edpm_iscsid/tasks/install.yml
```diff
@@ -46,12 +46,16 @@
...
       ansible.builtin.systemd:
-        name: iscsi.service
+        name: "{{ item }}"
         state: stopped
         enabled: false
       when:
-        - ansible_facts.services["iscsi.service"] is defined
-        - ansible_facts.services["iscsi.service"]["status"] != "not-found"
-        - ansible_facts.services["iscsi.service"]["status"] == "enabled"
+        - ansible_facts.services["item"] is defined
```
rabi, line 55: I guess you've to remove the quotes from "item" here.

roles/edpm_iscsid/tasks/install.yml
```diff
@@ -46,12 +46,16 @@
...
+        - ansible_facts.services["item"]["status"] != "not-found"
+        - ansible_facts.services["item"]["status"] == "enabled"
+      loop:
+        - iscsi.service
+        - iscsi-starter.service
+
```
rabi, line 61: And the redundant line here.

bshephar (approved):

PR #848, opened 2025-01-03
Making check for mandatory ovs_dpdk parms

vcandapp:


roles/edpm_ovs_dpdk/tasks/configure.yml
```diff
@@ -19,6 +19,11 @@
     msg: "List of PMD cores cannot be empty - edpm_ovs_dpdk_pmd_core_list"
   when: not edpm_ovs_dpdk_pmd_core_list|string or edpm_ovs_dpdk_pmd_core_list == 'null'
 
+- name: Check valid input for edpm_ovs_dpdk_socket_memory
+  ansible.builtin.fail:
+    msg: "Socket Memory cannot be empty - edpm_ovs_dpdk_socket_memory"
+  when: not edpm_ovs_dpdk_socket_memory|string or edpm_ovs_dpdk_socket_memory == 'null'
```
Jaganathancse, line 25: @vcandapp edpm_ovs_dpdk_socket_memory is "" default, so null check is not required.
Instead of failing, can we make default value for this parameter edpm_ovs_dpdk_socket_memory ?

roles/edpm_ovs_dpdk/meta/argument_specs.yml
```diff
@@ -22,7 +21,6 @@ argument_specs:
         description: OvS DPDK extra configs
       edpm_ovs_dpdk_socket_memory:
         type: str
-        default: ""
```
Jaganathancse, line 25: @vcandapp 
Need to remove in default main.yaml file if we need to remove default value.

Jaganathancse (changes requested):

roles/edpm_ovs_dpdk/tasks/configure.yml
vcandapp: https://github.com/openstack-archive/tripleo-ansible/blob/stable/wallaby/tripleo_ansible/playbooks/cli-overcloud-openvswitch-dpdk.yaml#L22, even though they are mandatory - it is set to NULL

vcandapp: Like in earlier deployment, https://github.com/openstack-archive/tripleo-ansible/blob/stable/wallaby/tripleo_ansible/playbooks/cli-overcloud-openvswitch-dpdk.yaml#L22 we set it to NULL even for mandatory params

roles/edpm_ovs_dpdk/meta/argument_specs.yml
vcandapp: Already removed

vcandapp:
recheck

PR #853, opened 2025-01-14
Only test modified roles

fao89:
closes [OSPRH-11820](https://issues.redhat.com//browse/OSPRH-11820)

### DIFF
--- a/.github/workflows/molecule.yaml
+++ b/.github/workflows/molecule.yaml
@@ -55,13 +55,25 @@ jobs:
     - uses: actions/setup-python@v5
       with:
         python-version: '3.9'
+    - name: Check whether the role is modified
+      id: check_role
+      run: |
+        if [[ $(git diff --name-only roles | grep "${{ matrix.tested_role }}/") ]]; then
+          echo "modified=true" >> $GITHUB_OUTPUT
+        else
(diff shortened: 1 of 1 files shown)


.github/workflows/molecule.yaml
```diff
@@ -55,13 +55,25 @@ jobs:
     - uses: actions/setup-python@v5
       with:
         python-version: '3.9'
+    - name: Check whether the role is modified
+      id: check_role
+      run: |
+        if [[ $(git diff --name-only roles | grep ${{ matrix.tested_role }}) ]]; then
```
slagle, line 61: if you grep for the role name followed by a `/` here, then you won't get substring matches. For instance, edpm_telemetry would also match edpm_telemetry_logging with the way it is now.

fao89: nice catch!

slagle (approved):

.github/workflows/molecule.yaml
karelyatin: also should have been "git show" , being checked in https://github.com/openstack-k8s-operators/edpm-ansible/pull/909

karelyatin: Also what if some base roles are modified, which are used by other roles, shouldn't we execute all the dependent roles?

PR #866, opened 2025-01-28
Openvswitch package install required

Jaganathancse:
This change is to fix the openvswitch package missing
issue during configure ovs dpdk for pre provisioned
EDPM nodes.

We already have ovs packages installation, ovs services
are enabled and running part in configure-network which
is common for all deployment.

To avoid issue in configure ovs dpdk, we are adding only
openvswitch package installation part and not using
existing edpm_ovs install.yml task to avoid complete
redundat like ovs services configuration, enable
and running part.

### DIFF
--- a/playbooks/configure_ovs_dpdk.yml
+++ b/playbooks/configure_ovs_dpdk.yml
@@ -5,6 +5,12 @@
   strategy: linear
   gather_facts: "{{ gather_facts | default(false) }}"
   tasks:
+    - name: Install openvswitch package
+      become: true
+      ansible.builtin.dnf:
+        name: openvswitch
+        state: present
+
     - name: Configure OvS DPDK configs
       ansible.builtin.import_role:
         name: osp.edpm.edpm_ovs_dpdk
--- a/roles/edpm_bootstrap/meta/argument_specs.yml
+++ b/roles/edpm_bootstrap/meta/argument_specs.yml
@@ -21,7 +21,6 @@ argument_specs:
           - jq
           - nftables
           - openstack-selinux
-          - os-net-config
           - python3-libselinux
           - python3-pyyaml
           - rsync


roles/edpm_bootstrap/defaults/main.yml
```diff
@@ -33,6 +33,8 @@ edpm_bootstrap_packages_bootstrap:
   - nftables
   - NetworkManager
   - openstack-selinux
+  - openvswitch
```
slagle, line 36: openvswitch should get installed by edpm_ovs, which is included by playbooks/configure_network.yml

roles/edpm_bootstrap/defaults/main.yml
```diff
@@ -33,6 +33,8 @@ edpm_bootstrap_packages_bootstrap:
...
+  - os-net-config
```
slagle, line 37: already installed by edpm_network_config, which actually includes edpm_bootstrap/tasks/packages.yml, and then the very next task is to install os-net-config, so this could be redundant.

slagle (changes requested):
can you elaborate or share details on the issue that's being fixed? this looks like it could be redundant given the other tasks that are already in place.

roles/edpm_bootstrap/defaults/main.yml
Jaganathancse: yes installed openvswitch during configure nework but failing for pre provisioned node. 
we need to install in configure ovs dpdk which is called before configure network. we cannot change the order.
so to avoid redundant, i am going to include edpm_ovs role in configure_ovs_dpdk.yml playbook.
this is only duplication for ovs dpdk. for other features only via configire_network.yml playbook.

roles/edpm_bootstrap/defaults/main.yml
Jaganathancse: but os-net-config mantioned in meta default arguments_spec.yaml,
can i remove this os-net-config also in arguments_spec default?

playbooks/configure_ovs_dpdk.yml
```diff
@@ -5,6 +5,12 @@
   strategy: linear
   gather_facts: "{{ gather_facts | default(false) }}"
   tasks:
+    - name: Install openvswitch package
+      become: true
+      ansible.builtin.dnf:
+        name: openvswitch
+        state: present
+
```
slagle, line 13: would you want to use edpm_ovs/tasks/install.yml here instead? The other task that would be done is to ensure that the openvswitch is started. Does dpdk require that?

Jaganathancse: using  edpm_ovs/tasks/install.yml in configure ovs dpdk playbook, will create the enabled and running ovs part duplication also.
Configure ovs dpdk is require only openvswitch package installation and not to be enabled and running.
configure network is common for all feature and configure ovs dpdk is specific to only ovs dpdk.
better we can add only openvswitch installation here and we can avoid redundant ovs configuration and enabled part if we are not using ovs install.yml here.

slagle (approved):

Jaganathancse:
recheck

Jaganathancse:
recheck

Jaganathancse:
recheck

bshephar:
recheck

PR #869, opened 2025-01-29
Apply configuration tasks for Nova on updates

bogdando:
When 'updates' service runs, it only calls install.yml from Nova service role, which is not sufficient to provide required configuration updates. Inlcude configure.yml as well to cover that for Nova service.

To properly address this for *all services* we should add migration/update tasks there, which can be called individually from the update playbook.

Jira: #[OSPRH-13415](https://issues.redhat.com/browse/OSPRH-13415)

### DIFF
--- a/roles/edpm_update/tasks/containers.yml
+++ b/roles/edpm_update/tasks/containers.yml
@@ -100,6 +100,15 @@
     - edpm_update
   when: '"nova" in edpm_update_running_services'
 
+- name: Updates configs for edpm_nova role
+  ansible.builtin.include_role:
+    name: osp.edpm.edpm_nova
+    tasks_from: configure.yml
+  tags:
+    - edpm_nova
+    - edpm_update
+  when: '"nova" in edpm_update_running_services'
+
 - name: Updates containers for edpm_neutron_sriov role
   ansible.builtin.include_role:
     name: osp.edpm.edpm_neutron_sriov


mrkisaolamb (commented):
lgtm, also errors in ci looks unrelated

roles/edpm_update/tasks/containers.yml
```diff
@@ -100,6 +100,17 @@
     - edpm_update
   when: '"nova" in edpm_update_running_services'
 
+- name: Updates configs for edpm_nova role
+  ansible.builtin.include_role:
+    name: osp.edpm.edpm_nova
+    tasks_from: configure.yml
+    apply:
+      become: true
```
bshephar, line 108: Is this actually necessary? I can see there are already some tasks in the `configure.yml` that have `become: true` as required. ie:
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_nova/tasks/configure.yml#L9

bogdando: I put it here as all other tasks are using it :)
I am not sure if this won't break resulted permissions.
How does edpm service executed by ansible runner, does it enfore become?
We should call this the same way we do for greenfield deployments

slagle: I don't think it's needed. It's not set when configure.yml is included from edpm_nova/tasks/main.yml, it's not set in playbooks/nova.yml, and there is nothing in how openstack-operator creates the runner jobs that sets a universal become:true.

slagle (approved):

bshephar (approved):

bshephar:
The VM couldn't be reached in the adoption job:
~~~
TASK [dataplane_adoption : verify connectivity to the existing test VM instance using Floating IP] ***
fatal: [localhost]: FAILED! => {"changed": true, "cmd": "ping -c4 192.168.122.20\n", "delta": "0:00:13.111819", "end": "2025-01-31 13:51:15.202861", "msg": "non-zero return code", "rc": 1,
~~~

bshephar:
recheck

PR #871, opened 2025-01-31
Rebase bootc branch onto main

slagle:
- **Move containers-tmpwatch to /var/lib/openstack**
- **Gather local facts where ansible_local.bootc is used**
- **Drop owner/group from /var/lib/openstack**


slagle:
opened against wrong branch

PR #872, opened 2025-01-31
Rebase bootc branch onto main

slagle:
- **Update actions**
- **Add noop job to projects**
- **Move containers-tmpwatch to /var/lib/openstack**
- **Gather local facts where ansible_local.bootc is used**
- **Drop owner/group from /var/lib/openstack**

### DIFF
--- a/roles/edpm_bootstrap/tasks/bootstrap.yml
+++ b/roles/edpm_bootstrap/tasks/bootstrap.yml
@@ -60,8 +60,9 @@
 - name: Execute bootstrap command
   ansible.builtin.import_tasks: bootstrap_command.yml
 
-- name: Import packages tasks
-  ansible.builtin.import_tasks: packages.yml
+- name: Include packages tasks
+  ansible.builtin.include_tasks: packages.yml
+  when: not ansible_local.bootc
 
 - name: Set selinux state
   ansible.posix.selinux:
@@ -107,8 +108,9 @@
       #!/bin/sh
       make_resolv_conf() { : ; }
 
-- name: Configure swap
-  ansible.builtin.import_tasks: swap.yml
+- name: Include swap tasks
+  ansible.builtin.include_tasks: swap.yml
+  when: not ansible_local.bootc
 
 - name: FIPS tasks
   ansible.builtin.import_tasks: fips.yml
--- a/roles/edpm_bootstrap/tasks/swap.yml
+++ b/roles/edpm_bootstrap/tasks/swap.yml
@@ -16,6 +16,7 @@
 
 - name: Configure swap file
   when:
+    - not ansible_local.bootc
     - not edpm_bootstrap_swap_partition_enabled|bool
     - edpm_bootstrap_swap_size_megabytes|int > 0
   become: true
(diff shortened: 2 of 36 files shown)


roles/edpm_bootstrap/tasks/bootstrap.yml
```diff
@@ -107,8 +108,9 @@
       #!/bin/sh
       make_resolv_conf() { : ; }
 
-- name: Configure swap
-  ansible.builtin.import_tasks: swap.yml
+- name: Include swap tasks
+  ansible.builtin.include_tasks: swap.yml
+  when: not ansible_local.bootc
```
bogdando, line 113: @SeanMooney PTAL. Why can't we configure swap with bootc?

slagle:
> we should always have swap configured on EDPM

this is just rebasing the bootc development branch onto main, not merging code into main. we'll address this before merging the bootc branch into main.

roles/edpm_bootstrap/tasks/swap.yml
```diff
@@ -16,6 +16,7 @@
 
 - name: Configure swap file
   when:
+    - not ansible_local.bootc
```
SeanMooney, line 19: we requrie swap for  all computre nodes, how will that requirement be met for bootc?

roles/edpm_logrotate_crond/tasks/configure.yml
```diff
@@ -22,8 +22,6 @@
         path: "/var/lib/openstack"
         mode: 0755
         state: directory
-        owner: "{{ ansible_user }}"
-        group: "{{ ansible_user }}"
```
SeanMooney: /var/lib/openstack  is not ment to be owned by root
its meant to be owned by the ansible user so this is a bug.

roles/edpm_bootstrap/tasks/bootstrap.yml
SeanMooney: swap is required for all comptue nodes so we will require swap even when using bootc

there are som either issues with this patch too.

for example it has regressed the ownership fo /var/lib/openstack which should always be owned by the ansibel user, never root

we have fixed that in the past on other roles like neutron.
https://github.com/openstack-k8s-operators/edpm-ansible/commit/e0a6307e1e9cfcb416d8928f51279b6ada27d9d8

SeanMooney:
> > we should always have swap configured on EDPM
> 
> this is just rebasing the bootc development branch onto main, not merging code into main. we'll address this before merging the bootc branch into main.

ah i see its rather confusing to have a separate feature development branch on the main repo like this but looking at the source and dest branches i see this is not the final content.

i assume you choose to do it this way as a means to share content with others before its merged

slagle:
> > > we should always have swap configured on EDPM
> > 
> > 
> > this is just rebasing the bootc development branch onto main, not merging code into main. we'll address this before merging the bootc branch into main.
> 
> ah i see its rather confusing to have a separate feature development branch on the main repo like this but looking at the source and dest branches i see this is not the final content.
> 
> i assume you choose to do it this way as a means to share content with others before its merged

we just needed to change the path of the swap file for bootc. I believe this will address it:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/873

PR #873, opened 2025-02-04
Re-enable swap for bootc

slagle:
The default path for a swapfile needs to be under a writeable location
for bootc. Use /var/swap. The value of the existing variable,
edpm_bootstrap_swap_path can't just be changed since there are already
existing deployments, so a new variable, edpm_bootstrap_swap_path_bootc
is added for bootc. An ansible task will select which variable to use
based on the type of system.

### DIFF
--- a/playbooks/bootstrap.yml
+++ b/playbooks/bootstrap.yml
@@ -15,6 +15,9 @@
   any_errors_fatal: "{{ edpm_any_errors_fatal | default(true) }}"
   max_fail_percentage: "{{ edpm_max_fail_percentage | default(0) }}"
   tasks:
+    - name: Gather ansible_local facts
+      ansible.builtin.setup:
+        filter: ansible_local
     - name: Grow volumes
       ansible.builtin.import_role:
         name: osp.edpm.edpm_growvols
--- a/roles/edpm_bootstrap/defaults/main.yml
+++ b/roles/edpm_bootstrap/defaults/main.yml
@@ -72,6 +72,9 @@ edpm_bootstrap_selinux_mode: enforcing
 # Swap management
 edpm_bootstrap_swap_size_megabytes: 1024
 edpm_bootstrap_swap_path: /swap
+# swap path needs to be under /var on bootc, so a separate default value is
+# needed.
+edpm_bootstrap_swap_path_bootc: /var/swap
 edpm_bootstrap_swap_partition_enabled: false
 edpm_bootstrap_swap_partition_label: swap1
 
--- a/roles/edpm_bootstrap/tasks/swap.yml
+++ b/roles/edpm_bootstrap/tasks/swap.yml
@@ -16,26 +16,29 @@
 
 - name: Configure swap file
   when:
-    - not ansible_local.bootc
     - not edpm_bootstrap_swap_partition_enabled|bool
     - edpm_bootstrap_swap_size_megabytes|int > 0
   become: true
   block:
+    - name: Set swap path
+      set_fact:
+        edpm_bootstrap_swap_path_value: |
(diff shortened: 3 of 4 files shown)


roles/edpm_bootstrap/tasks/swap.yml
```diff
@@ -16,26 +16,33 @@
...
 - name: Configure swap file
   when:
-    - not ansible_local.bootc
     - not edpm_bootstrap_swap_partition_enabled|bool
     - edpm_bootstrap_swap_size_megabytes|int > 0
   become: true
   block:
+    - name: Set swap var to use (bootc)
+      set_fact:
+        edpm_bootstrap_swap_path_value: "{{ edpm_bootstrap_swap_path_bootc }}"
```
bshephar, line 25: Didn't want to do this with Jinja2 instead of the two conditional tasks?

playbooks/bootstrap.yml
```diff
@@ -15,6 +15,9 @@
   any_errors_fatal: "{{ edpm_any_errors_fatal | default(true) }}"
   max_fail_percentage: "{{ edpm_max_fail_percentage | default(0) }}"
   tasks:
+    - name: Gather ansible_local facts
+      ansible.builtin.setup:
+        filter: ansible_local
```
bshephar, line 20: Yeah, we should realistically do this at the start of each of our service playbooks:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/874

bogdando (commented):
lgtm

roles/edpm_bootstrap/defaults/main.yml
```diff
@@ -72,6 +72,9 @@ edpm_bootstrap_selinux_mode: enforcing
 # Swap management
 edpm_bootstrap_swap_size_megabytes: 1024
 edpm_bootstrap_swap_path: /swap
+# swap path needs to be under /var on bootc, so a separate default value is
+# needed.
+edpm_bootstrap_swap_path_bootc: /var/swap
```
SeanMooney, line 77: so other then the upgrade impact of needing to move the swap file
there is no reason we could not use /var/swap for all deployments.

we may want to consider doing that.

for example  we could define new vars like this
```
edpm_bootstrap_legacy_swap_path: /swap
edpm_bootstrap_swap_path: /var/swap
```
then check if edpm_bootstrap_legacy_swap_path exists and edpm_bootstrap_swap_path does not.
if that is the case check fi there is enough space to allocate the second swap file, do swap on to the new location and then swappoff the old location and delete the legacy swap file.

this can be a followup pr and could be deferred to the next reboot also.

but that would give us a way to normalise the two flows.

bshephar: It would indeed be nice to bring consistency to both types of RHEL nodes. Just a small inconvenience during the transition

bshephar (approved):

roles/edpm_bootstrap/defaults/main.yml
bogdando: @bshephar could you please create a jira tracker for that?

SeanMooney: i would suggeest 

1 create new swap file
2 swapon new file
3 swapoff old file
4 delete old file

that should alow the kernel to move the swap pages between the fiels without impacting the workloads.

has the issue with 0 downtime update been fixed in bootc yet?
if not then any updates to the bootc image woudl need you to drain the compute node first
in which case the swap file should not be actively used so you should be able to just nuke the old one and create the new one in image mode as @bshephar suggets.

PR #876, opened 2025-02-07
Removing DPDK lcore list config

Jaganathancse:
This edpm_ovs_dpdk_lcore_list config is not
required now to enable ovs dpdk in EDPM
nodes.

### DIFF
--- a/docs/source/roles/role-edpm_ovs_dpdk.rst
+++ b/docs/source/roles/role-edpm_ovs_dpdk.rst
@@ -22,7 +22,6 @@ This Ansible role allows to do the following tasks:
         name: "osp.edpm.edpm_ovs_dpdk"
       vars:
          edpm_ovs_dpdk_pmd_core_list: "1,13,3,15"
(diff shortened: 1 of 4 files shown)


rjarry:
It is not only "not needed" it is "strongly discouraged and should never be used" :)

rjarry (changes requested):
It looks like you missed a spot left here:

https://github.com/openstack-k8s-operators/edpm-ansible/blob/898895d4128cae639b714108fc6d87bf214c83fc/docs/source/roles/role-edpm_ovs_dpdk.rst?plain=1#L25

Jaganathancse:
> It looks like you missed a spot left here:
> 
> https://github.com/openstack-k8s-operators/edpm-ansible/blob/898895d4128cae639b714108fc6d87bf214c83fc/docs/source/roles/role-edpm_ovs_dpdk.rst?plain=1#L25

yes Robin, Removed it. thanks.

rjarry (approved):

slagle (approved):

PR #877, opened 2025-02-10
[OSPRH-12074] Add TLS for powermonitoring

vyzigold:


vyzigold:
This is waiting until kepler version "release-0.7.13" or "release-0.8.1" is released (whatever is sooner) and the respective container images are used in the upstream and downstream CI.

jlarriba:
Closing because change of priorities in Kepler development

roles/edpm_telemetry_power_monitoring/templates/kepler.json.j2
```diff
@@ -4,7 +4,12 @@
     "restart": "always",
     "ports": ["8888:8888"],
     "net": "host",
-    "command": "-v=2",
+    "command": [
+{% if tls_cert_exists|bool %}
+        "--web.config.file=/etc/kepler/web-config.yml",
+{% endif %}
+        "-v=2"
```
rsafrono, line 11: my comment is not directly related to TLS config (has plans to test it as well) but I've tried to deploy power monitoring with kepler 0.10 image and found that "-v=2" causes an error like "unsupported parameter" or something like that. Also, Kepler 0.10 default port differs from 8888, so in order to avoid changing ports above in line 5 and also firewall config in other files I recommend to append "--web.listen-address=:8888" to the command line. This helped me to get kepler metrics on the same port as before without any extra configs.

PR #878, opened 2025-02-11
node_exporter should not be mounting dbus

jlarriba:
The dbus exclusive mount is preventing everything else to restart or to use it.

This solves the route cause for not being able to deploy an edpm twice.

### DIFF
--- a/roles/edpm_telemetry/templates/node_exporter.json.j2
+++ b/roles/edpm_telemetry/templates/node_exporter.json.j2
@@ -8,8 +8,6 @@
 {% if tls_cert_exists|bool %}
         "--web.config.file=/etc/node_exporter/node_exporter.yaml",
 {% endif %}
-        "--collector.systemd",
-        "--collector.systemd.unit-include=(edpm_.*|ovs.*|openvswitch|virt.*|rsyslog)\\.service",
         "--web.disable-exporter-metrics",
         "--no-collector.dmi",
         "--no-collector.entropy",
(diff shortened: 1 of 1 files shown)


vyzigold:
recheck
known issue with test-operator trying to install twice

roles/edpm_telemetry/templates/node_exporter.json.j2
```diff
@@ -39,8 +39,7 @@
     "volumes": [
 {% if tls_cert_exists|bool %}
         "{{ edpm_telemetry_config_dest }}/node_exporter.yaml:/etc/node_exporter/node_exporter.yaml:z",
-        "{{ edpm_telemetry_certs }}:/etc/node_exporter/tls:z",
+        "{{ edpm_telemetry_certs }}:/etc/node_exporter/tls:z"
 {% endif %}
-        "/var/run/dbus/system_bus_socket:/var/run/dbus/system_bus_socket:rw,z"
```
rabi, line 44: Why was it added at the first place?

rabi (approved):
/lgtm

roles/edpm_telemetry/templates/node_exporter.json.j2
jlarriba: I believe it is for this options

"--collector.systemd",
"--collector.systemd.unit-include=(edpm_.*|ovs.*|openvswitch|virt.*|rsyslog)\\.service",

to work. I think this is not needed anymore as it has been replaced by the heartbeat service.

slagle (approved):

roles/edpm_telemetry/templates/node_exporter.json.j2
paramite: Unfortunately it is not superseded, because we don't have all the services on compute nodes containerized any more. We gonna need to get the mount back and enable systemd collector, so that we cover all services' status. I'm gonna investigate to make it work as in the past before the change with selinux labels happened.

PR #881, opened 2025-02-13
Update task name to clarify disconnected env

knoha-rh:
Before this fix, the task name of writing registries.conf is the same between connected env and disconnected env. 
It requires to check the rendered registries.conf for identifying which is used.

To make it clear, this fix updates the name for disconnected ocp.


rabi (approved):
/lgtm

PR #885, opened 2025-02-24
exporters: add dataplane_node_exporter installation

atyronesmith:
The dataplane_node_exporter is a Prometheus exporter that collects networking metrics from OVN, DPDK, and OVS.

[Depends-upon](https://issues.redhat.com/browse/OSPRH-11765)

**DNM** until image is available on quay.io.


atyronesmith:
starting over

PR #624, opened 2024-04-15
Don't allow dhclient update resolv.conf

rabi:

PR #895, opened 2025-03-05
Change edpm_network_config_safe_defaults to false

knoha-rh:
While edpm_network_config_safe_defaults is true, os-net-config tries to configure DHCP to all physical interfaces and update resolv.conf. This behavior should not be default because a user cannot detect the failure by this.

To prevent the uninteded configuration, change the default to false.

Resolve: [OSPRH-14370](https://issues.redhat.com//browse/OSPRH-14370)

### DIFF
--- a/roles/edpm_network_config/defaults/main.yml
+++ b/roles/edpm_network_config/defaults/main.yml
@@ -50,7 +50,7 @@ edpm_network_config_interface_name: nic1
 edpm_network_config_manage_service: true
 edpm_network_config_nmstate: false
 edpm_network_config_os_net_config_mappings: {}
-edpm_network_config_safe_defaults: true
+edpm_network_config_safe_defaults: false
 edpm_network_config_template: ""
 edpm_bond_interface_ovs_options: "bond_mode=active-backup"
 edpm_dns_search_domains: []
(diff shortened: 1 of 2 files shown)


rabi:
We've this default[1] since TripleO days, though I think it's reasonable to change the default to false.  But I would expect if you've https://github.com/openstack-k8s-operators/edpm-ansible/pull/624, it won't update resolv.conf.

 
[1]  https://gitlab.cee.redhat.com/eng/openstack/tripleo-ansible/-/blob/rhos-17.1-trunk-patches/tripleo_ansible/roles/tripleo_network_config/defaults/main.yml?ref_type=heads#L32

slagle (commented):
+2, i think we can change this default

rabi:
To fallback and have networks configured with dhcp to allow ssh for troubleshooting is the reason mentioned[1]. But not sure how useful that has been. 

[1] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/plugins/modules/edpm_os_net_config.py#L228-L229

rabi (approved):
/lgtm

bogdando:
That is a good one.
We can indeed change the default as this safe default feature was introduced by https://review.opendev.org/c/openstack/tripleo-image-elements/+/213748 as a feature, not a bug fix. The feature was specific to BM provisioning design of the past days. So, unsetting the feature flag should not bring back any known issues

PR #896, opened 2025-03-06
Apply updates for edpm_nova role instead of config

bogdando:
Inlcuding configure tasks for nova role during updates doesn't work as there are no required data sources attached to osdpd calling the update service.

Introduce the update tasks which contents would be changing from release to release and include only tasks required to "migrate" the configuration and initialization logic of the nova EDPM service from A to B, then from B to C and so on.

When reproducing d/s, repo-setup fails on rhel during molecule
prepare because there is no RDO for rhel.
    
Specify centos9 distro for repo-setup explicitly.

Fix permissions to molecule CA public cert

Nova compute cannot start with the CA file created as 0600 and owned by
ansible user.

Change it to become world readable, so that nova stops failing to
access it in molecule tests. 

Related: [OSPRH-13415](https://issues.redhat.com//browse/OSPRH-13415)
Closes: [OSPRH-14533](https://issues.redhat.com//browse/OSPRH-14533)

### DIFF
--- a/roles/edpm_libvirt/molecule/default/prepare.yml
+++ b/roles/edpm_libvirt/molecule/default/prepare.yml
@@ -41,7 +41,7 @@
         pushd repo-setup-main
         python3 -m venv ./venv
         PBR_VERSION=0.0.0 ./venv/bin/pip install ./
-        ./venv/bin/repo-setup current-podified -b antelope
+        ./venv/bin/repo-setup -d centos9 current-podified -b antelope
         popd
         rm -rf repo-setup-main
 
--- a/roles/edpm_nova/molecule/default/prepare.yml
+++ b/roles/edpm_nova/molecule/default/prepare.yml
@@ -144,6 +144,11 @@
         csr_content: "{{ csr.csr }}"
         provider: selfsigned
 
+    - name: Set permissions for certificate authority public key
+      ansible.builtin.file:
+        path: "{{ edpm_nova_tls_ca_src_dir }}/tls-ca-bundle.pem"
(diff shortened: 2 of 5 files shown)


slagle (approved):

stuggi:
recheck

bogdando:
recheck molecule

bogdando:
molecule failure is unrelated and should be intermittent, I could not reproduce locally

bogdando:
recheck

bogdando:
recheck edpm-ansible-molecule-edpm_nova

bogdando:
debugging with molecule CI only, to have a shorter testing loop

bogdando:
recheck autohold

zuul.d/projects.yaml
```diff
@@ -12,7 +12,6 @@
         - edpm-ansible-molecule-edpm_module_load
         - edpm-ansible-molecule-edpm_kernel
         - edpm-ansible-molecule-edpm_libvirt
-        - edpm-ansible-molecule-edpm_nova
```
SeanMooney, line 15: this is very high risk.
disabling the job should be avoided.

you can disable the failing test or make it non voting but this should effectively never be the approch.

bogdando: I agree, however this job is no more functional anyway, unless fixed.

But I updated this with a fix instead of dropping the test
https://github.com/openstack-k8s-operators/edpm-ansible/pull/896/commits/aa702f82e632551f175c85186a8ee2c851d86d09

PR #898, opened 2025-03-07
Use loop instead of with_items

bshephar:
We have largely tried to avoid using with_items in this repo in favor of loop. This change removes a couple of with_items that have snuck in.


slagle (approved):

bshephar:
recheck

PR #900, opened 2025-03-10
DNM debug molecule nova jobs only

bogdando:


bogdando:
confirmed the job fails on main

PR #902, opened 2025-03-11
Revert "Skip molecule test for unmodified roles"

SeanMooney:
Reverts openstack-k8s-operators/edpm-ansible#854

this is unsafe as it allows changes in one role to break another

Related Jira: [OSPRH-14533](https://issues.redhat.com//browse/OSPRH-14533)
Partial Jira: [OSPRH-14651](https://issues.redhat.com//browse/OSPRH-14651)

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -14,99 +14,71 @@
 - job:
     name: edpm-ansible-molecule-edpm_bootstrap
     parent: edpm-ansible-molecule-base
-    files:
-     - ^roles/edpm_bootstrap/(defaults|files|handlers|library|lookup_plugins|module_utils|molecule|tasks|templates|vars).*
     vars:
       TEST_RUN: edpm_bootstrap
 - job:
     name: edpm-ansible-molecule-edpm_podman
     parent: edpm-ansible-molecule-base
-    files:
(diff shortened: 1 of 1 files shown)


gibizer:
> Reverts #854
> 
> this is unsafe as it allows changes in one role to break another

Only if there are dependencies between said roles. I guess we are overeating here thinking that this change caused that the CI is now blocked with OSPRH-14651. I think that is not true.

bshephar:
recheck

bogdando:
> > this is unsafe as it allows changes in one role to break another
> 
> I don't agree with this. It's poor design if changes in one role breaks onother (roles are supposed to only do one thing and do it well). Molecule jobs are to test a role and not dependencies between roles. Other than some common roles i.e. edpm_container-manage, edpm_user etc, we should not be importing roles in another role. Playbooks are the place to use more than one role. Rather than testing role dependencies in molecule jobs it should be tested in e2e jobs.

As I commented https://github.com/openstack-k8s-operators/edpm-ansible/pull/902#issuecomment-2715036750 this is not only about roles to roles but containers/host packages to roles interference that needs to be caught up

SeanMooney:
lets close this for not.

we do have some explcit role deps but for now we are not acitvly workign on this so there is littel point in keeping this open.

we can reevaluate if we want ot trigger the nova role on libvirt changes ans vis versa separately form all roles and if they shoudl trigger on other dep changes as well.

PR #904, opened 2025-03-13
Don't mask old tripleo services in systemd during the cleanup

slawqo:
This patch removes masking old Tripleo services in systemd (makes them to be
symlinks to /dev/null). All those unit files are removed by the same
role so there should be no need to mask them before.

Closes-issue: #[OSPRH-11323](https://issues.redhat.com//browse/OSPRH-11323)

### DIFF
--- a/roles/edpm_tripleo_cleanup/tasks/main.yml
+++ b/roles/edpm_tripleo_cleanup/tasks/main.yml
@@ -73,26 +73,6 @@
     loop_var: path
   when: edpm_remove_tripleo_unit_files
 
-- name: Mask tripleo services to make them unusable
-  tags:
-    - adoption
-  become: true
-  ansible.builtin.systemd_service:
-    name: "{{ item }}"
-    masked: true
-  register: result_mask
-  loop: "{{ tripleo_services }}"
-  failed_when: false
-
-- name: Check for errors
-  ansible.builtin.fail:
-    msg: "Attempt to mask service {{ item.item }} failed with {{ item.msg }}"
-  loop: "{{ result_mask.results }}"
-  when: >
-    ('msg' in item) and
-    (item.msg != '') and
-    ("Could not find the requested service" not in item.msg)
-
 - name: Adopt (stop tracking) certs from tripleo
   tags:
     - adoption


rabi:
> This patch removes masking old Tripleo services in systemd (makes them to be
> symlinks to /dev/null). All those unit files are removed by the same
> role so there should be no need to mask them before.

@slawqo  We don't mask them before removing the unit files. but after. I'm not sure I get the current issue. AFAIK, removing the unit files alone will disable the service in terms of it being available for systemd to start, but it won't fully prevent systemd from trying to start it if there are any remaining references or triggers to that service. This is why masking the service is an important step if you want to make sure the service is completely inactive, even if someone or something attempts to start it.

slawqo:
> > This patch removes masking old Tripleo services in systemd (makes them to be
> > symlinks to /dev/null). All those unit files are removed by the same
> > role so there should be no need to mask them before.
> 
> @slawqo We don't mask them before removing the unit files. but after. I'm not sure I get the current issue. AFAIK, removing the unit files alone will disable the service in terms of it being available for systemd to start, but it won't fully prevent systemd from trying to start it if there are any remaining references or triggers to that service. This is why masking the service is an important step if you want to make sure the service is completely inactive, even if someone or something attempts to start it.

The issue is that after this role is run on the node, there are still tripleo services visible in the `systemctl list-unit-files --type service | grep tripleo` command on the hosts. Please see comments in https://issues.redhat.com/browse/OSPRH-11323 for more info. I understand that the issue is like that as after we mask those units, systemd creates their files in the /etc/systemd/system directory to be symlinks to /dev/null. So currently, if I am not mistaken the whole workflow of this role is like:
1. Remove Tripleo unit files from /etc/systemd/system and /usr/lib/systemd/system,
2. Mask those services which will create symlinks to /dev/null in  /etc/systemd/system,
3. Run cleanup which will try to reload units and run `systemctl reset-failed <unit_name>` for each tripleo unit to clean them

And this last step was failing (I don't remember exact error message now). What I have checked, doing "unmask" of the service manually made its disappear from the list of the services in tripleo.

So either we don't need to run this mask services at all as we are removing unit files and reloading systemctl's config always or maybe the validation of that Jira is wrong and we should expect that old services will be on that list. Or maybe we should do this cleanup in some other way which I don't know unfortunatelly :/ Please let me know what do you think about it.

slawqo:
> > 3. Run cleanup which will try to reload units and run `systemctl reset-failed <unit_name>` for each tripleo unit to clean them
> 
> Looks like we added call to `reset-failed` in [c5dedca](https://github.com/openstack-k8s-operators/edpm-ansible/commit/c5dedcadd60d41820357c57ab97877208a219247) and I think that command would probably always fail after we disable the service[1] and then remove the unit file (with or without the masking task). However that task would not fail as we've `failed_when: false` there[2].
> 
> The commit message is confusing as it says don't mask services before removing unit files and we're not doing that.
> 
> [1] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_tripleo_cleanup/tasks/main.yml#L51 [2] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_tripleo_cleanup/tasks/main.yml#L116
> 
> If we have issues with services listed but symlinked to /dev/null we can merge this after changing the commit message.

Commit message updated :)

rabi (commented):
/lgtm

slagle (approved):

slawqo:
recheck

slagle (approved):

PR #905, opened 2025-03-17
Preliminary bootc update handling

bshephar:
This PR adds the basic, preliminary handling for a bootc container upgrade. We simply perform a bootc switch to the image provided via the edpm_bootc_os_container_image variable. It's also necessary to skip other plays responsible for updating packages and kernels params.

Jira: https://issues.redhat.com/browse/OSPRH-14327

### DIFF
--- a/roles/edpm_update/tasks/bootc.yml
+++ b/roles/edpm_update/tasks/bootc.yml
@@ -15,26 +15,7 @@
 # under the License.
 
 
-# "edpm_update" will search for and load any operating system variable file
-
-- name: Apply kernel patch via kpatch
-  ansible.builtin.include_tasks: kpatch.yml
-  when:
-  - edpm_update_enable_kpatch
-  - not ansible_local.bootc
-
-- name: Update packages
-  ansible.builtin.include_tasks: packages.yml
-  when:
-  - edpm_update_enable_packages_update
-  - not ansible_local.bootc
-
-- name: Update OS (bootc)
-  ansible.builtin.include_tasks: bootc.yml
-  when: ansible_local.bootc
-
-- name: Update containers
-  ansible.builtin.include_tasks: containers.yml
-  when:
-  - edpm_update_enable_containers_update
-
+- name: Update bootc image
+  ansible.builtin.shell: bootc switch "{{ edpm_bootc_os_container_image }}"
+  become: true
+  register: bootc_result
--- a/roles/edpm_update/defaults/main.yml
+++ b/roles/edpm_update/defaults/main.yml
@@ -31,3 +31,7 @@ edpm_update_exclude_packages:
   - openvswitch
 
 edpm_update_running_services: "{{ edpm_service_types }}"
+
+# The bootc container image is provided by the OpenStackVersion
+# CR using the bootcOsContainerImage parameter
+edpm_bootc_os_container_image: ""


roles/edpm_update/tasks/bootc.yml
bshephar, line 1: We might need to drop a file here as well to indicate that we need to perform a reboot. That could be automatically picked up by the reboot role.

roles/edpm_update/tasks/bootc.yml
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+- name: Update bootc image
+  ansible.builtin.shell: bootc switch "{{ edpm_bootc_os_container_image }}"
```
steveb, line 19: How are service container images updated? Are they logically bound and pre-fetched in this image?

roles/edpm_update/tasks/main.yml
```diff
@@ -19,12 +19,21 @@
...
-  when: edpm_update_enable_packages_update
+  when:
+  - edpm_update_enable_packages_update
+  - not ansible_local.bootc
 
 - name: Update containers
   ansible.builtin.include_tasks: containers.yml
-  when: edpm_update_enable_containers_update
+  when:
+  - edpm_update_enable_containers_update
```
bshephar, line 35: @steveb There still updated here currently. Since we haven't landed on an implementation for quadlets yet, this change is just working within the existing framework.

I haven't submitted a PR for the quadlets work yet, but I'm trying to work on something that is consistent between RHEL and Image mode RHEL initially. Then we can build on that and assess the value of logically bound containers again. If we did want to pre-fetch containers as part of our image, we could still do that in the build process and it would work in the existing framework anyway.

roles/edpm_update/tasks/bootc.yml
bshephar: https://github.com/openstack-k8s-operators/edpm-ansible/pull/905#discussion_r2004802217

roles/edpm_update/tasks/main.yml
steveb: OK, sounds good to me

roles/edpm_update/tasks/main.yml
```diff
@@ -19,12 +19,21 @@
...
+  - not ansible_local.bootc
 
 - name: Update packages
   ansible.builtin.include_tasks: packages.yml
-  when: edpm_update_enable_packages_update
+  when:
+  - edpm_update_enable_packages_update
+  - not ansible_local.bootc
 
 - name: Update containers
```
jistr, line 32: We plan to split away the "Update packages" from "Update containers" into two distinct steps in the user minor update workflow, as users want more control and the ability to only replace containers without triggering OS-level updates:

https://issues.redhat.com/browse/RHOSSTRAT-23 
https://issues.redhat.com/browse/OSPRH-13841 

Based on your experience with bootc updates so far, do you think running the "Update containers" step in a different Deployment CR from "Update OS (bootc)" step would cause some kind of problem? It looks to me like we should be able to make these independent regardless if we use package mode or image mode.

roles/edpm_update/tasks/main.yml
```diff
@@ -19,12 +19,21 @@
...
   ansible.builtin.include_tasks: containers.yml
-  when: edpm_update_enable_containers_update
+  when:
+  - edpm_update_enable_containers_update
+
+- name: Update OS (bootc)
+  ansible.builtin.include_tasks: bootc.yml
+  when: ansible_local.bootc
```
jistr, line 39: Is there a particular reason this comes after the containers update? My instinct would be to update the system first (the kernel and other system-tied components can provide new interfaces/features that the new containers can make use of, not vice versa).

jistr: Actually i see the bootc update task file does not trigger a reboot so the order here doesn't matter.

roles/edpm_update/tasks/main.yml
bshephar: Yeah, I think it can be independent with both of them without issue. The thing to remember with image mode is that it's not _actually_ updating any packages, but rather downloading them and adding them to a different ostree layer. So nothing will be updated until the node has been rebooted. That's the major different between the two, so in that regard, it doesn't really matter too much when it happens, as long as it does happen at some point.

roles/edpm_update/tasks/main.yml
bshephar: Yeah, that's right. This step just writes that new ostree layer and then we wait until the user is ready to reboot.

jistr:
:+1: Thank you for the replies.

roles/edpm_container_quadlet/defaults/main.yml
slagle, line 1: isn't this role already merged on main?
so probably just need to rebase bootc on main again to pick it up?
also there is a default/main.yml and defaults/main.yml here.

roles/edpm_update/tasks/main.yml
slagle: It's not consistent with how package mode is done though. In that, there will be a period of time when new containers are running with old OS. If the user forgets to reboot or it's skipped, that could be an unsupported state.

We might need to do the bootc update first, force a reboot and verify it was done, then update containers.

bshephar: Move it now, but I'm reluctant to force a reboot.

roles/edpm_update/tasks/bootc.yml
```diff
@@ -0,0 +1,21 @@
+---
+# Copyright 2024 Red Hat, Inc.
```
ciecierski, line 2: year / 2025

slagle (approved):

PR #910, opened 2025-03-21
Run github workflows molecule when PR changes a role

eduolivares:
github workflows molecule tests have not been run since https://github.com/openstack-k8s-operators/edpm-ansible/pull/853 was merged.

### DIFF
--- a/.github/workflows/molecule.yaml
+++ b/.github/workflows/molecule.yaml
@@ -58,7 +58,8 @@ jobs:
     - name: Check whether the role is modified
       id: check_role
       run: |
-        if [[ $(git diff --name-only roles | grep "${{ matrix.tested_role }}/") ]]; then
+        git fetch origin $GITHUB_BASE_REF
+        if [[ $(git diff --name-only origin/$GITHUB_BASE_REF roles | grep "${{ matrix.tested_role }}/") ]]; then
           echo "modified=true" >> $GITHUB_OUTPUT
         else
           echo "modified=false" >> $GITHUB_OUTPUT


.github/workflows/molecule.yaml
```diff
@@ -58,7 +58,8 @@ jobs:
     - name: Check whether the role is modified
       id: check_role
       run: |
-        if [[ $(git diff --name-only roles | grep "${{ matrix.tested_role }}/") ]]; then
+        git fetch origin main
+        if [[ $(git diff --name-only origin/main roles | grep "${{ matrix.tested_role }}/") ]]; then
```
slagle, line 62: you probably want to use the target branch for the PR here, not main. Could probably use $GITHUB_BASE_REF

slagle:
> github workflows molecule tests have not been run since #853 was merged.

I see the molecule worfklows have run since then. WDYM?

eduolivares:
> > github workflows molecule tests have not been run since #853 was merged.
> 
> I see the molecule worfklows have run since then. WDYM?

@slagle, the jobs are run, but the tests are skipped. For example:
https://github.com/openstack-k8s-operators/edpm-ansible/actions/runs/13985991875/job/39159685691?pr=908#step:4:1

.github/workflows/molecule.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
     - name: Check whether the role is modified
       id: check_role
       run: |
-        if [[ $(git diff --name-only roles | grep "${{ matrix.tested_role }}/") ]]; then
+        if [[ $(git diff --name-only $GITHUB_BASE_REF roles | grep "${{ matrix.tested_role }}/") ]]; then
```
slagle, line 61: you may still need the git fetch, b/c the checkout action only checks out the PR commit(s)

slagle:
> > > github workflows molecule tests have not been run since #853 was merged.
> > 
> > 
> > I see the molecule worfklows have run since then. WDYM?
> 
> @slagle, the jobs are run, but the tests are skipped. For example: https://github.com/openstack-k8s-operators/edpm-ansible/actions/runs/13985991875/job/39159685691?pr=908#step:4:1

ok, that makes sense

.github/workflows/molecule.yaml
eduolivares: thank you. It seems to be working now.

eduolivares:
edpm_telemetry molecule is broken, but that's out of the scope of this PR. I'm removing the two dummy commits I used to trigger some molecule jobs and this PR will be ready for review.

slagle (approved):

atyronesmith:
> edpm_telemetry molecule is broken, but that's out of the scope of this PR. I'm removing the two dummy commits I used to trigger some molecule jobs and this PR will be ready for review.
> 
> https://github.com/openstack-k8s-operators/edpm-ansible/actions/runs/13990837210/job/39174203440?pr=910

The epdm_telemetry role is failing due the requirement for ovs to be installed for the container to run.  I need to modify the role such that the openstack_network_exporter is not run during molecule tests.  There is probably an argument to be made that all containers should be installed an run.  However, the container should fail if OVS is missing in real deployments.  Thoughts?

PR #911, opened 2025-03-21
Add docs for update role and playbook

slagle:
Signed-off-by: James Slagle <james.slagle@gmail.com>


bogdando (commented):
LGTM if we are not going to go deep into details

bshephar (approved):

PR #912, opened 2025-03-24
Add CA cert to bootstrap service

vakwetu:
Add CA cert to bootstrap service
    
    We want to copy the combined-ca-cert-bundle to the compute nodes to
    ensure that any third party certs are trusted.  Lets use the bootstrap
    service to do this.  This change is paired with - but does not depend on
    a change in openstack-operator that will ensure that the bootstrap service has the
    cert bundle mounted to /var/lib/openstack/cacerts/bootstrap.
    
    The relevant change is https://github.com/openstack-k8s-operators/openstack-operator/pull/1374
    
    Jira: OSPRH-14205

### DIFF
--- a/roles/edpm_bootstrap/tasks/bootstrap.yml
+++ b/roles/edpm_bootstrap/tasks/bootstrap.yml
@@ -110,6 +110,34 @@
 - name: Configure swap
   ansible.builtin.import_tasks: swap.yml
 
+- name: Check if CA cert file exists
+  delegate_to: localhost
+  ansible.builtin.stat:
+    path: /var/lib/openstack/cacerts/bootstrap/tls-ca-bundle.pem
+  register: ca_file
+
+- name: Copy CA cert file if present
+  when: ca_file.stat.exists
+  become: true
+  block:
+    - name: Ensure that the CA destination directory exists
+      ansible.builtin.file:
+        path: "/etc/pki/ca-trust/source/anchors"
+        state: "directory"
+        setype: "cert_t"
+        owner: "root"
+        group: "root"
+        mode: "0755"
+
+    - name: Copy CA certs to the standard location on the compute node
+      ansible.builtin.copy:
+        src: "/var/lib/openstack/cacerts/bootstrap/tls-ca-bundle.pem"
+        dest: "/etc/pki/ca-trust/source/anchors/tls-ca-bundle.pem"
+        mode: '0644'
+        owner: root
+        group: root
+      notify: "update-ca-trust"
+
 - name: FIPS tasks
   ansible.builtin.import_tasks: fips.yml
   when: edpm_bootstrap_fips_mode != 'check'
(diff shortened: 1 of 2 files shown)


roles/edpm_bootstrap/tasks/bootstrap.yml
```diff
@@ -110,6 +110,33 @@
...
+        state: "directory"
+        setype: "cert_t"
+        owner: "root"
+        group: "root"
+        mode: "0755"
+
+    - name: Copy CA certs to the standard location on the compute node
+      ansible.builtin.copy:
+        src: "/var/lib/openstack/cacerts/bootstrap/tls-ca-bundle.pem"
+        dest: "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
```
stuggi, line 135: from /etc/pki/ca-trust/extracted/pem/README

~~~
This directory /etc/pki/ca-trust/extracted/pem/ contains 
CA certificate bundle files which are automatically created
based on the information found in the
/usr/share/pki/ca-trust-source/ and /etc/pki/ca-trust/source/
directories.
...
Please never manually edit the files stored in this directory,
because your changes will be lost and the files automatically overwritten,
each time the update-ca-trust command gets executed.
~~~

/etc/pki/ca-trust/source/README
~~~
QUICK HELP: To add a certificate in the simple PEM or DER file formats to the
            list of CAs trusted on the system:

            Copy it to the
                    /etc/pki/ca-trust/source/anchors/
            subdirectory, and run the
                    update-ca-trust
            command.
~~~

Should we instead copy it to /etc/pki/ca-trust/source/anchors/ and call `update-ca-trust` on the edpm nodes?
Otherwise if someone out side the deploy process adds a cert to the above dir and calls the `update-ca-trust` would result in our copied bundle gets overridden

vakwetu: agreed. Fixed in new patch

roles/edpm_bootstrap/tasks/bootstrap.yml
```diff
@@ -110,6 +110,36 @@
...
+    - name: Copy CA certs to the standard location on the compute node
+      ansible.builtin.copy:
+        src: "/var/lib/openstack/cacerts/bootstrap/tls-ca-bundle.pem"
+        dest: "/etc/pki/ca-trust/source/anchors/tls-ca-bundle.pem"
+        mode: '0644'
+        owner: root
+        group: root
+
+    - name: Run update-ca-trust on the compute node
+      ansible.builtin.command: /usr/bin/update-ca-trust
```
slagle, line 141: can we check if the previous task actually changed the file? b/c otherwise, we don't need to run update-ca-trust every time.

vakwetu: ack - latest version adds a notify: which will invoke a handler to do the update-ca-trust on on changes.

PR #914, opened 2025-03-24
Quadlets

bshephar:


roles/edpm_logrotate_crond/defaults/main.yml
```diff
@@ -34,9 +34,18 @@ edpm_logrotate_crond_image: "quay.io/podified-antelope-centos9/openstack-cron:cu
 edpm_logrotate_crond_config_use_ansible: true
 edpm_logrotate_crond_config_dir: /var/lib/config-data/ansible-generated/crond
 edpm_logrotate_crond_volumes:
-  - /var/lib/kolla/config_files/logrotate_crond.json:/var/lib/kolla/config_files/config.json:ro
-  - "{{ edpm_logrotate_crond_config_dir ~ ':/var/lib/kolla/config_files/src:ro' }}"
-  - /var/log/containers:/var/log/containers:z
+  - path: /var/lib/kolla/config_files/logrotate_crond.json
```
bshephar, line 37: We need to render out these kolla files now that we're not relying on `edpm_container_manage`:

```
Mar 27 02:13:57 edpm-compute-1 edpm-compute@logrotate[36629]: Error: failed to create volume "logrotate-kolla-config": in parsing HostPath in YAML: faccessat /var/lib/kolla/config_files/logrotate_crond.json: no such file or directory
Mar 27 02:13:57 edpm-compute-1 systemd[1]: edpm-compute@logrotate.service: Main process exited, code=exited, status=125/n/a
```

roles/edpm_logrotate_crond/tasks/run.yml
```diff
@@ -18,12 +18,29 @@
...
+    state: absent
+  become: true
+
+- name: Render logrotate quadlet file
+  ansible.builtin.template:
+    src: templates/logrotate_crond.yaml.j2
+    dest: /etc/containers/systemd/logrotate.yaml
+  register: edpm_logrotate_quadlet_changed
+  become: true
+
```
bshephar, line 41: Render kolla files here before we try to start the service.

roles/edpm_iscsid/tasks/run.yml
```diff
@@ -52,7 +68,7 @@
     - not manage_iscsid_stat.changed|bool
     - iscsi_restart_stat.stat.exists|bool
   ansible.builtin.systemd:
-    name: edpm_iscsid
+    name: iscsid@edpm-compute.service
```
slagle, line 71: is this the service name or is it edpm-compute@iscsid like used at line 49?

roles/edpm_container_manage/tasks/main.yml
```diff
@@ -40,6 +40,7 @@
   when:
     - (all_containers_hash|length) > 0
     - edpm_container_manage_cli == 'podman'
+    - not edpm_use_quadlets
```
slagle, line 43: where is this defined?

roles/edpm_iscsid/tasks/run.yml
bshephar: Yeah, my bad. Corrected.

roles/edpm_container_manage/tasks/main.yml
bshephar: It's not defined anywhere atm. I was wondering if I should just refactor this role to write out quadlets instead of calling the custom Ansible modules that we have today. The main thing I would like to achieve when this is all done is remove as many of the custom container orchestration modules as possible:
https://github.com/openstack-k8s-operators/edpm-ansible/tree/main/plugins/modules

We're doing a couple of common things in each service which makes the refactoring of this `edpm_container_manage` role appealing. Like:

1. Disable the existing container systemd service;
2. Remove the systemd service file
3. Render the quadlet file
4. Render the kolla files
5. Render the associated configs
6. Start the service via the `edpm-compute@<service>` systemd unit.

The current format for calling this role is pretty perfectly suited to continue doing that, ie this format:
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_logrotate_crond/tasks/logrotate_crond.yml#L18-L26

Where we're able to pass in a couple of variables and have the standardized approach to doing all of the needed things.

It's probably worth the effort to do that, then I don't need this `edpm_use_quadlets` variable. It's more just there while I'm testing so that I could skip some targeted things and see what the outcome is. But I'll remove it before I take this PR out of draft.

slagle: I think a common role makes sense, but instead of refactoring edpm_container_standalone, I'd just write a new one: edpm_container_quadlet.

That way we can migrate services to the new role one at a time. And if we later find something that prevents migrating a service, it won't hold up everything else.

bshephar: Yeah, sounds good. I'll add a new role

bshephar: We can discuss that role here:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/922

PR #922, opened 2025-03-31
Add edpm_container_quadlet role

bshephar:
Introduce new edpm_container_quadlet role that will be used to handle all tasks related to starting / stopping and configuring quadlets. This role will be used to replace current usage of the
edpm_container_standalone role.

Jira: https://issues.redhat.com/browse/OSPRH-15285


roles/edpm_container_quadlet/defaults/main.yml
```diff
@@ -0,0 +1,55 @@
...
+    readOnly: true
+    name: ca-bundle
+  - path: /etc/pki/tls/certs/ca-bundle.trust.crt
+    mountPath: /etc/pki/tls/certs/ca-bundle.trust.crt
+    readOnly: true
+    name: ca-trust-bundle
+  - path: /etc/pki/tls/cert.pem
+    mountPath: /etc/pki/tls/cert.pem
+    readOnly: true
+    name: tls-cert
```
slagle, line 39: i assume that all of the above mounts under /etc/pki will always exist?

roles/edpm_container_quadlet/tasks/cleanup.yml
```diff
@@ -0,0 +1,13 @@
...
+    name: "{{ edpm_container_quadlet_service }}"
+    state: stopped
+    enabled: false
+  become: true
+
+- name: Cleanup standalone service file for {{ edpm_container_quadlet_service }}
+  ansible.builtin.file:
+    path: "/etc/systemd/system/{{ edpm_container_quadlet_service }}.service"
+    state: absent
+  become: true
```
slagle, line 13: should we do a `systemctl daemon-reload` here?

roles/edpm_container_quadlet/molecule/defaults/molecule.yml
```diff
@@ -0,0 +1,31 @@
+
```
slagle, line 1: this directly should be `molecule/default` instead of `molecule/defaults`?

roles/edpm_container_quadlet/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+  ansible.builtin.copy:
+    content: "{{ item.value | to_nice_json }}"
+    dest: "{{ edpm_container_quadlet_kolla_config_dir ~ '/' ~ item.key ~ '.json' }}"
+    mode: "0600"
+  loop: "{{ edpm_container_quadlet_kolla_config_files | dict2items }}"
+
+- name: Copy systemd service template file
+  ansible.builtin.copy:
+    src: "files/service-template.kube"
+    dest: /etc/containers/systemd/edpm-compute@.kube
```
slagle, line 39: I'd drop the `-compute` and just use `edpm@.kube`. just b/c there can be other edpm node types besides compute.

roles/edpm_container_quadlet/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+    mode: "0644"
+  # NOTE(tkajinam): Some containers can contain secrets in their environments.
+  #                 Hide the output to avoid dumping these to output.
+  no_log: true
+
+- name: Starting quadlet {{ edpm_container_quadlet_service }}
+  ansible.builtin.systemd_service:
+    name: edpm-compute@{{ edpm_container_quadlet_service }}
+    state: restarted
+    daemon_reload: true
```
slagle, line 57: `enabled: true`?

slagle (commented):
overall, looks good to me and I like the approach. Some inline comments.

roles/edpm_container_quadlet/tasks/cleanup.yml
bshephar: We _could_, but I don't think the `systemd_service` module will let me just do a `daemon-reload` without specifying some unit. I would probably have to use `shell:` to do that.

Not sure we really need to though tbh, because this will run first, then the configuration task with run for the quadlet and do a `daemon-reload` as part of that.

roles/edpm_container_quadlet/defaults/main.yml
bshephar: Pretty sure they do, I'll check the base bootc image to be sure.

bshephar: Yeah, these are always there:

```
❯ podman run -it registry.redhat.io/rhel9/rhel-bootc:9.4 ls /etc/pki/tls
cert.pem  certs  ct_log_list.cnf  fips_local.cnf  misc  openssl.cnf  openssl.d  private

❯ podman run -it registry.redhat.io/rhel9/rhel-bootc:9.4 ls /etc/pki/ca-trust
README  ca-legacy.conf  extracted  source

❯ podman run -it registry.redhat.io/rhel9/rhel-bootc:9.4 ls /etc/pki/ca-trust/source
README  anchors  blocklist  ca-bundle.legacy.crt
```

bshephar: CentOS Stream:
```
❯ podman run -it quay.io/centos-bootc/centos-bootc:stream9 ls /etc/pki
ca-trust  consumer  entitlement  entitlement-host  fwupd  fwupd-metadata  java  nssdb  rpm-gpg  tls
```

RHEL:
```
❯ podman run -it registry.redhat.io/rhel9/rhel-bootc:9.4 ls /etc/pki/tls
cert.pem  certs  ct_log_list.cnf  fips_local.cnf  misc  openssl.cnf  openssl.d  private
```

So, yeah, they both have that file available. We should be fine with the default.

bshephar: CentOS Stream:
```
❯ podman run -it quay.io/centos-bootc/centos-bootc:stream9 ls /etc/pki
ca-trust  consumer  entitlement  entitlement-host  fwupd  fwupd-metadata  java  nssdb  rpm-gpg  tls
```

RHEL:
```
❯ podman run -it registry.redhat.io/rhel9/rhel-bootc:9.4 ls /etc/pki/tls
cert.pem  certs  ct_log_list.cnf  fips_local.cnf  misc  openssl.cnf  openssl.d  private
```

So, yeah, they both have that file available. We should be fine with the default.

slagle (approved):

PR #923, opened 2025-03-31
Iscsid use edpm container quadlets

bshephar:
Iscsid use edpm_container_quadlets role

 This change moves the iscsid configuration from using
 edpm_container_standalone, to the new edpm_container_quadlets role.

Jira: https://issues.redhat.com/browse/OSPRH-15121
Signed-off-by: Brendan Shephard <bshephar@redhat.com>

### DIFF
--- a/roles/edpm_iscsid/tasks/run.yml
+++ b/roles/edpm_iscsid/tasks/run.yml
@@ -14,25 +14,27 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Ensure /usr/libexec/edpm-start-podman-container exists
-  ansible.builtin.import_role:
-    name: edpm_container_manage
-    tasks_from: shutdown.yml
-
 - name: Update iscsi health check script
   ansible.builtin.include_tasks:
     file: healthchecks.yml
 
-- name: Manage iscsid containers
-  ansible.builtin.include_role:
-    name: edpm_container_standalone
+- name: Cleanup existing standalone conatiner
+  ansible.builtin.import_role:
+    name: edpm_container_quadlet
+    tasks_from: cleanup.yml
   vars:
-    edpm_container_standalone_service: iscsid
-    edpm_container_standalone_container_defs:
-      iscsid: "{{ lookup('template', 'iscsid.yaml.j2') | from_yaml }}"
-    edpm_container_standalone_kolla_config_files:
+    edpm_container_quadlet_service: iscsid
+
+- name: Configure and start iscsid quadlet
+  ansible.builtin.import_role:
+    name: edpm_container_quadlet
+  vars:
+    edpm_container_quadlet_service: iscsid
+    edpm_container_quadlet_service_use_host_network: true
+    edpm_container_quadlet_service_volumes: "{{ edpm_iscsid_volumes }}"
+    edpm_container_quadlet_service_image: "{{ edpm_iscsid_image }}"
+    edpm_container_quadlet_kolla_config_files:
       iscsid: "{{ lookup('file', 'files/iscsid.yaml') | from_yaml }}"
-  register: manage_iscsid_stat
 
 - name: Check if the iscsid container restart is required
   ansible.builtin.stat:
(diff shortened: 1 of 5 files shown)


bshephar:
```
TASK [osp.edpm.edpm_container_quadlets : Starting quadlet iscsid name=edpm-compute@{{ edpm_container_quadlets_service }}, state=restarted] ***
task path: /usr/share/ansible/collections/ansible_collections/osp/edpm/roles/edpm_container_quadlets/tasks/main.yml:47
Monday 31 March 2025  06:15:09 +0000 (0:00:01.178)       0:00:33.086 **********
fatal: [compute-0]: FAILED! => {"changed": false, "msg": "Could not find the requested service edpm-compute@iscsid: host"}
```

roles/edpm_iscsid/tasks/run.yml
```diff
@@ -23,16 +23,22 @@
...
 
-- name: Manage iscsid containers
-  ansible.builtin.include_role:
-    name: edpm_container_standalone
+- name: Cleanup existing standalone conatiner
+  ansible.builtin.import_role:
+    name: edpm_container_quadlets
+    tasks_from: cleanup.yml
+  vars:
+    edpm_container_quadlet_service: iscsid
```
slagle, line 31: this var should be defined in the new role's defaults/main.yml, even with an empty value. And I suggest calling the new role just `edpm_container_quadlet` (without an 's').

roles/edpm_iscsid/templates/iscsid.yaml.j2
```diff
@@ -1,13 +1,37 @@
...
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  annotations:
+    bind-mount-options: /etc/iscsi:z
+  creationTimestamp: "2024-12-09T02:00:14Z"
+  labels:
+    app: iscsid
+  name: iscsid
```
slagle, line 10: would it make sense to move some of the above boilerplate to the edpm_container_quadlet role? Could be populated with the service name, etc.

bshephar: Yeah, that's a good idea. Probably a little more templating required for things like `hostNetwork` I guess. I'll have to check the other templates to see what other things are different. But should be easy enough to template the rest of it.

roles/edpm_iscsid/tasks/run.yml
bshephar: ack

bshephar:
Need to adjust the name of the container / healthcheck script that telemetry is searching for:
```
TASK [osp.edpm.edpm_telemetry : Wait until container is up and running name={{ container }}] ***
Friday 04 April 2025  04:57:50 +0000 (0:00:00.053)       0:02:08.152 **********
FAILED - RETRYING: [edpm-compute-0]: Wait until container is up and running (5 retries left).
FAILED - RETRYING: [edpm-compute-0]: Wait until container is up and running (4 retries left).
FAILED - RETRYING: [edpm-compute-0]: Wait until container is up and running (3 retries left).
FAILED - RETRYING: [edpm-compute-0]: Wait until container is up and running (2 retries left).
FAILED - RETRYING: [edpm-compute-0]: Wait until container is up and running (1 retries left).
fatal: [edpm-compute-0]: FAILED! => {"attempts": 5, "changed": false, "containers": [], "stderr": "Error: no such container iscsid\n", "stderr_lines": ["Error: no such container iscsid"]}
```

bshephar:
Ohh, actually, it's looking for a container called `iscsid` in the `podman ps` output, but it's not going to find that because of the pod naming convention. Hmm:

```
43be808260e8  quay.io/podified-antelope-centos9/openstack-iscsid@sha256:ff1d43bb9abfb8f64efb843f85dfef19dfe5112015c471395d8586f7326a64ac                      kolla_start  5 seconds ago  Up 5 seconds                       iscsid-iscsid
```

roles/edpm_iscsid/tasks/run.yml
slagle: edpm_container_quadlet/tasks/cleanup.yml needs to be changed to:

`
edpm_{{ edpm_container_quadlet_service }}
`

in order for it to find the servie to stop

slagle (commented):
see my comment in the run.yml. other than that, this lgtm. Once I manually stopped the old service, the quadlet based pod was able to start.

slagle:
we need to be sure and request review from the service owners for each service we migrate. In this case, @ASBishop and @fultonj at least

ASBishop:
> we need to be sure and request review from the service owners for each service we migrate. In this case, @ASBishop and @fultonj at least

Please see the comment I added to the jira.

PR #932, opened 2025-04-09
Consolidate "redhat" service conditions

cjeanner:
A node can be running RHEL, but doesn't necessarly need to
register/subscribe.

This PR ensures the node is running RHEL (as before), but also checks we
have at least `rhc_auth` parameter defined, and not empty.
This should provide enough security to re-add the "redhat" service in
the default list.

Related-Issue: [OSPRH-15644](https://issues.redhat.com//browse/OSPRH-15644)

### DIFF
--- a/playbooks/redhat.yml
+++ b/playbooks/redhat.yml
@@ -19,6 +19,8 @@
         name: "{{ lookup('ansible.builtin.env', 'EDPM_SYSTEMROLES', default='redhat.rhel_system_roles') + '.rhc' }}"
         apply:
           become: true
-      when: ansible_facts['distribution'] | lower == 'redhat'
+      when:
+        - ansible_facts['distribution'] | lower == 'redhat'
+        - rhc_auth | length > 0
       tags:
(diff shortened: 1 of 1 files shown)


playbooks/redhat.yml
```diff
@@ -19,6 +19,9 @@
         name: "{{ lookup('ansible.builtin.env', 'EDPM_SYSTEMROLES', default='redhat.rhel_system_roles') + '.rhc' }}"
         apply:
           become: true
-      when: ansible_facts['distribution'] | lower == 'redhat'
+      when:
+        - ansible_facts['distribution'] | lower == 'redhat'
+        - rhc_auth is defined
+        - rhc_auth | length > 0
       tags:
```
artificial-intelligence, line 26: technically you can write this as:
```suggestion
        - rhc_auth | length > 0
      tags:
```

as the `| length` filter will already check if the item is defined.

slagle (approved):

cjeanner:
recheck

PR #939, opened 2025-04-17
Force rhc_organization variable to be interger

bshephar:
In situations where we're unmarshaling data from json to form variables, it's possible that those values can be unmarshalled as floating point numbers. Which would be incompatible with the current org ID's. This change asserts that the type of number for rhc_organization should be an integer.

Jira: https://issues.redhat.com/browse/OSPRH-15097


bshephar:
recheck

slagle (approved):

PR #941, opened 2025-04-19
Fix undefined rhc_auth var

pojadhav:


playbooks/redhat.yml
```diff
@@ -21,6 +21,7 @@
           become: true
       when:
         - ansible_facts['distribution'] | lower == 'redhat'
+        - rhc_auth is defined
```
slagle, line 24: it was raised on the PR that added the next line that the check for defined isn't needed:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/932
is there an associated issue showing that it is?

pojadhav: I dont know but I was just trying to fix this because it is blocking cloudops component from promotion. 
https://issues.redhat.com/browse/OSPCIX-794

we need fix this issue.

slagle: ok, indeed. not sure why we concluded previously the "is defined" check wasn't needed. thanks!

slagle: note this will also need to be backported to 18.0-fr2 branch once this merges.

slagle (approved):

PR #946, opened 2025-04-24
Add a validation against edpm_network_config_template

cjeanner:
Depending on the tool used to configure network, the configuration
template is different.

This patch adds a quick validation to avoid obvious failures, by
ensuring a specific key is present in the template.

If the admin is considering "nmstate", "interfaces" key must be
present.

If the admin is considering "ifcfg", "network_config" key must be
present.

### DIFF
--- a/roles/edpm_network_config/tasks/main.yml
+++ b/roles/edpm_network_config/tasks/main.yml
@@ -23,6 +23,29 @@
     - edpm_network_config_tool == 'nmstate'
     - not edpm_network_config_tool_nmstate_override|bool
 
+- name: Validate edpm_network_config_template against configuration tool
+  vars:
+    _tmpl: "{{ edpm_network_config_template | from_yaml }}"
+  block:
+    - name: Assert nmstate compatible edpm_network_config_template
+      when:
+        - edpm_network_config_tool == 'nmstate' or edpm_network_config_nmstate | bool
+      ansible.builtin.assert:
+        that:
+          - _tmpl.interfaces is defined
+        msg: |
+          nmstate is enabled, but edpm_network_config_template doesn't seem correct.
+    - name: Assert ifcfg compatible edpm_network_config_template
+      when:
+        - edpm_network_config_tool != 'nmstate'
+        - not edpm_network_config_nmstate | bool
+      ansible.builtin.assert:
+        that:
+          - _tmpl.network_config is defined
+        msg: |
+          ifcfg is enabled, but edpm_network_config_nmstate doesn't have "network_config"
+          key.
+
 - name: Configure network with network role from system roles [nmstate]
   when: edpm_network_config_tool == 'nmstate'
   become: true


roles/edpm_network_config/tasks/main.yml
```diff
@@ -23,6 +23,27 @@
     - edpm_network_config_tool == 'nmstate'
     - not edpm_network_config_tool_nmstate_override|bool
 
+- name: Validate edpm_network_config_template against configuration tool
+  vars:
+    _tmpl: "{{ edpm_network_config_template | from_yaml }}"
+  block:
+    - name: Assert nmstate compatible edpm_network_config_template
+      when:
+        - edpm_network_config_tool == 'nmstate' or edpm_network_config_nmstate | bool
```
sathlan, line 32: Not really related to your patch, but why do we have *two* variable here instead of just one ?

roles/edpm_network_config/tasks/main.yml
```diff
@@ -23,6 +23,27 @@
...
+      ansible.builtin.assert:
+        that:
+          - _tmpl.interfaces is defined
+        msg: |
+          nmstate is enabled, but edpm_network_config_template doesn't seem correct.
```
sathlan, line 37: "is missing the interfaces configuration" seems more helpful.

roles/edpm_network_config/tasks/main.yml
```diff
@@ -23,6 +23,27 @@
...
+    - name: Assert ifcfg compatible edpm_network_config_template
+      when:
+        - edpm_network_config_tool != 'nmstate' and not edpm_network_config_nmstate | bool
```
sathlan, line 40: another `-` to the when instead clause might be easier to read ?

roles/edpm_network_config/tasks/main.yml
```diff
@@ -23,6 +23,27 @@
...
+      ansible.builtin.assert:
+        that:
+          - _tmpl.network_config is defined
+        msg: |
+          ifcfg is enabled, but edpm_network_config_nmstate doesn't seem correct.
```
sathlan, line 45: ifcfg is enabled, but edpm_network_config_nmstate "is missing network_config configuration." migth be more descriptive?

sathlan (approved):
Nothing seems wrong with the logic here and ci seems happy so far.  Ideally we could improve the error message.

roles/edpm_network_config/tasks/main.yml
cjeanner: because there are two of them apparently:
- one to use nmstate directly ("protected" by the first task of this file)
- one to instruct os-net-config to use nmstate as a backend instead of ifcfg

roles/edpm_network_config/tasks/main.yml
cjeanner: Depends on the final test content - for now it only checks that key, but there are some others we can check, for instance related to DNS

roles/edpm_network_config/tasks/main.yml
cjeanner: that one seems better - there's only one key in the template for this case, afaik.

Jaganathancse (changes requested):
@cjeanner 
I think, this validation is wrong for nmstate provider
nmstate provider and nmstate tool uses different config yaml fomat.
os-net-config  should have same config yaml format for both  ifcfg provider and nmsate provider .
and this config yaml should have 'network_config' only.
os-net-config needs 'network_config' and nmstate tool needs 'interfaces'.
@karthiksundaravel ^

cjeanner:
> @cjeanner I think, this validation is wrong for nmstate provider nmstate provider and nmstate tool uses different config yaml fomat. os-net-config should have same config yaml format for both ifcfg provider and nmsate provider . and this config yaml should have 'network_config' only. os-net-config needs 'network_config' and nmstate tool needs 'interfaces'. @karthiksundaravel ^

from my experience, this statement is wrong - I had failures using nmstate backend, while it was working fine with ifcfg. Either the os-net-config provider is doing something wrong with the template, or there's something else. I'm building an env so that you can investigate.

PR #948, opened 2025-04-28
Move OVN controller to quadlets

bshephar:
This change swaps ovn_controller from edpm_container_manage over to the new edpm_container_quadlet role. This aligns ovn_controller with our efforts to leverage native container tools provided by Podman and Systemd.

Jira: https://issues.redhat.com/browse/OSPRH-15124

PR #958, opened 2025-05-08
Cleanup old edpm and tripleo services

bshephar:
This change adds a block and a loop to ensure we're cleaning up both edpm and tripleo services. This change is necessary to handle cases where the user is migrating from TripleO to RHOSO.

### DIFF
--- a/roles/edpm_container_quadlet/tasks/cleanup.yml
+++ b/roles/edpm_container_quadlet/tasks/cleanup.yml
@@ -1,13 +1,19 @@
 ---
-- name: Stop and disable old standalone container for {{ edpm_container_quadlet_service }}
-  ansible.builtin.systemd_service:
-    name: "edpm_{{ edpm_container_quadlet_service }}"
-    state: stopped
-    enabled: false
-  become: true
 
-- name: Cleanup standalone service file for {{ edpm_container_quadlet_service }}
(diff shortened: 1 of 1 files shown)


roles/edpm_container_quadlet/tasks/cleanup.yml
```diff
@@ -1,13 +1,19 @@
...
 
-- name: Cleanup standalone service file for {{ edpm_container_quadlet_service }}
-  ansible.builtin.file:
-    path: "/etc/systemd/system/{{ edpm_container_quadlet_service }}.service"
-    state: absent
-  become: true
+- name: Cleanup old edpm and tripleo services
+  loop:
+    - "edpm"
+    - "tripleo"
```
slagle, line 6: if this is for adoption, aren't the tripleo services handled already by edpm_tripleo_cleanup/tasks/main.yml?

bshephar: Yeah, it's an attempt to resolve the issue I have with the OVN quadlets PR:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/948#issuecomment-2861476157

But I agree, it would be better to keep everything in one place. It will become cumbersome if we need to handle it separately.

PR #959, opened 2025-05-08
DNS config is not updated for nmstate provider

Jaganathancse:
This change is to update the DNS config for nmstate provider and making default changes to update
'/etc/resolv.conf' for DNS nameserver.


roles/edpm_bootstrap/defaults/main.yml
```diff
@@ -64,7 +64,7 @@ edpm_bootstrap_network_service: NetworkManager
 # Control resolv.conf management by NetworkManager
 # false = disable NetworkManager resolv.conf update (default)
 # true = enable NetworkManager resolv.conf update
-edpm_bootstrap_network_resolvconf_update: false
+edpm_bootstrap_network_resolvconf_update: "{{ edpm_network_config_nmstate | default(False) }}"
```
karelyatin, line 67: This will need edpm_network_config_nmstate to be set in nodeset in order to override it. But you also trying to update role default in https://github.com/openstack-k8s-operators/edpm-ansible/pull/908/files and role default change will not work unless nodeset have overrides

Jaganathancse: updating default(True) then will work for all scenarios.

Jaganathancse:
edpm_bootstrap_network_resolvconf_update: "{{ edpm_network_config_nmstate | default(True) }}"i
edpm_network_config_nmstate is True then edpm_bootstrap_network_resolvconf_update=True
edpm_network_config_nmstate=False then edpm_bootstrap_network_resolvconf_update=False
edpm_network_config_nmstate is not defined then edpm_bootstrap_network_resolvconf_update=True and also default edpm_network_config_nmstate will be True.
For already deployed cluster edpm_network_config_nmstate is False then edpm_bootstrap_network_resolvconf_update=True based on default but we have edpm_bootstrap_network_resolvconf_update variable to override it. manually add in nodeset CR edpm_bootstrap_network_resolvconf_update=False.

Jaganathancse:
otherwise we need to code changes in either bootstrap or network_config role.

roles/edpm_bootstrap/defaults/main.yml
karelyatin: No it will not as then will require explicit edpm_network_config_nmstate=false or edpm_bootstrap_network_resolvconf_update=false in already deployed cluster nodeset with ifcfg provider.

Also will break the CI as current CI running with ifcfg provider.

Jaganathancse: yes we need to add in nodeset CR to fix issue, we cannot add default value for all scenarios, this default change handles for nmstate scenarios, for ifcfg we need to update nodeset CR for update scenarios.

Jaganathancse:
Moved this changes part of another PR: https://github.com/openstack-k8s-operators/edpm-ansible/pull/908

PR #964, opened 2025-05-26
Create tls config in ceilometer_ipmi for ceilometer_exporter

mgirgisf:


mgirgisf:
recheck

roles/edpm_telemetry_power_monitoring/tasks/configure.yml
```diff
@@ -113,6 +113,14 @@
...
     - tls_crt_stat.stat.exists and tls_key_stat.stat.exists
   block:
+    - name: Create config file for exporters
+      ansible.builtin.include_tasks:
+        file: exporter_tls.yml
+      loop:
+        - ceilometer_prom_exporter
+      loop_control:
+        loop_var: exporter
+
```
slagle, line 123: why a loop with 1 item?

mgirgisf: I'm adding that to follow the same technique in the ceilometer_compute config for any future exporters.

slagle (commented):
+2

PR #965, opened 2025-05-27
Replace tripleo-repos with repo-setup

raukadah:
In Podified land, we use https://github.com/openstack-k8s-operators/repo-setup to generate repos.

By using tripleo-repos for edpm-ansible molecule job testing causes dependency issue.

This pr
- Updates tripleo-repos instructions with repo-setup
- Fix ubi repos names
- replaces current-tripleo with current-podified

### DIFF
--- a/molecule/common/Containerfile.j2
+++ b/molecule/common/Containerfile.j2
@@ -5,10 +5,13 @@ FROM {{ item.registry.url }}/{{ item.image }}
 FROM {{ item.image }}
 {% endif %}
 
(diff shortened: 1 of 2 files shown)


slagle (commented):
LGTM, but I think we need to have the molecule workflow triggering on changes to these files. I've proposed a PR to do that: https://github.com/openstack-k8s-operators/edpm-ansible/pull/970
Once that merges, we can retest this one.

slagle (approved):

PR #971, opened 2025-05-30
Follow-ups from split update review

jistr:
This implements three things requested in reviews of #962:

- Simpler and faster (fewer tasks executed) way of dealing with package exclusions.

- Reformulation of 'Apply package updates needed for service updates' into 'Apply essential package updates required by services'

- Making sure we have the latest kpatch package

Related: https://issues.redhat.com/browse/OSPRH-16546


slagle (commented):
+2, dfg-upgrades should review @holser @ciecierski

PR #978, opened 2025-06-04
Fixes to test_deps role to avoid repos issues

raukadah:
Below are the fixes:

- molecule jobs runs via github-actions and mirror info.sh file does not exists there. Due to that we cannot disable ubi repos. Make it under or conditional fixes the ubi repo removal.

- centos-stream-release package installs centos.repo file. We rely on repo-setup to lay down the repos. Let's do not install centos-stream-release repo. It is not needed as centos repos are lay down by repo-setup tool.

  It also avoid package conflicts issues.


Jaganathancse (approved):

slagle (approved):

rabi:
@raukadah I think this has broken[1] most of the molecule jobs that install dependencies. We now remove ubi9.repo and depend on repo-setup, but those tasks only run when `test_deps_setup_edpm: true` and not every molecule role sets it as true. Proposed https://github.com/openstack-k8s-operators/edpm-ansible/pull/996.


[1] https://github.com/openstack-k8s-operators/edpm-ansible/actions/runs/16667833303/job/47177528460?pr=993

PR #968, opened 2025-05-28
Improvements to download retries in roles

sdatko:
This pull request contains two changes proposed:

**Add variable to override all download retries**

> Currently each edpm_* role defines their own set of variables
> to control the number of retries and delay between retries
> when attempting to download and install the system packages.
> 
> There is over 20 such roles that involve it already, however
> there is no way to set those variables easily for all roles
> at once – the only option for someone is to set all about
> 40 variables in worst scenario.
> 
> This commit introduces two variables that would be used
> as first choice, if provided, by all the roles to allow
> the setting the download retries much easier.

**Increase default delay between download retries**

> From my experience on our testing environments,
> e.g. when there is problem accessing REPOMD.xml file,
> the current default default of 5 seconds is not realistic.
> If repo is temporarily unavailable, it usually takes
> a few minutes before it is back online.
> 
> Hence, let's use at least a minute between retries.

PR #981, opened 2025-06-13
[libvirt] Give more time to install

jirimacku:
Wait longer and repeat more times before giving up the installation. It's better to wait 5 min than loose 4 hours of previous deployment. We have seen couple of times failing this because of temporary network glitch or repo sync.


rabi (commented):
I think there is already a PR to improve it across services https://github.com/openstack-k8s-operators/edpm-ansible/pull/968/

jirimacku:
Ahh, I missed the #968. Closing this one.

PR #983, opened 2025-06-20
Fix permissions for Swift EDPM role

cschwede:
Depending on the services deployed before the Swift role, the directory /var/lib/edpm-config/firewall might not exist and/or permissions are (correctly) limited to root, failing later when adding firewall rules.

JIRA: [OSPRH-17624](https://issues.redhat.com//browse/OSPRH-17624)

### DIFF
--- a/roles/edpm_swift/tasks/firewall.yml
+++ b/roles/edpm_swift/tasks/firewall.yml
@@ -14,7 +14,17 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Ensure firewall directory is present
+  become: true
+  ansible.builtin.file:
+    path: "/var/lib/edpm-config/firewall/"
+    state: directory
+    owner: root
+    group: root
+    mode: '0750'
+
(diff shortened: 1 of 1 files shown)


cschwede:
Looks like the SF error is unrelated to this change:

```
2025-06-20 10:50:12.453234 | crc | tar: Removing leading `/' from member names
2025-06-20 10:50:12.453284 | crc | tar: /var/home/core/ci-framework-data: Cannot stat: No such file or directory
2025-06-20 10:50:12.456625 | crc | tar: Exiting with failure status due to previous errors
2025-06-20 10:50:12.459805 | crc | mv: cannot move '/var/home/core/ci-framework-data-crc-all-logs.tar.gz' to '/var/home/core/ci-framework-data/logs/ci-framework-data-crc-all-logs.tar.gz': No such file or directory
2025-06-20 10:50:12.462161 | crc | chown: cannot access '/var/home/core/ci-framework-data/logs/ci-framework-data-crc-all-logs.tar.gz': No such file or directory
```

PR #989, opened 2025-06-30
Allow unprivileged rsync use port 873 for Swift

cschwede:
rsync uses port 873 to replicate data between Swift storage nodes. However, rsync is running unprivileged and thus not able by default to use port 873. This can be done by using the bootstrap role and setting edpm_kernel_sysctl_extra_settings accordingly.

Because this is always needed for Swift, this patch sets the required sysctl when configuring Swift, making it easier for human operators to deploy correctly.

### DIFF
--- a/roles/edpm_swift/tasks/configure.yml
+++ b/roles/edpm_swift/tasks/configure.yml
@@ -25,6 +25,15 @@
   tags:
     - edpm_users
 
+- name: Allow unprivileged rsync use port 873
+  ansible.posix.sysctl:
+    name: net.ipv4.ip_unprivileged_port_start
+    value: '873'
+    sysctl_set: true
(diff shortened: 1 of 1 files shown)


roles/edpm_swift/tasks/configure.yml
```diff
@@ -25,6 +25,15 @@
   tags:
     - edpm_users
 
+- name: Allow unprivileged rsync use port 873
+  ansible.posix.sysctl:
+    name: net.ipv4.ip_unprivileged_port_start
```
rabi, line 30: Hmm.. AFAIR this not a security best practice as it will allow user processes to bind to any ports >=873. Can we use capabilities instead or run rsync with escalated previleges in archive mode(-a)?

PR #995, opened 2025-07-30
Drop edpm_logrotate_crond role and it's usage

rabi:
We now write all service logs to journal and this is not required.


roles/edpm_update_services/tasks/containers.yml
```diff
@@ -128,17 +128,6 @@
...
-  ansible.builtin.include_role:
-    name: osp.edpm.edpm_logrotate_crond
-    tasks_from: run.yml
-    apply:
-      become: true
-  tags:
-    - edpm_logrotate_crond
-    - edpm_update_services
-  when: '"run-os" in edpm_update_services_running_services'
-
```
slagle, line 141: I think we'll want to add something here to stop, delete, cleanup the existing container.
Perhaps put all of the migration related tasks for FR4 into their own tasks file which can be included from here.

rabi: yes, I have thought about it. I don't see anything to cleanup orphaned containers and services that are deployed and then removed unlike `TripleO` where we used to delete all orphaned containers/systemd units (not in config). That's probably something generic to update workflow and could be handled in a separate PR(?). We've a `tripleo_cleanup` role which cleans up tripleo services duuring adption and probably could be extended to do this as well.

rabi: My point is that we need something generic which would cleanup orphaned services/containers like before. As we don't have a single config to know all what's deployed on an edpm node, probably we need to find a way to discover what's deployed rather than knowing beforehand that update from x to y version need to cleanup `these` services, though it may work if we know that users won't add/remove services other than doing an minor update.

slagle: Is there a Jira for the work to cleanup removed services? I think we need something so it isn't forgotten. I think this PR should probably depend on the cleanup work being in place, but we at least need a task for it in the backlog for FR4.

In the operator, we could get a record of the deployed services by looking at the history of Deployments done (which we already iterate over). We could then past that history down to the ansible for it to act on do the right cleanup. Or, we could store some state on the EDPM nodes themselves.

I wasn't a huge fan of how it was done in TripleO, since it seemed like that approach could cleanup anything not deployed by TripleO ansible, which might inadvertently clean up 3rd party or custom deployed containers.

rabi: >  I think this PR should probably depend on the cleanup work being in place

I thought we should have taken care of that as part of update/upgrade work.  As per this PR dependancy on cleanup work, I don't think it's necessary as `logrotate_crond` container is running without doing anything and reports unhealthy and we should remove it from fresh deployments. I'll create a jira to track the service cleanup work.

slagle (approved):

PR #997, opened 2025-08-01
[WIP] reduce libvirtd log levels

SeanMooney:
This change reduces the libvirtd log levels by default
from info to warning level. It also makes both the
log_output and log_filter configurable.

TODO:
- update the role parmater docs
- update molecule testing

Signed-off-by: Sean Mooney <work@seanmooney.info>


bogdando:
looks correct, but mind explaining the need of this change (syslog ratelimiting?)

SeanMooney:
@bogdando the need for this change is partly because https://issues.redhat.com/browse/RHOSRFE-148 is not implemented.

tl;dr at info level livbirt and ti deamon log alot.
normally that would not be a problem however edpm-ansible is not currently configuring journald to do log rotation.

thats a problem that need to be address via a new role.

curreently libvirt and ceiolometer i belvie are teh two most verbose services on teh edpm node.

given we rarely look at the libvirt logs wehn debuging issues reducign the log leve back ot what we had in tripleo seams reasonable.

i intentionally increased them to info level for the dev preview released because it seam to be an acceptable trade off in the level of log output. i still think it likely is, but this change is just helping reduce the problem until someone can work on the real fix.

we may decied to go back to info level in the future but in the 7 years i have been at redhat i can count the time i have needed the libvirt deamon logs to debug an issue on one hand.

we almost always have the error we care about in the nova log or the qemu instnce log.

bogdando:
LGTM

roles/edpm_libvirt/molecule/default/prepare.yml
```diff
@@ -67,6 +67,17 @@
...
+    - name: remove vagrant image packages
+      become: true
+      ansible.builtin.dnf:
+        name: "{{ item }}"
+        state: absent
+      loop:
+        # NOTE(sean-k-mooney): without this network connectivity will be lost
+        # when the nft rules are applied as firewalld conflicts with the rules
+        # we set.
+        - firewalld
```
SeanMooney, line 79: this is nice to have but ..

gibizer (approved):
still looks good

SeanMooney:
check-rdo

PR #998, opened 2025-08-06
Add retries while Verify a FQDN is set

amartyasinha:
Due to some race condition, FQDN check is running before dnsmasq is returning FQDN for compute nodes. To avoid that, we want that this check should retry for couple of minutes until dnsmasq is ready and returns FQDN.

Ref: https://github.com/openstack-k8s-operators/architecture/pull/595

### DIFF
--- a/roles/edpm_nodes_validation/tasks/main.yml
+++ b/roles/edpm_nodes_validation/tasks/main.yml
@@ -72,12 +72,13 @@
       ansible.builtin.command: hostname -f
       register: hostname
       changed_when: false
-
-    - name: Verify a FQDN is set
-      ansible.builtin.assert:
-        that: hostname.stdout.find(".") != -1
-        fail_msg: "{{ hostname.stdout }} does not contain . and does not appear to be an FQDN."
-        success_msg: "{{ hostname.stdout }} contains . and appears to be an FQDN."
+      retries: 5
+      delay: 60
+      until: hostname.stdout.find(".") != -1
+  rescue:
+    - name: Fail if hostname is not FQDN
+      ansible.builtin.fail:
+        msg: "{{ hostname.stdout }} does not contain . and does not appear to be an FQDN."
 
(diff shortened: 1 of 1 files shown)


roles/edpm_nodes_validation/tasks/main.yml
```diff
@@ -72,12 +72,13 @@
...
-      ansible.builtin.assert:
-        that: hostname.stdout.find(".") != -1
-        fail_msg: "{{ hostname.stdout }} does not contain . and does not appear to be an FQDN."
-        success_msg: "{{ hostname.stdout }} contains . and appears to be an FQDN."
+      retries: 5
+      delay: 60
+      until: hostname.stdout.find(".") != -1
+  rescue:
+    - name: Print the error
+      ansible.builtin.debug:
```
sdatko, line 80: It should be `.fail` I guess?

Now it will succeed and continue the normal execution... (Which I wonder what would lead to ~ why we need that verification of FQDN?)

roles/edpm_nodes_validation/tasks/main.yml
```diff
@@ -72,12 +72,13 @@
...
-        that: hostname.stdout.find(".") != -1
-        fail_msg: "{{ hostname.stdout }} does not contain . and does not appear to be an FQDN."
-        success_msg: "{{ hostname.stdout }} contains . and appears to be an FQDN."
+      retries: 5
+      delay: 60
+      until: hostname.stdout.find(".") != -1
+  rescue:
+    - name: Fail if hostname is not FQDN
+      ansible.builtin.fail:
+        msg: "{{ hostname.stdout }} does not contain . and does not appear to be an FQDN."
```
slagle, line 81: I don't think we should add delays/retries when we aren't sure what we're delaying for. Do we have enough debug information to tell what the missing configuration actually was? If not, perhaps we can add some debug tasks to show that (e.g., can't actually reach DNS, etc).

PR #1007, opened 2025-08-20
PAckages installation failing changes

Jaganathancse:
This changes to fix the package installation issue due to DNS NM config changes part.

jira: https://issues.redhat.com/browse/OSPRH-19018


Jaganathancse:
recheck

roles/edpm_network_config/tasks/main.yml
```diff
@@ -23,23 +23,8 @@
...
-    - name: Restart NetworkManager after plugin installation [nmstate]
-      ansible.builtin.systemd:
-        name: NetworkManager
-        state: restarted
-      when: nm_ovs_status.changed  # noqa: no-handler
+- name: Install required packages
+  ansible.builtin.import_tasks: install.yml
 
 - name: Import DNS NetworkManager configs tasks
   ansible.builtin.import_tasks: dns_nm_configs.yml
```
rabi, line 30: This would still not work after minor updates when edpm_network_config_nmstate default is changed to true (earlier it would be false) which would remove dns=none and you want to re-run `configure_network` service (probably part of scale-out) 

This is what will happen

- Reloading NetworkManager would wipe out /etc/resolv.conf as cloud-init would have configured network with sysconfig renderer and existing nodes will permanently lose entries in resolv.conf.
- There would be no os-net-config run on existing nodes as the os-net-config has run once to reconfigure resolv.conf.

I would suggest we add a check to not touch NetworkManager.conf for existing nodes where network has already been configured.

Please test all scenarios including minor updates.

Jaganathancse: @rabi only Network Manager reload causing this issue for existing nodes where network already configured.
Only Network Manager reload triggered when Network Manager conf  changes done first time either removed or set dns line.

Jaganathancse: @rabi before default was nmstate provider mean, ifcfg to nmstate provider change is not supported in minor update scenario, only same provider to be used in minor update either ifcfg or nmstate.
We are working on migration (ifcfg to nmstate provider) and in progress.

Jaganathancse: > @rabi before default was nmstate provider mean, ifcfg to nmstate provider change is not supported in minor update scenario, only same provider to be used in minor update either ifcfg or nmstate. We are working on migration (ifcfg to nmstate provider) and in progress.

@karthiksundaravel @vcandapp ^

rabi: How do you restrict it when you say unsupported? Earlier default was ifcfg and new default is nmstate. Without any change in CRs, update to FR3 will change the default and it would break things.

Jaganathancse:
Duplication PR: https://github.com/openstack-k8s-operators/edpm-ansible/pull/1012/files
so closing my PR for same issue.

PR #1010, opened 2025-08-25
Update kepler container template

rsafrono:
Update kepler image default version to v0.10.2.
    
The new kepler (starting from 0.10) has completely different set of
config options so kepler's template was updated correspondingly.
    
- Removed '-v' option as it's no longer supported.
- Added option --config.file to specify a file with settings
- There is no longer need to specify desired metrics using
  environment variables as all metrics are enabled by default.
    
Also changed kepler healthcheck script:
- 'healthz' URL path no longer exist, so use 'metrics' instead.
- Fetch only headers and status code for healthcheck in order to
  minimize load.
- Removed check for 'ok' in response as it is no longer relevant
  after 'healthz' endpoint removal from kepler.

### DIFF
--- a/roles/edpm_telemetry_power_monitoring/defaults/main.yml
+++ b/roles/edpm_telemetry_power_monitoring/defaults/main.yml
@@ -38,7 +38,7 @@ edpm_telemetry_image_download_retries: "{{ edpm_download_retries | default(5) }}
 edpm_telemetry_old_tripleo_compute_sevices:
   - tripleo_ceilometer_agent_ipmi.service
 # Image to use for kepler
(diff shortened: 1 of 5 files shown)


rsafrono:
recheck

rsafrono:
recheck molecule

karelyatin:
As commented in slack need to adapt this for CI with https://github.com/sustainable-computing-io/kepler/blob/905238acb52da8f40a24ae264a5c096231c3ff4d/docs/developer/design/architecture/configuration.md?plain=1#L275-L277

rsafrono:
recheck molecule

rsafrono:
recheck molecule

rsafrono:
recheck

PR #1013, opened 2025-09-02
Remove all /var/log/containers/* mounts

rabi:
All services log to journald and /var/log/containers is not used at all. Remove it from roles.

### DIFF
--- a/plugins/modules/edpm_container_manage.py
+++ b/plugins/modules/edpm_container_manage.py
@@ -73,6 +73,7 @@ options:
   log_base_path:
     description:
       - Log base path directory
+      - DEPRECATED - Services now log to journald, this parameter is ignored
     type: str
     default: '/var/log/containers/stdouts'
   concurrency:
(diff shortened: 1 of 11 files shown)


rabi:
recheck

plugins/modules/edpm_container_manage.py
```diff
@@ -70,11 +70,6 @@
     default: {}
     required: False
     type: dict
-  log_base_path:
-    description:
-      - Log base path directory
-    type: str
-    default: '/var/log/containers/stdouts'
```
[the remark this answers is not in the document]
rabi: OK. I thought third-party playbooks/roles are expected to use edpm_container_standalone role rather than edpm_container_manage role/module. Ideally we should have deprecated it with https://github.com/openstack-k8s-operators/edpm-ansible/pull/547 when switched to journald directly or could have made the driver configurable. 

We don't yet have any collection versioning (we use 0.0.1) to provide deprecation details, but I'll assume that next version would be 0.0.2.

rabi:
recheck

plugins/modules/edpm_container_manage.py
```diff
@@ -288,7 +298,6 @@ def manage_container(self, name, config):
             'debug': self.debug,
             'log_driver': 'journald',
             'log_level': 'info',
-            'log_opt': {"path": f"{self.log_base_path}/{name}.log"},
```
slagle, line 291: by dropping this, the functionality is removed, not just deprecated. Could we leave this in place until the next release?

rabi: Not really. `log_opt/path` is not relevant when using journald log driver[1]. As I mentioned in the earlier comment, the functionality has been removed since we switched the `log_driver`.

[1] https://github.com/containers/podman/blob/main/pkg/specgen/specgen.go#L22-L25

slagle: No, this lgtm, thanks.

slagle (approved):

PR #1015, opened 2025-09-09
Use separate registries.conf.d files for config

slagle:
Depending on the ansible vars used, there are different ways to
configure registries with edpm_podman. Instead of one way overwriting
the other by both using the same /etc/containers/registries.conf file,
use separate files under /etc/containers/registries.conf.d.

Jira: OSPRH-19793


slagle:
recheck

slagle:
the rdo jobs that passed look like they did the right thing, sosreport-compute-0-2025-09-09-smzidky/etc/containers/registries.conf.d/20-edpm-podman-registries.conf exists with the expected content.

bshephar (approved):
lgtm, this is a better design pattern. But how does this save you from the two different methods? Are you going to also explicitly set `edpm_podman_registries_conf_file` in:
https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/dataplane/inventory.go#L151-L153

Or something?

bshephar:
> @bshephar: changing LGTM is restricted to collaborators

Rude, I'm clearly collaborating.

slagle:
> lgtm, this is a better design pattern. But how does this save you from the two different methods? Are you going to also explicitly set `edpm_podman_registries_conf_file` in: https://github.com/openstack-k8s-operators/openstack-operator/blob/main/pkg/dataplane/inventory.go#L151-L153

The defaults for the conf_file vars I added here would take effect when those vars are set. The disconnected registries config would then go to a different file than the file for any user config.

The user could still undo or configure something conflicting (intentionally or not) , but this way it at least gives more control over what config overrides the other. I think this is typical for the standard ".d" dir pattern. If the user wants to configure things further, it's up to them to put additional files there and ensure they're indexed in the right order, taking into consideration what was laid down by packaging or something else.

Hello, btw.

PR #1017, opened 2025-09-11
Fix rsync permission issue for Swift

cschwede:
rsync uses port 873 to replicate data between Swift storage nodes. Because it is running unprivileged it needs the NET_BIND_SERVICE cap to make this work by default without any customization.


rabi (approved):
/lgtm

PR #1018, opened 2025-09-11
Merge bootc branch into main

slagle:
Merges the bootc branch into main.
Jira: [OSPRH-15746](https://issues.redhat.com//browse/OSPRH-15746)

- **Gather ansible_local facts at the start of each service**
- **Re-enable swap for bootc**
- **Skip download_cache packages tasks with bootc**
- **Add bootc fact tasks**
- **Trim newline from end of swap path**
- **Gather local facts for bootc**
- **Fix ansible-lint complaints**
- **bootc fact bashate spacing**
- **Do not change FIPS status for bootc and warn if requested**
- **Preliminary bootc update handling**
- **edpm_ovn_bgp_agent: use /var paths instead of /usr for bootc**
- **Gather local facts for neutron_dhcp playbook**
- **Gather local facts for configure_ovs_dpdk playbook**
- **Gather local facts for update.yml playbook**
- **Gather local facts for playbooks/telemetry_logging.yml**
- **edpm_kernel bootc support**
- **edpm_kernel: bootc support**
- **edpm_kernel: bootc molecule tests**
- **select_kernel_ddp_package.yml bootc support**
- **Fail when edpm_ddp_package used on a bootc system**
- **Gather local ansible facts if not done for roles**
- **Add prepare.yml for edpm_sshd molecule tests**

### DIFF
--- a/playbooks/bootstrap.yml
+++ b/playbooks/bootstrap.yml
@@ -17,7 +17,10 @@
   tasks:
     - name: Gather ansible_local facts
       ansible.builtin.setup:
-        filter: ansible_local
+        gather_subset:
+          - "!all"
+          - "!min"
+          - "local"
     - name: Grow volumes
       ansible.builtin.import_role:
         name: osp.edpm.edpm_growvols
--- a/roles/edpm_ddp_package/tasks/main.yml
+++ b/roles/edpm_ddp_package/tasks/main.yml
@@ -16,7 +16,10 @@
 
 - name: Gather ansible_local facts
   ansible.builtin.setup:
-    filter: ansible_local
+    gather_subset:
+      - "!all"
+      - "!min"
+      - "local"
 
 - name: Fail if bootc
   ansible.builtin.debug:
--- a/roles/edpm_kernel/tasks/kernelargs.yml
+++ b/roles/edpm_kernel/tasks/kernelargs.yml
@@ -16,7 +16,10 @@
 
 - name: Gather ansible_local facts
   ansible.builtin.setup:
-    filter: ansible_local
+    gather_subset:
+      - "!all"
+      - "!min"
+      - "local"
   when:
     - ansible_local is not defined
 
--- a/playbooks/configure_network.yml
+++ b/playbooks/configure_network.yml
@@ -9,7 +9,10 @@
   tasks:
     - name: Gather ansible_local facts
       ansible.builtin.setup:
-        filter: ansible_local
+        gather_subset:
+          - "!all"
+          - "!min"
+          - "local"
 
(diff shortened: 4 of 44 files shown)


slagle:
recheck

playbooks/bootstrap.yml
```diff
@@ -15,6 +15,9 @@
   any_errors_fatal: "{{ edpm_any_errors_fatal | default(true) }}"
   max_fail_percentage: "{{ edpm_max_fail_percentage | default(0) }}"
   tasks:
+    - name: Gather ansible_local facts
+      ansible.builtin.setup:
+        filter: ansible_local
```
rabi, line 20: Isn't using `filter`  would be expensive as it collects all facts and then filters `local`? `gather_subset` would be more efficient everywhere I guess.

```
 gather_subset:
      - "!all"
      - "!min"
      - "local"
```

roles/edpm_ddp_package/tasks/main.yml
```diff
@@ -14,6 +14,15 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Gather ansible_local facts
+  ansible.builtin.setup:
+    filter: ansible_local
+
+- name: Fail if bootc
```
rabi, line 21: Are we going to just document this?

roles/edpm_kernel/tasks/kernelargs.yml
```diff
@@ -108,6 +127,50 @@
       changed_when: _grub2_mkconfig.rc == 0
       failed_when: _grub2_mkconfig.rc != 0
 
+
+- name: Kernel args configuration (image mode)
```
rabi, line 131: Do we have any molecule tests for this?

playbooks/bootstrap.yml
rabi: Hmm.. We don't gather facts by default https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/playbooks/bootstrap.yml#L5

Also, this additional task has been added for every role and would be running for non bootc deployments as well and the it takes around 12 secs for a single node in the bootstrap service. My point is this kind of negates the work we had done to disable gather facts by default in edpm. 

TASK [Gather ansible_local facts filter=ansible_local] *************************
Tuesday 16 September 2025  16:09:57 +0000 (0:00:11.483)       0:00:11.514 ***** 
[0;32mok: [edpm-compute-0][0m

slagle: ok, i updated to use:

```
 gather_subset:
      - "!all"
      - "!min"
      - "local"
```

everywhere

rabi (approved):
/lgtm

PR #1020, opened 2025-09-12
Move some vars from internal role vars to default

rabi:
vars/main.yaml is for internal role variables and have a higher precedence. They can't be overridden by inventory vars. Let's move them to default/main.yaml.

We don't see this issue in 17.1 because the tripleo_sshd role has a vars/redhat.yaml which is used instead of vars/main.yaml and it does not have those role vars.

jira: https://issues.redhat.com/browse/OSPRH-19915


rabi:
recheck

PR #1024, opened 2025-09-22
Nvme cleanup container cleaner

auniyal61:
Related: [OSPRH-13022](https://issues.redhat.com//browse/OSPRH-13022)
Closes: [OSPRH-20052](https://issues.redhat.com//browse/OSPRH-20052)

### DIFF
--- a/roles/edpm_nova/tasks/install.yml
+++ b/roles/edpm_nova/tasks/install.yml
@@ -82,3 +82,15 @@
     edpm_container_manage_healthcheck_disabled: true
     edpm_container_manage_config_patterns: 'nova_nvme_cleaner.json'
     edpm_container_manage_clean_orphans: false
+
+- name: Remove nvme cleaner when disabled
+  when: not edpm_nova_enable_nvme_cleaner
+  tags:
+    - install
+    - nova
+    - cleanup
+  ansible.builtin.include_role:
+    name: osp.edpm.edpm_container_rm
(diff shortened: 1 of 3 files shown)


roles/edpm_nova/tasks/install.yml
```diff
@@ -82,3 +82,15 @@
...
+  when: not edpm_nova_enable_nvme_cleaner
+  tags:
+    - install
+    - nova
+    - cleanup
+  ansible.builtin.include_role:
+    name: osp.edpm.edpm_container_rm
+  vars:
+    edpm_containers_to_rm:
+      - nova_nvme_cleaner
```
gibizer, line 96: Does it stops the running container, removes it and then removes the service file as well?

gibizer: Wondering if we can add a new molecule scenario that runs the our ansible role twice once with nvme agent enabled and right after with disabled and then assert that the stuff was removed

auniyal61:
> This looks good to me and I tested it locally with install_yamls and it works as expected. I have one request for adding molecule coverage for the cleanup logic. But that can be done in a follow up PR. Lets merge this so we are good for the feature freeze from dev perspective. The molecule test can land after FF as well.

Added followup PR for molecule test https://github.com/openstack-k8s-operators/edpm-ansible/pull/1029

PR #1025, opened 2025-09-22
Check if the bootc fact exits before evaluating

rabi:
Sometimes these roles are used without the bootstrap role and we should add the check if the bootc fact has been defined to the conditions to avoid backward incompatibility.

Jira: https://issues.redhat.com/browse/OSPCIX-1058


slagle:
I think this is ok, although it could hide other errors where we failed to collect the fact for some reason. I don't know of any of those specific cases, as we tried to cover everything.

PR #1027, opened 2025-09-24
Reload sshd for banner text changes

rabi:
/etc/issue content is not dynamically read by the sshd process for new connections.


roles/edpm_sshd/tasks/configure.yml
```diff
@@ -87,7 +88,7 @@
 
     - name: Set sshd config changed fact
       ansible.builtin.set_fact:
-        _sshd_config_result_changed: _sshd_config_result.changed
+        _sshd_config_result_changed: _sshd_config_result.changed or (_sshd_banner_result.changed | default(false))
```
slagle, line 91: This gets used from run.yml. While this change is correct, it might not actually fix the issue since playbooks/run_os.yml only includes the tasks from run.yml, not configure.yml, so that's kind of a separate bug.

rabi: I suggested to run both `configure-os` and `run-os`.  However, you're right, I was dumb to not realize that these facts wont be available across playbooks running in different containers on different worker nodes.

rabi:
recheck unrelated adoption job failure

slagle (approved):

PR #1030, opened 2025-09-26
Add bootc update tasks to edpm_update_system

slagle:
Exclude the kpatch and packages tasks and instead use the bootc update
tasks for bootc systems. This matches the logic that was previously
added to the edpm_update role before the split edpm_update_system and
edpm_update_services roles were added.

Jira: [OSPRH-19238](https://issues.redhat.com//browse/OSPRH-19238)


slagle:
recheck

slagle:
recheck

PR #1031, opened 2025-09-29
nmstate: save generated network config in a temp file

bogdando:
When debugging templating or nmstate conf issues, it is hard to figure which config it was applying. Store the generated template in a tempfile.

### DIFF
--- a/roles/edpm_network_config/tasks/nmstate_tool.yml
+++ b/roles/edpm_network_config/tasks/nmstate_tool.yml
@@ -34,9 +34,26 @@
 - name: Configure network with network role from system roles [nmstate]
   become: true
   block:
-    - name: Render network_state variable
+    - name: Create temporary file for network_state
+      ansible.builtin.tempfile:
+        state: file
+        suffix: .yaml
+      register: network_state_tempfile
+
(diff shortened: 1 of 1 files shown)


bogdando:
> `edpm_network_config_tool=nmstate`, which is experimental

yes, that is my case

roles/edpm_network_config/tasks/nmstate_tool.yml
```diff
@@ -34,9 +34,26 @@
...
+      ansible.builtin.tempfile:
+        state: file
+        suffix: .yaml
+      register: network_state_tempfile
+
+    - name: Render edpm_network_config_template to temporary file
+      ansible.builtin.copy:
+        content: "{{ edpm_network_config_template }}"
+        dest: "{{ network_state_tempfile.path }}"
+        mode: '0600'
```
slagle: what's the point of storing it in a tempfile? Just so that you can see the output in the ansible log? Or the rendered output? Maybe this should be saved under /var/lib/openstack/network/?

bogdando:
thank you for review! I will return to this when the native nmstate provider becomes a defautl, and I have some extra cycles to refactor this

PR #1034, opened 2025-09-29
Don't try to ensure  network service is running

rabi:
Some adoption jobs have issues when ensuring that network.service has been started possibly because of conflict with NetworkManager. We were not doing that earlier[1], so let's change that.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/blob/18.0-fr2/roles/edpm_network_config/tasks/network_config.yml#L152C13-L159

jira: https://issues.redhat.com/browse/OSPCIX-1042
Related-Issue: https://issues.redhat.com/browse/OSPRH-19018

### DIFF
--- a/roles/edpm_network_config/tasks/package_management.yml
+++ b/roles/edpm_network_config/tasks/package_management.yml
@@ -60,8 +60,7 @@
       retries: "{{ edpm_network_config_download_retries }}"
       delay: "{{ edpm_network_config_download_delay }}"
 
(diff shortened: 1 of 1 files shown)


karelyatin:
/lgtm
was wrongly started as part of previous PR https://github.com/openstack-k8s-operators/edpm-ansible/pull/1012 and now it also matches with what we used to do in previous OSP releases

PR #1035, opened 2025-09-30
Fix growvols shell command YAML syntax

ratailor:
Use folded style for multi-line growvols_args to remove trailing newlines from shell command to prevent execution issue.


bogdando (approved):
Yes, the multi-string values end with \n, which goes to the CLI args and nnop the growvols command (also not reporting errors!)

PR #1039, opened 2025-10-03
Add group_vars in edpm-ansible repo

amartyasinha:
edpm-ansible repo uses molecule config file from ci-framework repo. Recent change in that config file reads group_vars from group_vars dir. Since edpm-ansible don't have one, this commit adds it to make config file happy.


evallesp (approved):

jbadiapa:
I was going to suggest to create it in the Makefile.  https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/Makefile#L29 
As there are several roles that fails.
Keeping in mind that also local test is available.

danpawlik:
Alternative way: https://github.com/openstack-k8s-operators/ci-framework/pull/3385

PR #1067, opened 2025-11-24
Fix: update_system and reboot roles for bootc image update

oliashish:
- Added better edge cases handling along with better error/debug messages

Cases handled for debug/error messages, reboot directory creation and check for needs-restarting for bootc mode
- [x] Fresh installation on a machine
- [x] Update request to same image (same digest)
- [x] Update to new/different image (new or same tag, different digest)
- [x] Already existing staged image
- [x] Better debug/error message (open for suggestions on detailed log info)
- [x] Handling reboot/restart Directory/file creation

Open to any other edge cases I missed here.

### DIFF
--- a/roles/edpm_update_system/tasks/bootc.yml
+++ b/roles/edpm_update_system/tasks/bootc.yml
@@ -18,8 +18,31 @@
 - name: Update bootc image
   ansible.builtin.shell: bootc switch "{{ edpm_update_system_bootc_os_container_image }}"
   become: true
-  register: bootc_result
-  changed_when: bootc_result.rc != 0
+  register: bootc_switch_result
+  failed_when: bootc_switch_result.rc != 0
+  changed_when: false
+
+- name: Get bootc status after switch
+  ansible.builtin.shell: bootc status --format json
+  become: true
+  register: bootc_status
+  changed_when: false
+
+- name: Determine if update was applied
+  ansible.builtin.set_fact:
+    bootc_result:
+      changed: "{{
+        (bootc_status.stdout | from_json).status.staged is not none and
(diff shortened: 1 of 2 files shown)


roles/edpm_update_system/tasks/bootc.yml
```diff
@@ -18,8 +18,31 @@
 - name: Update bootc image
   ansible.builtin.shell: bootc switch "{{ edpm_update_system_bootc_os_container_image }}"
   become: true
-  register: bootc_result
-  changed_when: bootc_result.rc != 0
+  register: bootc_switch_result
+  failed_when: bootc_switch_result.rc != 0
+  changed_when: false
```
slagle: I think we can drop this changed_when.
Unless we want to compare the staged value before and after the switch command, and if the value has changed, then set changed_when based on that.

jistr:
Looks like we worked in parallel on the same thing, i posted my approach here:

https://github.com/openstack-k8s-operators/edpm-ansible/pull/1069 

FWIW i don't think we need to look at what image is staged? I took the approach "if there's a staged image, we want to set the reboot flag".

I didn't touch the os-reboot role in my PR, +1 on your changes for that.

roles/edpm_update_system/tasks/bootc.yml
jistr: We can also drop the `bootc_switch_result` altogether. If the command returns RC != 0, it will fail the task anyway, we don't need to register that var. It's not used anywhere else. (Did that in https://github.com/openstack-k8s-operators/edpm-ansible/pull/1069/files )

oliashish: Yeah, it makes sense. I checked the PR and also realized I'm never using it anywhere and is not required.

PR #1096, opened 2026-01-09
Fix: duplicate jump rules in nftables INPUT chain

oliashish:
The template check was using '{{ edpm_nftables_chains_prefix }}_' as a literal string instead of evaluating the variable. This caused the check to always fail, resulting in duplicate jump rules being inserted for each service that updated firewall rules.

Fix: [OSPRH-16563](https://issues.redhat.com//browse/OSPRH-16563)


rabi (approved):

rabi:
recheck

PR #1104, opened 2026-01-28
Fix: Load empty ruleset does not work in all scenerio

oliashish:
It looks like nftables is not 100% idempotent when applying ruleset. When trying to add a new chain with the same name as an existing chain, which is invalid and will result in "Operation Not Supported". Nftables does not support converting chain types in-place

Jira: [OSPRH-16941](https://issues.redhat.com//browse/OSPRH-16941)

This is a workaround solution for RHOSP, as this is a kernel level constraint.

Explaination for [deleting the chain](https://wiki.nftables.org/wiki-nftables/index.php/Configuring_chains#:~:text=The%20only%20condition%20is%20that%20the%20chain%20you%20want%20to%20delete%20needs%20to%20be%20empty%2C%20otherwise%20the%20kernel%20will%20complain%20that%20the%20chain%20is%20still%20in%20use.)

Detailed read - https://wiki.nftables.org/wiki-nftables/index.php/Configuring_chains

Known issue at ubuntu bugs - [nft cannot load certain rulesets after kernel upgrade](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2035285)


slagle (approved):

slagle:
recheck

oliashish:
recheck

oliashish:
recheck

slagle (approved):

PR #1107, opened 2026-01-30
[frr][neutron_ovn] Changes needed to enable ovn-bgp at ovn-agent

eduolivares:
New ovn-bgp service plugin will be supported soon, replacing ovn-bgp-agent. When the new BGP solution is configured, changes in frr and neutron_ovn roles are necessary.

[OSPRH-22480](https://issues.redhat.com//browse/OSPRH-22480)

### DIFF
--- a/roles/edpm_frr/defaults/main.yml
+++ b/roles/edpm_frr/defaults/main.yml
@@ -90,3 +90,8 @@ edpm_frr_images_download_delay: "{{ edpm_download_delay | default(60) }}"
 
 # number of retries for download tasks
 edpm_frr_images_download_retries: "{{ edpm_download_retries | default(5) }}"
+
+# ovn-bgp extension related vars
+edpm_frr_ovn_vrf_advertise: ''
+edpm_frr_ovn_vrf_learn: ''
+edpm_frr_conf_custom_router_bgp_ovn: ''
--- a/roles/edpm_frr/templates/config/frr.conf.j2
+++ b/roles/edpm_frr/templates/config/frr.conf.j2
@@ -9,6 +9,62 @@ service integrated-vtysh-config
 line vty
 {{ edpm_frr_conf_custom_globals }}
 
+{% if edpm_frr_ovn_vrf_learn|length %}
+vrf {{ edpm_frr_ovn_vrf_learn }}
+{% if edpm_frr_bgp_ipv4 %}
+  ip protocol bgp route-map rm-install-only-default
+{% endif %}
+{% if edpm_frr_bgp_ipv6 %}
+  ipv6 protocol bgp route-map rm-install-only-default
+{% endif %}
+exit-vrf
+
+router bgp {{ edpm_frr_bgp_asn }} vrf {{ edpm_frr_ovn_vrf_learn }}
+{% if edpm_frr_bgp_ipv4 %}
+  address-family ipv4 unicast
+    ! Leak everything from root (including dynamic next-hops)
+    import vrf default
+  exit-address-family
+{% endif %}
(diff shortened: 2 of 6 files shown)


ccamposr (approved):

roles/edpm_neutron_ovn/meta/argument_specs.yml
```diff
@@ -102,3 +102,9 @@ argument_specs:
         default: /var/lib/openstack/neutron-ovn-agent
         description: 'The directory that contains configuration files for Neutron OVN Agent.'
         type: str
+      edpm_neutron_ovn_agent_bgp_peer_bridges:
+        default: ''
+        description: >
+          BGP peer bridges for the ovn-bgp extension. Set this variable when the
```
cubeek: Maybe we can tell what the BGP peer bridge is - that the BGP peer bridge connects the OVN overlay with the peering leaf switch on the given node, or something along those lines :)

roles/edpm_frr/defaults/main.yml
```diff
@@ -90,3 +90,8 @@ edpm_frr_images_download_delay: "{{ edpm_download_delay | default(60) }}"
 
 # number of retries for download tasks
 edpm_frr_images_download_retries: "{{ edpm_download_retries | default(5) }}"
+
+# ovn-bgp extension related vars
+edpm_frr_ovn_vrf_expose: ''
```
cubeek: nit: Maybe `advertise` is better to align with the OVN terminology?

cubeek (approved):

roles/edpm_neutron_ovn/meta/argument_specs.yml
```diff
@@ -102,3 +102,10 @@ argument_specs:
         default: /var/lib/openstack/neutron-ovn-agent
         description: 'The directory that contains configuration files for Neutron OVN Agent.'
         type: str
+      edpm_neutron_ovn_agent_bgp_peer_bridges:
+        default: ''
+        description: >
+          BGP peer bridges required for the ovn-bgp extension.
+          These bridges connects the OVN overlay from the corresponding EDPM node with its BGP peering switches (leaf, ToR, ...).
```
cubeek: "bridges connect"

cubeek (approved):

roles/edpm_frr/templates/config/frr.conf.j2
```diff
@@ -9,6 +9,62 @@ service integrated-vtysh-config
...
+  address-family ipv6 unicast
+    ! Leak everything from root (including dynamic next-hops)
+    import vrf default
+  exit-address-family
+{% endif %}
+exit
+
+{% if edpm_frr_bgp_ipv4 %}
+route-map rm-install-only-default permit 10
+  match ip address prefix-list only-default
```
rabi, line 39: Where are these prefix-list  `only-default` is expected to be defined?

rabi (commented):
A question inline, otherwise LGTM.

roles/edpm_frr/templates/config/frr.conf.j2
eduolivares: Line 164
https://github.com/openstack-k8s-operators/edpm-ansible/pull/1107/changes#diff-3cad81f26cfe4a3cd3e29e326de257ec73f196a38c00a5f6707c0c73ce8e7b56R164

We need two route-maps for the same prefix-list because:
`route-map rm-install-only-default` doesn't include `set src`.
`route-map rm-only-default` includes `set src`.

rabi (approved):

PR #1147, opened 2026-03-20
[edpm_users] Prevent system user account expiration on hardened systems

slagle:
Set account expiration to never expire for system users created by
the edpm_users role. This prevents authentication failures on systems
with PASS_MAX_DAYS configured in /etc/login.defs.

Jira: [OSPRH-26189](https://redhat.atlassian.net/browse/OSPRH-26189)
Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>


rabi (approved):
OK, I noticed the linked jira is talking about a workaround[1] for an existing deployment where as part of hardening they have set PASS_MAX_DAYS to something much smaller than the default. 

Also, unless we make it same across all nodes in a nodeset, customers would end up with different expiry across nodes after scale-out and it would probably be  difficult for them to manage.

[1] https://redhat.atlassian.net/browse/OSPRH-26189?focusedCommentId=11131451

PR #1148, opened 2026-03-23
Quadlet Migration

rabi:
Convert existing container definitions to Quadlet .container files and use podman quadlet for managing containers and their systemd units.


slagle (commented):
I don't think we should migrate any existing environments to quadlet in 18.

For 19, we can completely get rid of the container config json file, which is a completely bespoke format. Instead just have the container unit files for each service, with templated parts where needed. I don't think we would even necessarily need an edpm role wrapping that logic, the service roles (edpm_ovn, etc) can just use the linux system podman role directly. That would enable us to get rid of both edpm_container_standalone and edpm_container_manage in 19.

On upgrade from 18 to 19, we can have a role to stop the old containers, clean up the old systemd units, and then just run the deployment code to deploy the new quadlet style.

This work will likely need to be done in a feature branch until main is re-targeted to be for 19, b/c I don't think the new functionality should be going into 18.

(I just marked this one as "requested changes" so it doesn't merge, since it's not marked as draft)

slagle (changes requested):

PR #1157, opened 2026-04-13
Share NIC mapping tasks for nmstate and os-net-config

karthiksundaravel:
Add nic_mapping.yml to merge edpm_network_config_os_net_config_mappings with edpm_network_config_mappings (recursive combine) and run edpm_os_net_config_mappings, failing when lookup data is present but no host match is found, then write the resolved YAML to the path chosen by each tool (template_nic_mapping.yaml for nmstate, /etc/os-net-config/ mapping.yaml for os-net-config).

Remove the duplicated mapping block from the os-net-config apply path and document both mapping variables in defaults and meta/argument_specs, including that edpm_network_config_mapping_file is internal and that extra vars can override set_fact.

Assisted-by: Cursor Agent

Signed-off-by: Karthik Sundaravel <ksundara@redhat.com>

PS: I'll make a followup PR that does the NIC mapping for nmstate tool

### DIFF
--- a/roles/edpm_network_config/meta/argument_specs.yml
+++ b/roles/edpm_network_config/meta/argument_specs.yml
@@ -55,7 +55,20 @@ argument_specs:
         default: true
       edpm_network_config_os_net_config_mappings:
         type: dict
-        description: "Per node and/or per node group configuration map. Used by the edpm_os_net_config_mappings module."
+        description: >
+          Legacy role variable: per-node / node-group NIC mapping for the nmstate and os-net-config
+          paths (merged in nic_mapping.yml). Merged with edpm_network_config_mappings; the latter
+          wins on duplicate keys. The merged dict is passed to the edpm_os_net_config_mappings Ansible
+          module as net_config_data_lookup.
+        default: {}
+      edpm_network_config_mappings:
+        type: dict
+        description: >
+          Optional additional or overriding NIC mapping for the nmstate and os-net-config paths
+          (merged in nic_mapping.yml). Merged with edpm_network_config_os_net_config_mappings; this
+          variable wins on duplicate keys.
+          Must be a YAML mapping (dict). Do not use a block literal (|) after the key — that
+          produces a string and role validation will fail.
         default: {}
       edpm_network_config_purge_provider:
         type: str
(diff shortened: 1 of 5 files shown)


roles/edpm_network_config/meta/argument_specs.yml
```diff
@@ -55,7 +55,20 @@ argument_specs:
...
-        description: "Per node and/or per node group configuration map. Used by the edpm_os_net_config_mappings module."
+        description: >
+          Legacy role variable: per-node / node-group NIC mapping for the nmstate and os-net-config
+          paths (merged in nic_mapping.yml). Merged with edpm_network_config_mappings; the latter
+          wins on duplicate keys. The merged dict is passed to the edpm_os_net_config_mappings Ansible
+          module as net_config_data_lookup.
+        default: {}
+      edpm_network_config_mappings:
+        type: dict
+        description: >
```
vcandapp, line 77: Description can be updated to clarify points like:
- Is this variable to be set "ONLY" when edpm_network_config_tool is "nmstate" Or, is it backward compatible with tool "os_net_config"
- Clarify that default_mapping (ie nic name) will be used ?

roles/edpm_network_config/tasks/nmstate_tool.yml
```diff
@@ -23,6 +23,14 @@
     # For nmstate, always enable NetworkManager DNS management
     edpm_bootstrap_network_resolvconf_update: true
 
+- name: Set NIC mapping file path for nmstate path (internal, not user-overridable)
+  ansible.builtin.set_fact:
+    edpm_network_config_mapping_file: /var/lib/edpm-config/template_nic_mapping.yaml
```
vcandapp, line 28: For os-net-config tool, the config.yaml and mapping.yaml path was same = /etc/os-net-config/. Is this updated now ?

karthiksundaravel: This is for nmstate tool. For os-net-config tool the path remains "/etc/os-net-config/mapping.yaml" as set in roles/edpm_network_config/tasks/os_net_config_tool.yml

roles/edpm_network_config/meta/argument_specs.yml
karthiksundaravel: Done

PR #1158, opened 2026-04-17
Generate SSL certificates for ovn-controller

slawqo:
With this patch instead of using the same SSL certificate by ovn-controller on each EDMP node in the environment, there is separate certificate generated, with uniq CN name which match system-id set in that chassis and signed with certificate from SB.
That way OVN RBAC can be used for the connections from ovn-controller running on the EDPM nodes to the OVS SB database.

Certificates are generated in the ansibleee POD as that's where secret with the OVN SB Certificate is also mounted.

Closes: #[OSPRH-1921](https://redhat.atlassian.net/browse/OSPRH-1921)

### DIFF
--- a/roles/edpm_ovn/defaults/main.yml
+++ b/roles/edpm_ovn/defaults/main.yml
@@ -68,13 +68,18 @@ edpm_ovn_controller_tls_cacert_volumes: []
 
 edpm_ovn_controller_tls_volumes:
   - "/var/lib/openstack/certs/{{ edpm_ovn_service_name }}/default/ca.crt:/etc/pki/tls/certs/ovndbca.crt:ro,z"
-  - "/var/lib/openstack/certs/{{ edpm_ovn_service_name }}/default/tls.crt:/etc/pki/tls/certs/ovndb.crt:ro,z"
-  - "/var/lib/openstack/certs/{{ edpm_ovn_service_name }}/default/tls.key:/etc/pki/tls/private/ovndb.key:ro,Z"
+  - "/var/lib/openstack/certs/{{ edpm_ovn_service_name }}/rbac/tls.crt:/etc/pki/tls/certs/ovndb.crt:ro,z"
+  - "/var/lib/openstack/certs/{{ edpm_ovn_service_name }}/rbac/tls.key:/etc/pki/tls/private/ovndb.key:ro,Z"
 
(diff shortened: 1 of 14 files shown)


roles/edpm_ovn/defaults/main.yml
```diff
@@ -66,15 +66,35 @@ edpm_ovn_controller_tls_cacert_bundle_src: "/var/lib/openstack/cacerts/{{ edpm_o
...
+
+# OVN controller certificate paths used in the kolla config command line.
+# When TLS is enabled, per-node RBAC certs are used for OVN RBAC identity.
+edpm_ovn_controller_cert: "{{ edpm_ovn_rbac_cert_path }}"
+edpm_ovn_controller_privkey: "{{ edpm_ovn_rbac_key_path }}"
+edpm_ovn_controller_cacert: "/etc/pki/tls/certs/ovndbca.crt"
+
 edpm_ovn_controller_tls_volumes:
-  - "/var/lib/openstack/certs/{{ edpm_ovn_service_name }}/default/ca.crt:/etc/pki/tls/certs/ovndbca.crt:ro,z"
-  - "/var/lib/openstack/certs/{{ edpm_ovn_service_name }}/default/tls.crt:/etc/pki/tls/certs/ovndb.crt:ro,z"
```
slagle, line 71: these certs are already per node aren't they? and generated by certmanager. why do we need to switch to generating with openssl?

slawqo: To use RBAC in OVN ovn-controller services on each node have to use certificates generated with CN name matching "system-id" and signed with the certificate from OVN SB. That's why I changed it to be generated with ansible as that way we can also set system-id which will be configured for ovn-controller service on each node.

slagle: If we generate system-id per node ahead of time in openstack-operator, then could we use cert-manager to generate the certs? We could set a var in the inventory per node for system-id. I'd like to be able to do that generically in some fashion, but if not then I think this specific case for ovn would be ok.

I'm concerned about switching to use openssl as we lose the management of the certs that cert-manager provides.

slawqo: I changed it, now certificates are generated by certmanager

slagle (commented):
@slawqo this lgtm now, do still plan to move forward with this?

PR #1160, opened 2026-04-20
edpm_network_config: nmstate path NIC mapping and nmstate alias substitution

karthiksundaravel:
- Derive interface_mapping via edpm_derived_nic_mapping.py (PyYAML dep, dummy/Molecule
    handling, skip rewrite when mapping unchanged; stdout marker for Ansible changed).
  - Substitute aliases in nmstate YAML via collection filter edpm_substitute_nic_aliases.
  - Document RHEL-only package list; clarify check mode does not refresh derived mapping.
  - Ensure /var/lib/edpm-config and nmstate.returncode handling for the system network role.
  
  Assisted-by: Cursor AI
  
  Signed-off-by: Karthik Sundaravel <ksundara@redhat.com>

### DIFF
--- a/roles/edpm_network_config/files/edpm_derived_nic_mapping.py
+++ b/roles/edpm_network_config/files/edpm_derived_nic_mapping.py
@@ -1,0 +1,303 @@
+#!/usr/bin/env python3
+# Copyright 2026 Red Hat, Inc.
+# Licensed under the Apache License, Version 2.0
+#
+# EDPM-only NIC alias -> Linux interface name resolution. No dependency on
+# os-net-config: mapping logic lives entirely in this module.
+#
+# Reads and writes /var/lib/edpm-config/derived_nic_mapping.yaml by default
+# (same as role var edpm_network_config_derived_nic_mapping_file; override with -f).
+# YAML shape:
+#   interface_mapping:
+#     nic1: <MAC or interface name>
+#     nic2: ...
+#
+# After a run, values are resolved to Linux interface names; unmapped active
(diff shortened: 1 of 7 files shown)


danpawlik:
recheck

roles/edpm_network_config/tasks/nmstate_tool.yml
```diff
@@ -31,6 +31,77 @@
...
+      ansible.builtin.command:
+        argv:
+          - python3
+          - -c
+          - "import yaml"
+      register: edpm_derived_nic_mapping_pyyaml_check
+      changed_when: false
+      failed_when: false
+
+    - name: Fail when PyYAML is not available for derived NIC mapping
```
vcandapp, line 77: This can be moved to the already existing section here - https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_network_config/defaults/main.yml#L34

roles/edpm_network_config/files/edpm_derived_nic_mapping.py
```diff
@@ -0,0 +1,303 @@
...
+                        ", ".join(available_nics),
+                    )
+                    continue
+            elif nm not in available_nics:
+                LOG.error(
+                    "nic %s not found in available nics %s",
+                    nic_mapped,
+                    ", ".join(available_nics),
+                )
+                continue
```
vcandapp, line 177: Are invalid mapping errors silently ignored ?

vcandapp:
@karthiksundaravel ,what is the use case for adoption - can we try to use this existing mapping file as input ?

PR #1161, opened 2026-04-22
Fix: edpm_container_quadlet role to support .container quadlet type

oliashish:
Switch from static .kube file to dynamically templated .container quadlet units as the default, while preserving .kube as an option via edpm_container_quadlet_type. Volumes simplified for structured dicts to plain Podman-format strings. Molecule tests updated with content assertions for both types.

Fix: [OSPRH-29433](https://redhat.atlassian.net/browse/OSPRH-29433)


roles/edpm_container_quadlet/defaults/main.yml
```diff
@@ -1,55 +1,27 @@
...
 edpm_container_quadlet_service_use_host_network: false
-edpm_container_quadlet_service_image: ""
-# List of common volumes that can be included in the template of a container
-# definition to include the most common volumes.
+edpm_container_quadlet_service_privileged: true
+edpm_container_quadlet_service_environment: {}
+edpm_container_quadlet_service_healthcheck_cmd: ""
+edpm_container_quadlet_service_after: []
+edpm_container_quadlet_service_wants: []
+
```
oliashish, line 15: 1. **edpm_container_quadlet_service_privileged: true**
Moved from hardcoded privileged: true in the old Pod YAML/container_defs templates into a configurable variable; defaults to true since nearly all EDPM services need it.

2. **edpm_container_quadlet_service_environment: {}**
 Replaces per-service environment: blocks from old container_defs templates; KOLLA_CONFIG_STRATEGY=COPY_ALWAYS is always included in the template, this dict is for extras.                                        
 
3. **edpm_container_quadlet_service_healthcheck_cmd: ""**        
Replaces the old healthcheck.test/healthcheck.mount fields that container_systemd parsed into separate timers; now uses Podman's native HealthCmd= directive directly. 

4. **edpm_container_quadlet_service_after: [] / edpm_container_quadlet_service_wants: []**                                                                                                                               
Previously hardcoded as After=openvswitch.service in the static .kube file for all services; now per-service so only services that actually depend on OVS (or anything else) declare it.

roles/edpm_container_quadlet/molecule/default/defaults/converge.yml
```diff
@@ -15,35 +15,122 @@
...
-    - "osp.edpm.edpm_container_quadlet"
+  tasks:
+    - name: Run edpm_container_quadlet role
+      ansible.builtin.include_role:
+        name: osp.edpm.edpm_container_quadlet
+        tasks_from: main.yml
+
+    - name: Verify .container quadlet file was rendered
+      ansible.builtin.stat:
+        path: /etc/containers/systemd/edpm@test_container.container
```
oliashish, line 48: @slagle is this naming `edpm@<service_name>.container` and `edpm@<service_name>.service` okay to use? I read some Jira discussions and this seems fine. let me know if we are okay going forward with this.

slagle: yes, that looks fine

roles/edpm_container_quadlet/tasks/cleanup.yml
```diff
@@ -1,13 +1,23 @@
 ---
-- name: Stop and disable old standalone container for {{ edpm_container_quadlet_service }}
+- name: "Stop and disable old standalone service for {{ edpm_container_quadlet_service }}"
   ansible.builtin.systemd_service:
     name: "edpm_{{ edpm_container_quadlet_service }}"
     state: stopped
     enabled: false
   become: true
+  failed_when: false
```
slagle, line 8: why failed_when: false? If it fails to stop, we probably want to know

roles/edpm_container_quadlet/tasks/main.yml
```diff
@@ -32,28 +30,75 @@
...
+      unique=($(printf '%s\n' "${resolved[@]}" | sort -u))
+      find "${unique[@]}" -type f -print0 \
+        | sort -z \
+        | xargs -0 -r sha256sum \
+        | sha256sum \
+        | cut -d' ' -f1
+    executable: /bin/bash
+  register: _edpm_config_hash_result
+  changed_when: false
+  when: _edpm_quadlet_config_vol_paths | length > 0
```
slagle, line 67: this looks ok , but I would probably just stick to using the implementation in python if we can't get rid of the hashing altogether

slagle (commented):
just a couple of small comments inline, otherwise, this looks pretty good

oliashish:
Superseeded by #1175 
Moving with edpm_container_standalone role for quadlet changes. No new role required as per team discussion.

PR #1164, opened 2026-04-29
Fix test_sidecar_container molecule test for newer kernels

eduolivares:
Newer kernels (6.19+) tighten cross-user-namespace setns()
restrictions, causing two failures in the molecule test:

1. ip netns exec inside the agent container fails with EINVAL when
   the namespace was created on the instance (different user namespace).
   Fix: create the namespace inside the agent container to match
   production behavior.

2. The kill-script's nsenter can't cross user namespaces after an
   agent container restart. Fix: reorder the test to run the
   kill-script before the stop/start cycle (same container context),
   then recreate the sidecar for the survival test.

[OSPRH-29561](https://redhat.atlassian.net/browse/OSPRH-29561)

Assisted-By: Claude Opus 4.6 <noreply@anthropic.com>
Signed-off-by: Eduardo Olivares <eolivare@redhat.com>


karelyatin (approved):
/lgtm

karelyatin (approved):
/lgtm

PR #1165, opened 2026-04-30
Deduplicate resolved config roots

rabi:
Avoid repeated EDPM_CONFIG_HASH entries when multiple mounts resolve to the same service-scoped config tree. Add  coverage for direct config roots and repeated mounts that should hash only once.

jira: [OSPRH-29555](https://redhat.atlassian.net/browse/OSPRH-29555)


rabi:
Now addressed in https://github.com/openstack-k8s-operators/edpm-ansible/pull/1166

PR #1166, opened 2026-04-30
Fix: container_config_hash over-scoped directory hashing

oliashish:
get_config_base() walked up from each volume path to the first child of /var/lib/openstack/, then hashed all files under that shared parent. This caused a single file change (e.g. one service's cert rotation) to change the EDPM_CONFIG_HASH for every service mounting from the same parent directory, triggering unnecessary container restarts.

Introduce SHARED_CONFIG_NAMESPACES (certs, cacerts, configs) to scope
the walk-up to the service-level directory under shared namespaces to prevent duplicate hash generation for same service

Fix: [OSPRH-29551](https://redhat.atlassian.net/browse/OSPRH-29551) [OSPRH-29555](https://redhat.atlassian.net/browse/OSPRH-29555)

### DIFF
--- a/plugins/modules/container_config_hash.py
+++ b/plugins/modules/container_config_hash.py
@@ -65,6 +65,7 @@ EXAMPLES = """
 
 CONTAINER_STARTUP_CONFIG = '/var/lib/edpm-config/container-startup-config'
 BUF_SIZE = 65536
+SHARED_CONFIG_NAMESPACES = ('certs', 'cacerts', 'configs')
 
 
 class ContainerConfigHashManager:
@@ -178,14 +179,15 @@ class ContainerConfigHashManager:
         :param volume: string
         :returns: string
         """
-        # crawl the volume's path upwards until we find the
-        # volume's base, where the hashed config file resides
         path = volume
         base = prefix.rstrip(os.path.sep)
-        base_generated = os.path.join(base, 'ansible-generated')
+        shared_bases = tuple(
+            os.path.join(base, namespace)
+            for namespace in SHARED_CONFIG_NAMESPACES
+        )
         while path.startswith(prefix):
             dirname = os.path.dirname(path)
-            if dirname == base or dirname == base_generated:
+            if dirname == base or dirname in shared_bases:
                 return path
             else:
                 path = dirname
@@ -208,8 +210,12 @@ class ContainerConfigHashManager:
             self.module.fail_json(
                 msg='Error fetching volumes. Prefix: '
                     '{} - Config: {}'.format(prefix, config))
-        return sorted([self._get_config_base(prefix, v.split(":")[0])
-                       for v in volumes if v.startswith(prefix)])
+        matched = set()
+        for v in volumes:
+            host_path = v.split(":")[0]
+            if host_path.startswith(prefix):
+                matched.add(self._get_config_base(prefix, host_path))
+        return sorted(matched)
 
     def _update_hashes(self):
         """Update container startup config with new config hashes if needed.


plugins/modules/container_config_hash.py
```diff
@@ -208,8 +201,12 @@ def _match_config_volumes(self, config):
             self.module.fail_json(
                 msg='Error fetching volumes. Prefix: '
                     '{} - Config: {}'.format(prefix, config))
-        return sorted([self._get_config_base(prefix, v.split(":")[0])
-                       for v in volumes if v.startswith(prefix)])
+        matched = set()
```
rabi, line 213: hmm.. with this if a service has multiple sibling subdirectories, using the direct mounted path (or its parent for file mounts) creates multiple hashes for the same service instead of a single service-scoped hash root. For example, mounts under /var/lib/openstack/certs/ovn/default and /var/lib/openstack/certs/ovn/internal would be hashed separately, even if the intended unit is the full /var/lib/openstack/certs/ovn tree.

oliashish: The way I was thinking about it was that the fix scopes the hash to what the service actually mounts rather than what belongs to the service at the directory level. So the hash segments are driven by the volume list which resolves to its immediate parent directory, gets deduplicated, and hashed independently. PodmanManager does plain string equality on the full hash, so more segments don't change the restart behavior, any segment changing still triggers a restart.                                      

But what you also makes total sense to me, and sounds more logical to have service-level directory tree instead (certs/ovn/ rather than certs/ovn/default/). But I think the mount based approach only hashes files the container actually uses, while the service tree approach would also pick up changes to sibling files the container doesn't mount (if any exist). 

To be honest, I'm not sure which is more appropriate here as I don't know exact service level directory structure. Do you think there are cases where files exist in a service's directory but aren't mounted, and changes there should still trigger a restart?

oliashish: I just saw you're also working on something similar here - https://github.com/openstack-k8s-operators/edpm-ansible/pull/1165 right? I think this resolve the same issue.

rabi: #1165 is different, though touching the same area of the code. Currently there are duplicate hashes created for three mounts which point to the same root (i.e _/var/lib/openstack/certs_) for there different volume mounts.

"EDPM_CONFIG_HASH": "3f6c48674a9a9ef3b95cfecc2a29c44ac2365f7e5eb877ce2dd508eb68a0d55a-**185e0165fd292d786437e18d35ec16b62d52255af563f394eac08030817e3109-185e0165fd292d786437e18d35ec16b62d52255af563f394eac08030817e3109-185e0165fd292d786437e18d35ec16b62d52255af563f394eac08030817e3109**-0823bd3e096c75f72e4a95820d41b0d4b6a1172bd2892ddb9f29b788a11bc87d"


The issue you're working on is different as it uses same root for all services i.e uses the first child in /var/lib/openstack. 

Looking at your fix, it uses the volume mount dir, so if there are certs/ovn/default/* and cert/ovn/internal/* there would be 2 hashes created unnecessarily (one for default and other for internal) where as we only need one for cert/<service>. Therefore, I suggested the other approach as in https://privatebin.corp.redhat.com/?62ae10145e9c713a#8tqzPjBYFftAik7DvHg3Ch2XDyPnH3dJp3rg6rtFz9BH.

oliashish: Hi @rabi , what do you think of the new changes now. Let me know if you have any suggestions.

rabi: You can add OSPRH-29555 in this PR and I'll close [my PR](https://github.com/openstack-k8s-operators/edpm-ansible/pull/1165) as this one would also deduplicate.

rabi (approved):

rabi:
> Use the direct volume path instead: directory as-is, file paths resolved to their parent directory

commit message does not look correct, can you please update it and this has to be backported to 18.0-fr5

rabi:
You don't have to add merge commits, we don't squash and merge.

PR #1175, opened 2026-05-15
Fix: rework edpm_container_standalone for quadlet-only deployment

oliashish:
Reworked the standalone role to use only Quadlet for container deployment, removing the old container_manage path. It now detects config and certificate changes by computing a hash before rendering templates, so containers restart only when something actually changed. Added automatic cleanup of leftover artifacts from the old deployment method and molecule tests covering the full lifecycle - deploy, idempotency, config-change restart, and cleanup.

Jira: [OSPRH-30094](https://redhat.atlassian.net/browse/OSPRH-30094)
Jira: [OSPRH-30963](https://redhat.atlassian.net/browse/OSPRH-30963)

### DIFF
--- a/molecule/test-helpers/verify_systemd_unit.yaml
+++ b/molecule/test-helpers/verify_systemd_unit.yaml
@@ -1,16 +1,16 @@
 
 - name: verify systemd unit exists
   block:
-    - name: Check if systemd file exists {{ item.name }}
+    - name: Query systemd for unit fragment path {{ item.name }}
       become: true
-      ansible.builtin.stat:
-        path: "/etc/systemd/system/{{ item.name }}"
-      register: unit_exists
+      ansible.builtin.command: systemctl show {{ item.name }} -p FragmentPath --value
+      register: unit_fragment
+      changed_when: false
       when: item.osp_service | default(true)
     - name: Assert systemd unit exists {{ item.name }}
       ansible.builtin.assert:
         that:
-          - unit_exists.stat.exists
+          - unit_fragment.stdout | length > 0
         fail_msg: "Systemd unit file for {{ item.name }} does not exist"
       when: item.osp_service | default(true)
     - name: Check if systemd unit is enabled {{ item.name }}
@@ -21,10 +21,13 @@
       changed_when: false
       ignore_errors: true
     - name: Assert systemd unit is enabled {{ item.name }}
+      vars:
+        _allowed_states: >-
+          {{ item.enabled | default(["enabled", "generated"]) }}
       ansible.builtin.assert:
         that:
-          - unit_enabled.stdout == item.enabled | default("enabled")
-        fail_msg: "Systemd unit {{ item.name }} is not enabled"
+          - unit_enabled.stdout in (_allowed_states if _allowed_states is not string else [_allowed_states])
+        fail_msg: "Systemd unit {{ item.name }} is not enabled (got '{{ unit_enabled.stdout }}')"
     - name: Check if systemd unit is running {{ item.name }}
       when: item.running | default(true)
       ansible.builtin.command: systemctl is-active {{ item.name }}
--- a/plugins/modules/container_config_hash.py
+++ b/plugins/modules/container_config_hash.py
@@ -23,6 +23,7 @@ import glob
 import hashlib
 import json
 import os
+import re
 import yaml
 
 
@@ -42,7 +43,12 @@ version_added: '2.9'
 short_description: Generate config hashes for container startup configs
 notes: []
 description:
-  - Generate config hashes for container startup configs
+  - Compute SHA-256 hashes from container config volume mounts.
+    When called with 'quadlet_staging_dir' and 'containers', reads
+    Volume= lines from each staged Quadlet .container file, computes
+    the hash, and writes it into the 'config_hash' comment placeholder
+    (Quadlet path). When called without parameters, scans container
+    startup configs and updates them in place (container_manage path).
 requirements:
   - None
 options:
@@ -51,6 +57,18 @@ options:
       - Ansible check mode is enabled
     type: bool
     default: False
+  quadlet_staging_dir:
+    description:
+      - Path to the directory containing staged Quadlet .container files.
+        Used together with 'containers' to compute and inject config hashes.
+    type: path
+    default: ''
+  containers:
+    description:
+      - List of container names to process in the staging directory.
+        Each name maps to a staged file named edpm_<name>.container.
+    type: list
+    default: []
   config_vol_prefix:
     description:
       - Config volume prefix
@@ -61,18 +79,28 @@ options:
 EXAMPLES = """
 - name: Update config hashes for container startup configs
   container_config_hash:
+
+- name: Inject config hashes into staged Quadlet files
+  container_config_hash:
+    quadlet_staging_dir: /var/lib/edpm-config/quadlet-rendered
+    containers:
+      - ovn_controller
 """
 
 CONTAINER_STARTUP_CONFIG = '/var/lib/edpm-config/container-startup-config'
 BUF_SIZE = 65536
-SHARED_CONFIG_NAMESPACES = ('certs', 'cacerts', 'configs')
+SHARED_CONFIG_NAMESPACES = ('certs', 'cacerts', 'config', 'healthchecks')
 
 
 class ContainerConfigHashManager:
-    """Notes about this module.
+    """Generate config hashes for container change detection.
 
-    It will generate container config hash that will be consumed by
-    the edpm-container-manage role that is using podman_container module.
+    Quadlet path: reads Volume= lines from staged .container files in
+    the staging directory, computes hashes, and writes them into the
+    '# config_hash=' comment placeholder in each file.
+
+    Container_manage path (legacy): scans JSON startup configs and updates
+    EDPM_CONFIG_HASH in place.
     """
 
     def __init__(self, module, results):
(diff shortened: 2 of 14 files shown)


plugins/modules/container_config_hash.py
```diff
@@ -39,40 +37,47 @@
...
+      - "/var/lib/openstack/certs/ovn/default/ca.crt:/etc/pki/tls/certs/ca.crt:ro,z"
+      - "/var/lib/openstack/healthchecks/ovn_controller:/openstack:ro,z"
+  register: result
+
+# result.hash contains the computed SHA-256 hash string
 """
 
-CONTAINER_STARTUP_CONFIG = '/var/lib/edpm-config/container-startup-config'
 BUF_SIZE = 65536
 SHARED_CONFIG_NAMESPACES = ('certs', 'cacerts', 'configs')
```
oliashish: I can add config, because some service use that as well. And healthcheck also if that is required for hash calculation?
So it would look like
SHARED_CONFIG_NAMESPACES = ('certs', 'cacerts', 'configs', 'config', 'healtchecks'). 
Let me know your thoughts

slagle: we should probably add them.

roles/edpm_container_standalone/tasks/cleanup.yml
slagle, line 1: we may want to consider splitting cleanup tasks between a stop/disable and a delete. Just in case a rollback is needed. Just an idea ... it might not be needed.

slagle (commented):
Overall the change looks good. However, it might be challenging to land this as-is since it removes the old method as well. It means this would need to all be in the same PR with all other service migrations to quadlet.

And, there might be partners or customers using edpm_container_standalone as-is. So, we need to think about how to handle that as well. Might need some combination of this and what Rabi was originally doing in his PR as well, so that backwards compatibility is maintained for some time period.

roles/edpm_container_standalone/tasks/main.yml
```diff
@@ -43,32 +43,13 @@
...
-  ansible.builtin.copy:
-    content: "{{ item.value | to_nice_json }}"
-    dest: "{{ edpm_container_startup_config_dir }}/{{ edpm_container_standalone_service }}/{{ item.key }}.json"
-    mode: "0644"
-  # NOTE(tkajinam): Some containers can contain secrets in their environments.
-  #                 Hide the output to avoid dumping these to output.
-  no_log: true
-  loop: "{{ edpm_container_standalone_container_defs | dict2items }}"
+- name: Clean up old container_manage artifacts
+  ansible.builtin.include_tasks: cleanup.yml
```
rabi: IMO, we should not worry about cleanup of old artifacts in this role. They can be cleaned up separately as an upgrade task.

rabi (commented):
Yeah this can't be merged in main without keeping the old logic unless you migrate all services in one PR along with this. As mentioned we should do it incrementally on main keeping things working on the main after FR6 branch has been cut.

cleanup tasks have to be upgrade tasks as we don't want to run them for every deploy.  It does not belong in edpm_container_standalone role tbh.

roles/edpm_container_standalone/tasks/cleanup.yml
oliashish: I think Rabi solution is better here, we can have an upgrade role so that we don't run it every time after the first time.

roles/edpm_container_standalone/tasks/main.yml
oliashish: ack

plugins/modules/container_config_hash.py
```diff
@@ -51,6 +57,13 @@
       - Ansible check mode is enabled
     type: bool
     default: False
+  staging_file:
```
rabi: staged_quadet_def may be better name?

plugins/modules/container_config_hash.py
```diff
@@ -61,18 +74,27 @@
...
+  container_config_hash:
+    staging_file: /var/lib/edpm-config/quadlet-rendered/edpm_ovn_controller.container
+  register: result
+# result.hash contains the computed SHA-256 hash string
 """
 
 CONTAINER_STARTUP_CONFIG = '/var/lib/edpm-config/container-startup-config'
 BUF_SIZE = 65536
-SHARED_CONFIG_NAMESPACES = ('certs', 'cacerts', 'configs')
+SHARED_CONFIG_NAMESPACES = ('certs', 'cacerts', 'configs', 'config', 'healthchecks')
```
rabi: why `config`?

roles/edpm_container_standalone/molecule/quadlet/templates/test_service_updated.container.j2
```diff
@@ -0,0 +1,23 @@
...
+[Container]
+ContainerName=test_service
+Image=quay.io/centos/centos:stream9
+Network=host
+User=root
+PodmanArgs=--privileged
+LogDriver=journald
+Exec=sleep 7200
+Environment=KOLLA_CONFIG_STRATEGY=COPY_ALWAYS
+Environment=EDPM_CONFIG_HASH=
```
rabi: We probably don't need this as environment as it would be misleading, can be just a comment  `# config_hash=<sha256>`

roles/edpm_container_standalone/tasks/quadlet.yml
```diff
@@ -0,0 +1,84 @@
...
+- name: "Render Quadlet templates for {{ edpm_container_standalone_service }}"
+  become: true
+  ansible.builtin.template:
+    src: "{{ item.value }}"
+    dest: "{{ edpm_container_standalone_quadlet_staging_dir }}/edpm_{{ item.key }}.container"
+    mode: "0644"
+  loop: "{{ edpm_container_standalone_quadlet_defs | dict2items }}"
+  loop_control:
+    label: "{{ item.key }}"
+  changed_when: false
```
rabi: Why changed_when false  here? We don't need to supress this unnecessarily.

roles/edpm_container_standalone/tasks/quadlet.yml
```diff
@@ -0,0 +1,84 @@
...
+
+- name: "Compute and inject config hash for {{ edpm_container_standalone_service }}"
+  become: true
+  container_config_hash:
+    staging_file: "{{ edpm_container_standalone_quadlet_staging_dir }}/edpm_{{ item.key }}.container"
+  loop: "{{ edpm_container_standalone_quadlet_defs | dict2items }}"
+  loop_control:
+    label: "{{ item.key }}"
```
rabi: We don't need to call this multiple times in a loop 
```
- name: "Compute and inject config hashes for all staged quadlet defs"
  container_config_hash:
    quadlet_staging_dir: "{{ edpm_container_standalone_quadlet_staging_dir }}"
```
```

File: roles/edpm_container_standalone/tasks/quadlet.yml [hunk 8]
```diff
@@ -0,0 +1,84 @@
...
+- name: Reload systemd daemon for quadlet generator
+  become: true
+  ansible.builtin.systemd:
+    daemon_reload: true
+  when: _edpm_quadlet_files is changed  # noqa: no-handler
+
+- name: "Ensure Quadlet container services are running for {{ edpm_container_standalone_service }}"
+  become: true
+  ansible.builtin.systemd:
+    name: "edpm_{{ item.item.key }}.service"
```
rabi: May be simplify this by identifying the changed containers...

```
- name: Identify changed containers
  ansible.builtin.set_fact:
    _edpm_changed_containers: >-
      {{ _edpm_quadlet_files.results
         | selectattr('changed')
         | map(attribute='item.key')
         | list }}
- name: Reload systemd daemon
  ansible.builtin.systemd:
    daemon_reload: true
  when: _edpm_changed_containers | length > 0
- name: Start or Restart changed containers
  ansible.builtin.systemd:
    name: "edpm_{{ item }}.service"
    state: restarted
  loop: "{{ _edpm_changed_containers }}"
```

plugins/modules/container_config_hash.py
oliashish: I think config is a directory under /var/lib/openstack which has ceph, telemetry, nova_nvme_cleaner configs.

roles/edpm_container_standalone/tasks/quadlet.yml
oliashish: This will render every time no matter the step pass or fail because of the EDPM_CONFIG_HASH line as that is not present in the template.

roles/edpm_container_standalone/tasks/quadlet.yml
oliashish: I need the staging file for a particular service at a time to calculate the hash not directory. `edpm_container_standalone_quadlet_staging_dir` container all the rendered files for all services.
If I use this, then probably I have to change the script to for this, and if I'm not wrong script does not have any information about which service is calling it, so that might be problem.

roles/edpm_container_standalone/tasks/quadlet.yml
rabi: Was this flagged by lint? As you're not capturing result it's not useful.

plugins/modules/container_config_hash.py
rabi: I don't think there is `configs`, config/ceph is there.

roles/edpm_container_standalone/tasks/quadlet.yml
rabi: yes, the ansible module can calculate the hash for all configs in the directory like we used to do before.

plugins/modules/container_config_hash.py
oliashish: yes, I have crc and cifmw both set up and I don't see configs. 
I'll remove it.

roles/edpm_container_standalone/tasks/quadlet.yml
rabi: to avoid it doing it for all services, we can probably pass containers: "{{ edpm_container_standalone_quadlet_defs.keys() | list }}"

roles/edpm_container_standalone/tasks/quadlet.yml
oliashish: removed it

roles/edpm_container_standalone/tasks/quadlet.yml
oliashish: Done

slagle (commented):
lgtm

rabi:
Can update the ovn PR to test with this as Depends-On? I suspect adoption jobs won't work there.

roles/edpm_container_standalone/molecule/quadlet/templates/test_service.container.j2
```diff
@@ -0,0 +1,24 @@
...
+[Container]
+ContainerName=test_service
+Image=quay.io/centos/centos:stream9
+Network=host
+User=root
+PodmanArgs=--privileged
+LogDriver=journald
+Exec=sleep 3600
+Environment=KOLLA_CONFIG_STRATEGY=COPY_ALWAYS
+{# DO NOT REMOVE - used by container_config_hash module #}
```
rabi: We should move this to the beginning of the file.

rabi (commented):
As expected adoption job failed in the ovn PR. Anyway we can merge this and merge the ovn change once we've branched for 19. Let's move the config hash comment to the beginning of the file. Otherise LGTM.

roles/edpm_container_standalone/molecule/quadlet/templates/test_service.container.j2
oliashish: ack

rabi:
recheck

rabi (commented):
/lgtm

molecule/test-helpers/verify_systemd_unit.yaml
```diff
@@ -21,10 +21,13 @@
...
       ignore_errors: true
     - name: Assert systemd unit is enabled {{ item.name }}
+      vars:
+        _allowed_states: >-
+          {{ item.enabled | default(["enabled", "generated"]) }}
       ansible.builtin.assert:
         that:
-          - unit_enabled.stdout == item.enabled | default("enabled")
-        fail_msg: "Systemd unit {{ item.name }} is not enabled"
+          - unit_enabled.stdout in (_allowed_states if _allowed_states is iterable and _allowed_states is not string else [_allowed_states])
```
slagle: in what cases would _allowed_states not be an iterable?

oliashish: Actually it is not, it is redundant. I'll fix that

oliashish:
recheck

plugins/modules/container_config_hash.py
```diff
@@ -86,12 +114,81 @@ def __init__(self, module, results):
...
+
+        # TODO(quadlet-migration): Remove _update_hashes() and its helpers
+        # (_find, _match_config_volumes, _update_container_config, _remove_file)
+        # once all services are migrated to Quadlet.
+        if staging_dir and containers:
+            self._update_hash_from_quadlet(staging_dir, containers)
+        else:
+            self._update_hashes()
 
         self.module.exit_json(**self.results)
```
slagle, line 128: claude is telling me that self.results['changed'] might not be set propertly, but that nothing relies on it. i'm fine with that, but it's something we could see about addressing later

slagle (approved):

PR #1176, opened 2026-05-15
Feat: Migrate OVN service to Quadlet

oliashish:
Migrated the OVN service from the old container_manage deployment to Quadlet by switching run.yml to use the standalone role's new quadlet path and adding a native .container template with TLS support, config hash integration, and healthcheck. Updated molecule tests to verify the quadlet file structure, kolla config, healthcheck script, and state file tracking.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/4038
Jira: [OSPRH-15124](https://redhat.atlassian.net/browse/OSPRH-15124)


roles/edpm_ovn/tasks/run.yml
```diff
@@ -42,7 +41,14 @@
...
+    edpm_container_standalone_quadlet_defs:
+      ovn_controller: "{{ _edpm_ovn_role_path }}/templates/quadlet/ovn_controller.container.j2"
     edpm_container_standalone_kolla_config_files:
       ovn_controller: "{{ lookup('template', 'kolla_config/kolla_ovn_controller.yaml.j2') | from_yaml }}"
+    edpm_container_standalone_config_volumes: >-
+      {{
+        edpm_ovn_controller_common_volumes +
+        (edpm_ovn_controller_tls_cacert_volumes | default([])) +
+        (edpm_ovn_controller_tls_volumes if edpm_ovn_tls_enabled | bool else []) +
+        (edpm_ovn_controller_healthcheck_volumes if edpm_ovn_healthcheck | bool else [])
```
oliashish: Do we want the healthcheck volumes to be part of hash calculation and detect change and restart service?

roles/edpm_ovn/tasks/run.yml
```diff
@@ -42,7 +41,14 @@
   vars:
     edpm_container_standalone_service: ovn_controller
     edpm_service_name: "{{ edpm_ovn_service_name }}"
-    edpm_container_standalone_container_defs:
-      ovn_controller: "{{ lookup('template', 'container_defs/ovn_controller.yaml.j2') | from_yaml }}"
```
slagle, line 46: could delete this file as well

slagle (commented):
this looks good.

roles/edpm_ovn/tasks/run.yml
```diff
@@ -42,7 +41,14 @@
...
+    edpm_container_standalone_quadlet_defs:
+      ovn_controller: "{{ _edpm_ovn_role_path }}/templates/quadlet/ovn_controller.container.j2"
     edpm_container_standalone_kolla_config_files:
       ovn_controller: "{{ lookup('template', 'kolla_config/kolla_ovn_controller.yaml.j2') | from_yaml }}"
+    edpm_container_standalone_config_volumes: >-
```
rabi: For services like nova that does not use ansible vars you'll end up with 2 sources of truth, one in *.container.j2 and other here. we should avoid it.

rabi:
why did zuul jobs did not run on this?

rabi:
recheck

roles/edpm_ovn/defaults/main.yml
```diff
@@ -106,6 +106,9 @@ edpm_ovn_ovs_other_config: {}
 # If container health check should be enabled
 edpm_ovn_healthcheck: true
 
+edpm_ovn_controller_healthcheck_volumes:
```
rabi: this is redundant I guess?

oliashish: Yes, this is. I'll remove it

roles/edpm_ovn/tasks/run.yml
oliashish: ack

oliashish:
recheck

oliashish:
recheck

oliashish:
recheck

oliashish:
recheck

oliashish:
recheck

oliashish:
recheck

roles/edpm_ovn/tasks/run.yml
```diff
@@ -14,6 +14,10 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Set OVN role path for quadlet templates
+  ansible.builtin.set_fact:
+    _edpm_ovn_role_path: "{{ role_path }}"
```
rabi, line 19: We don't need this task, we can use role_path directly for the quadlet template in L47.

roles/edpm_ovn/tasks/run.yml
```diff
@@ -27,22 +31,19 @@
...
 - name: Update ovn_controller health check script
   ansible.builtin.include_tasks:
     file: healthchecks.yml
 
+# TODO(quadlet-migration): Remove use_quadlet flag once all services are migrated.
 - name: Run ovn_controller container
   ansible.builtin.include_role:
     name: osp.edpm.edpm_container_standalone
   vars:
+    edpm_container_standalone_use_quadlet: true
```
rabi, line 43: We probably should not set it inline, we can do that in default/main.yml instead.

rabi (commented):
Two minor changes, otherwise LGTM.

roles/edpm_ovn/tasks/run.yml
oliashish: As discussed, let's go with this approach because of how role_path substitution works with include_role.

roles/edpm_ovn/tasks/run.yml
oliashish: Actually we cannot do it inline, because this is used in standalone role, and if put this in role's default/main.yaml it gets overridden by edpm_container_standalone/default/main.yaml variable which is default and is set to false. 

Also if I put this in default/main.yaml and pass it as variable, that does not make sense and is redundant as well.

rabi: For that you would need a separate variable. We may go without that, but in that case we should remove all old artefacts in the role,

oliashish:
recheck

slagle (approved):

PR #1181, opened 2026-05-22
Add edpm_instanceha_monitoring role for heartbeat sender

lmiccini:
New role to deploy a systemd timer on compute nodes that sends periodic UDP heartbeat packets to the instanceha pod. This enables dual-channel failure detection: when Nova marks a host down but heartbeats are still arriving, instanceha knows the host OS is alive and skips fencing (likely just a nova-compute process crash or partial network issue).

The role deploys a Python script, systemd service (oneshot), and systemd timer. Gated behind edpm_instanceha_monitoring_heartbeat_enabled (default: false) for opt-in activation.

The timer uses RandomizedDelaySec to add jitter and prevent thundering herd when many compute nodes boot simultaneously, that should also help spreading heartbeats uniformly across the interval.

Tested with:
~~~
  apiVersion: dataplane.openstack.org/v1beta1
  kind: OpenStackDataPlaneService
  metadata:
    name: instanceha-monitoring
  spec:
    playbook: osp.edpm.instanceha_monitoring
~~~
~~~
  spec:
    nodeTemplate:
      ansible:
        ansibleVars:
          edpm_instanceha_monitoring_heartbeat_enabled: true
          edpm_instanceha_monitoring_heartbeat_ip: "<instanceha-pod-IP>"
          edpm_instanceha_monitoring_heartbeat_port: 7411
          edpm_instanceha_monitoring_heartbeat_interval: 30
~~~

~~~
apiVersion: dataplane.openstack.org/v1beta1
kind: OpenStackDataPlaneDeployment
metadata:
  name: iha-hb
  namespace: openstack
spec:
  ansibleEEEnvConfigMapName: openstack-aee-default-env
  backoffLimit: 6
  deploymentRequeueTime: 15
  nodeSets:
  - openstack-edpm
  preserveJobs: true
  servicesOverride:
    - instanceha-monitoring
~~~

setting edpm_instanceha_monitoring_heartbeat_enabled: false and running the deployment removes the timer.

### DIFF
--- a/roles/edpm_instanceha_monitoring/tasks/configure.yml
+++ b/roles/edpm_instanceha_monitoring/tasks/configure.yml
@@ -1,0 +1,72 @@
+---
+# Copyright 2026 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Validate heartbeat IP is configured
+  ansible.builtin.assert:
+    that:
+      - edpm_instanceha_monitoring_heartbeat_ip | length > 0
+      - edpm_instanceha_monitoring_heartbeat_ip | regex_search('^[0-9a-fA-F.:]+$') is not none
+    fail_msg: >-
+      edpm_instanceha_monitoring_heartbeat_ip must be set to a valid IP address
+      when edpm_instanceha_monitoring_heartbeat_enabled is true.
+      Set it to the instanceha pod IP on the heartbeat network.
+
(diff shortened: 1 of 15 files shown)


roles/edpm_instanceha_monitoring/tasks/configure.yml
```diff
@@ -0,0 +1,72 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Validate heartbeat IP is configured
+  ansible.builtin.assert:
+    that:
+      - edpm_instanceha_monitoring_heartbeat_ip | length > 0
+      - edpm_instanceha_monitoring_heartbeat_ip | regex_search('^[0-9a-fA-F.:]+$') is not none
```
abays, line 21: This allows for IPv6, right?  But wouldn't IPv6 later cause an error at [1], since it's using `socket.AF_INET`?

[1] https://github.com/openstack-k8s-operators/edpm-ansible/pull/1181/changes#diff-327b11cc5533dad031b5691366f9ca29d32d2c62bd5ca5036e79c8c44681b08cR14

lmiccini: ouch thanks Andrew let me add ipv6 support

abays (approved):
/lgtm

PR #1184, opened 2026-06-01
Add HMAC-SHA256 support to heartbeat sender (HBV2)

lmiccini:
The heartbeat sender signs packets with HMAC-SHA256 and a timestamp (HBV2 protocol). The HMAC key is auto-detected from the standard EDPM config mount path
(/var/lib/openstack/configs/instanceha-monitoring/hmac-key), mounted automatically by the OpenStackDataPlaneService dataSource referencing the instanceha-heartbeat-hmac K8s Secret.

The key is deployed to /var/lib/openstack/instanceha-monitoring/hmac-key on the compute node.

Related: https://github.com/openstack-k8s-operators/infra-operator/pull/591


lmiccini:
~~~
apiVersion: dataplane.openstack.org/v1beta1
kind: OpenStackDataPlaneService
metadata:
  name: instanceha-monitoring
spec:
  dataSources:
    - secretRef:
        name: instanceha-heartbeat-hmac
        optional: true
  playbook: osp.edpm.instanceha_monitoring
  edpmServiceType: instanceha-monitoring
~~~

lmiccini:
recheck

lmiccini:
recheck

dciabrin (approved):

lmiccini:
@abays would you mind having a look please? infra-operator side is merged. Thank you

lmiccini:
Recheck

roles/edpm_instanceha_monitoring/molecule/default/test-data/hmac-key
```diff
@@ -0,0 +1 @@
+test-hmac-key-for-molecule
```
abays: Shouldn't this be a hex value since it is consumed by [1]?  I think it would fail otherwise.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/pull/1184/changes#diff-327b11cc5533dad031b5691366f9ca29d32d2c62bd5ca5036e79c8c44681b08cR18

roles/edpm_instanceha_monitoring/tasks/configure.yml
```diff
@@ -40,6 +40,32 @@
     fail_msg: >-
       edpm_instanceha_monitoring_heartbeat_interval must be a positive integer.
 
+- name: Read HMAC key from EDPM-mounted config
+  delegate_to: localhost
+  ansible.builtin.slurp:
+    src: "{{ edpm_instanceha_monitoring_config_src }}/hmac-key"
+  register: _hmac_key_file
+  no_log: true
```
abays, line 48: The associated `OpenStackDataPlaneService` has this secret as `optional: true`, right?  Should we add some sort of guard here to handle the scenario where it was not provided?

lmiccini: I think I will send a pr to openstack-operator with an example that has optional: false (or no optional at all). The secret is always created automatically, and will document that users should  create a copy of the service definition if the instanceha CR name is different than the default one.

abays (approved):
/lgtm

lmiccini:
recheck

lmiccini:
Recheck

PR #1185, opened 2026-06-04
Split frr configuration to bgp and bgp+evpn

mtomaska:
Neutron is introducing a new frr driver which will configure frr dynamically at runtime.

This patch introduces new variable called `edpm_frr_evpn_mode` which when set will configure frr with minum configuration. Rest will be handled at runtime by the frr driver

JIRA: [OSPRH-29356](https://redhat.atlassian.net/browse/OSPRH-29356)

### DIFF
--- a/roles/edpm_frr/defaults/main.yml
+++ b/roles/edpm_frr/defaults/main.yml
@@ -19,6 +19,7 @@
 
 # All variables within this role should have a prefix of "edpm_frr"
 
+edpm_enable_evpn: false
 edpm_frr_bfd: true
 # Configures the detection multiplier to determine packet loss.
 # The remote transmission interval will be multiplied by this
--- a/roles/edpm_frr/meta/argument_specs.yml
+++ b/roles/edpm_frr/meta/argument_specs.yml
@@ -146,6 +146,12 @@ argument_specs:
         default: traditional
         description: ''
         type: str
+      edpm_enable_evpn:
+        default: false
+        description: |
+          When true, load the vrf kernel module and always render the l2vpn evpn
+          address-family in frr.conf.j2.
+        type: bool
       edpm_frr_hostname:
         default: '{{ ansible_facts["hostname"] | default("")}}'
         description: Dynamically retrieved from ansible facts.
--- a/roles/edpm_frr/molecule/evpn_mode/verify.yml
+++ b/roles/edpm_frr/molecule/evpn_mode/verify.yml
@@ -1,0 +1,48 @@
+---
+# Copyright 2026 Red Hat, LLC
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- import_playbook: ../default/verify.yml
+
+- name: Verify EVPN mode frr.conf
+  gather_facts: false
+  hosts: compute-1
+  vars:
+    frr_conf_path: /var/lib/openstack/frr/etc/frr/frr.conf
+  tasks:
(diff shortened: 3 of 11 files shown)


mtomaska:
recheck

roles/edpm_frr/molecule/evpn_mode/verify.yml
```diff
@@ -0,0 +1,53 @@
...
+      ansible.builtin.command: grep -Fq '{{ item }}' {{ frr_conf_path }}
+      changed_when: false
+      loop:
+        - 'frr version'
+        - 'line vty'
+
+    - name: Verify vrf module will be loaded at boot
+      become: true
+      ansible.builtin.command: grep -q ^vrf$ /etc/modules-load.d/vrf.conf
+      changed_when: false
```
eduolivares, line 43: This is causing the Zuul failure. The `edpm_module_load` role templates the file via `module-load.conf.j2` which renders as `vrf ` (with a trailing space, because `{{ item.params | default('') }}` produces an empty string after the space). The `^vrf$` pattern does not match `vrf `, so the grep fails.

A fix like `grep -q '^vrf'` (drop the `$` anchor) or `grep -qP '^vrf\s*$'` would work.

https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/rdoproject.org/build/ce921e270a3f4e2fa37d7b928bec9f6b

roles/edpm_frr/tasks/configure_bgp.yml
```diff
@@ -0,0 +1,60 @@
+---
+# Copyright 2023 Red Hat, Inc.
```
eduolivares: nit: 2026

mtomaska:
recheck

mtomaska:
recheck

roles/edpm_frr/defaults/main.yml
```diff
@@ -19,6 +19,10 @@
 
 # All variables within this role should have a prefix of "edpm_frr"
 
+# Also used in edpm_neutron_ovn role
+edpm_frr_evpn_mode: false
+edpm_frr_evpn_local_ip: ''
```
karelyatin: edpm_frr_evpn_local_ip and edpm_frr_evpn_vxlan_ports can be cleaned up from edpm_frr role as unused

roles/edpm_frr/meta/argument_specs.yml
```diff
@@ -146,6 +146,20 @@ argument_specs:
         default: traditional
         description: ''
         type: str
+      edpm_frr_evpn_mode:
+        default: false
+        description: |
+          When true, generate a minimal frr.conf with only global settings. Frr configuration will be managed by the Neutron
+          frr driver. Also used in the edpm_neutron_ovn role.
+        type: bool
+      edpm_frr_evpn_local_ip:
```
karelyatin: same here

roles/edpm_frr/defaults/main.yml
eduolivares: I would then rename them as `edpm_neutron_ovn_evpn_local_ip` and `edpm_neutron_ovn_evpn_vxlan_ports`

eduolivares (approved):

roles/edpm_neutron_ovn/defaults/main.yml
```diff
@@ -70,3 +70,9 @@ edpm_neutron_ovn_agent_bgp_peer_bridges: ''
 # BGP interconnect bridge for the ovn-bgp extension.
 # Set this variable when the ovn-agent is configured with the ovn-bgp extension.
 edpm_neutron_ovn_agent_bgp_interconnect_bridge: "{{ neutron_physical_bridge_name | default('br-ex') }}"
+
+# EVPN mode variables
+# edpm_frr_evpn_mode is also used in edpm_frr role
+edpm_frr_evpn_mode: false
+edpm_neutron_ovn_evpn_local_ip: ''
```
slawqo: I am not really an ansible expert that much so I may be wrong but for me having the same variable defined in 2 roles is not good practice. And IIUC it is used only to configure some external_ids in the OVS on the host. Can't we move those settings of external_ids to the `edpm_frr` role and then don't include FRR related stuff into this role at all?

roles/edpm_frr/templates/config/frr.conf.evpn.j2
cubeek, line 1: Should it contain at least the BGP router for the control plane?

mtomaska:
recheck

PR #1186, opened 2026-06-08
Delete Release Branch Sync workflow

stuggi:
Starting with FR5 we no long rely on the specific release sync branch. Lets remove the job.

Jira: [OSPRH-25747](https://redhat.atlassian.net/browse/OSPRH-25747)


abays (approved):
/lgtm

PR #1196, opened 2026-06-17
Feat: Migrate edpm_telemetry to Quadlet

oliashish:
Jira: [OSPRH-27890](https://redhat.atlassian.net/browse/OSPRH-27890)
Depeds-on: #1175


oliashish:
recheck

oliashish:
recheck

slagle (commented):
LGTM. I'd like to see a review from telemetry before merging.

oliashish:
> LGTM. I'd like to see a review from telemetry before merging.

Yes, waiting for CI to pass, before tagging anyone

slagle:
recheck

time out updating the controlplane, probably not related to the PR

`
2026-07-23 17:55:31.115433 | controller |     - + oc wait openstackversion.core.openstack.org/controlplane --for=condition=MinorUpdateDataplane
2026-07-23 17:55:31.115441 | controller |       --timeout=1200s
2026-07-23 17:55:31.115450 | controller |     - 'error: timed out waiting for the condition on openstackversions/controlplane'
2026-07-23 17:55:31.115458 | controller |     - 'make: *** [Makefile:977: openstack_update_run] Error 1'
`

roles/edpm_telemetry/molecule/default/verify.yml
```diff
@@ -54,3 +58,111 @@
...
+        that:
+          - "'Volume={{ edpm_telemetry_config_dest }}:/var/lib/kolla/config_files/src:z' in _container_content"
+          - "'Volume=/var/lib/kolla/config_files/ceilometer_agent_compute.json:/var/lib/kolla/config_files/config.json:z' in _container_content"
+          - "'Volume=/run/libvirt:/run/libvirt:shared,ro' in _container_content"
+          - "'Volume=/etc/hosts:/etc/hosts:ro' in _container_content"
+          - "'Volume=/etc/pki/tls/certs/ca-bundle.trust.crt:/etc/pki/tls/certs/ca-bundle.trust.crt:ro' in _container_content"
+          - "'Volume=/etc/localtime:/etc/localtime:ro' in _container_content"
+          - "'Volume=/etc/pki/ca-trust/source/anchors:/etc/pki/ca-trust/source/anchors:ro' in _container_content"
+          - "'Volume=/dev/log:/dev/log' in _container_content"
+          - "'Volume={{ edpm_telemetry_certs }}:/etc/ceilometer/tls:z' in _container_content"
```
vyzigold, line 97: Looking at the templates/quadlet/ceilometer_agent_compute.container.j2. This volume will be added conditionally only when `tls_cert_exists=true`. So what happens here when TLS is disabled on that edpm node? I think this task will always fail.

vyzigold: ditto for the ca-bundle volume, which is also mounted conditionally, although I'm not currently sure in which case it wouldn't be present.

vyzigold: Now I see this is a molecule test, I thought this will run in the actual role. Ignore the comments :D

vyzigold:
Looks good to me as well, although I have a very limited knowledge regarding quadlets. I deployed the code though and I can see all four exporter containers are running and we're getting metrics from all of them as expected.

danpawlik:
@oliashish I guess the CI job edpm-ansible-baremetal-minor-update  is failing because you don't cleanup old service before migration. 
```
Update path looks broken: this PR writes Quadlet units under /etc/containers/systemd/edpm_*.container but never removes the old /etc/systemd/system/edpm_*.service units from container_manage. Service names are unchanged (edpm_podman_exporter.service, etc.), so on update the old unit still wins and systemctl restart fails (seen in CI for podman/node/network/ceilometer exporters).
```

rabi:
> @oliashish I guess the CI job edpm-ansible-baremetal-minor-update is failing because you don't cleanup old service before migration.
> 
> ```
> Update path looks broken: this PR writes Quadlet units under /etc/containers/systemd/edpm_*.container but never removes the old /etc/systemd/system/edpm_*.service units from container_manage. Service names are unchanged (edpm_podman_exporter.service, etc.), so on update the old unit still wins and systemctl restart fails (seen in CI for podman/node/network/ceilometer exporters).
> ```

Yeah, these legacy units are planned to be cleaned up by the upgrade tasks, but that cleanup is not currently part of the minor update flow as this change is meant for OSP19. For now, we can probably add the cleanup explicitly to the minor update path to make the CI happy.

The issue is that Quadlet-generated units under /run/systemd/generator have lower precedence than native units in /etc/systemd/system, so the old legacy edpm_*.service files can still win during minor update. We did not hit this in the OVN PR because there were no config-file path changes there, so starting the old service did not break the job.
[systemd.unit load path docs](https://www.freedesktop.org/software/systemd/man/249/systemd.unit.html#Unit%20File%20Load%20Path)

roles/edpm_telemetry/tasks/update.yml
```diff
@@ -1,4 +1,18 @@
...
+  tags:
+    - upgrade # This tag is for reference, to be removed later
+    - update 
+    - telemetry
+  become: true
+  ansible.builtin.file:
+    path: "/etc/systemd/system/edpm_{{ item }}.service"
+    state: absent
+  loop: "{{ edpm_telemetry_enabled_exporters }}"
+
```
slagle, line 15: do you think a daemon-reload is needed after this?

oliashish: No, daemon reload happens in edpm_container_standalone role once the new Quadlet generated .service unit file is in place. Otherwise we'll have no service container for a brief period until the reload.

oliashish:
@danpawlik @vyzigold 
The CI is passing except adoption job which failed with mariadb password issue (unrelated to our changes), I'm re-running it now. Whenever you can, please have another look. 

@danpawlik I have addressed you comment with new changes. 

Thank you.

slagle:
recheck

roles/edpm_telemetry/tasks/update.yml
rabi, line 94: Though not related to this PR and an  exsiting bug, this condition is probably checking for the wrong file and hence running all the time.

rabi:
recheck

roles/edpm_telemetry/tasks/update.yml
oliashish: Ack!
I'm unaware of what it does.
Thanks

slagle (approved):

PR #1197, opened 2026-06-17
Feat: Migrate telemetry_power_monitoring to Quadlet

oliashish:
Jira: [OSPRH-27891](https://redhat.atlassian.net/browse/OSPRH-27891)
Depends-on: #1175


oliashish:
recheck

vyzigold:
This will need an update, because Kepler was completely removed from RHOSO in https://github.com/openstack-k8s-operators/edpm-ansible/pull/1209

oliashish:
> This will need an update, because Kepler was completely removed from RHOSO in #1209

Thanks for letting us know. I'll update it

oliashish:
recheck

vyzigold:
I tried to test the PR in my environment. Unfortunately, in this case I can't trigger any metric to appear, because the ceilometer_ipmi_agent doesn't expose any metrics when running on a VM, which is what install_yamls will use for computes by default. But the container is running, logs look clean and ceilometer seems to be listening on the expected port, so from telemetry side this lgtm.

slagle (approved):

PR #1202, opened 2026-06-25
EDPM nmstate tool support

Jaganathancse:
This change is to provide EDPM deployment based on nmstate tool without providing override variables and avoid failures if nmstate tool is used.


slagle (commented):
the var can also be removed from roles/edpm_network_config/molecule/nmstate/converge.yml

slagle:
recheck

slagle (approved):

PR #1210, opened 2026-07-02
edpm_telemetry: add node_exporter_extra_args var

bogdando:
Introduce edpm_telemetry_node_exporter_extra_args (default: []) and wire it into the node_exporter container definition template, allowing operators to pass extra flags to node_exporter on a per-nodeset basis.

This may be needed to disable collectors that fail on hardware not fully supported by node_exporter (e.g. on hosts with NVLink CX7 InfiniBand adapters that reject ETHTOOL and IB counter reads with EINVAL).

Resolves: [OSPRH-32325](https://redhat.atlassian.net/browse/OSPRH-32325)
Generated-by: claude-4.6-opus-high

### DIFF
--- a/roles/edpm_telemetry/defaults/main.yml
+++ b/roles/edpm_telemetry/defaults/main.yml
@@ -75,3 +75,7 @@ edpm_telemetry_openstack_network_exporter_collectors:
 # Supported levels are: debug info notice warning error critical
 # Default: info
 edpm_telemetry_openstack_network_exporter_log_level: info
+# Extra CLI flags passed verbatim to node_exporter (list of strings).
+# Use to disable collectors that generate spurious errors on hardware that
+# does not support them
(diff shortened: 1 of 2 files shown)


jlarriba:
Hi @bogdando I did not realized that you had already launched this PR, so I created a twin one here: https://github.com/openstack-k8s-operators/edpm-ansible/pull/1223 to resolve this very same issue.

It is not exactly a duplicate because yours decided to leave the current "opinionated" disabled collectors harcoded and just add the arguments at the end, while mine moves the current config options to the variable as default value so the user has a say on the opinionated disabled collectors and he can choose to enable some or all of them.

Would you mind to close this one?

jlarriba (approved):

jlarriba:
Sorry that was a mistake, was not intended to approve it due to my unaddressed comment

PR #1212, opened 2026-07-06
Make /run/frr volume mount conditional on ovn-evpn extension

eduolivares:
The /run/frr bind mount was unconditionally added to the ovn_agent container volumes, creating a silent dependency on the edpm_frr role even when the extension is enabled (which is the only extension that requires to access frr from ovn-agent).

Move the /run/frr mount into a new edpm_neutron_ovn_frr_volumes variable that is only populated when the ovn-evpn extension is present in edpm_neutron_ovn_agent_agent_extensions. Add a molecule verification to the evpn_disabled scenario confirming the mount is absent.

Assisted-By: claude-opus-4.6


eduolivares:
recheck cifmw-crc-podified-edpm-baremetal-bootc

eduolivares:
Closing this PR because https://github.com/openstack-k8s-operators/edpm-ansible/pull/1213 was already merged instead

PR #1213, opened 2026-07-06
Improve evpn extension configuration

mtomaska:
Improve how evpn configuration and volume mount are handled when `ovn-evpn` extension is set. The change partially mimics how `edpm_neutron_ovn_tls_enabled` configures when TLS is enabled.
In this change.

`edpm_neutron_ovn_evpn_extension_enabled` gets set to `True` if the `ovn-evpn` extension is specified in the extensions list. Then:
-  Sets extra volumes in `ovn_agent.yaml.j2`
-  Sets related configuration in `neutron-ovn-agent.conf.j2`

### DIFF
--- a/roles/edpm_neutron_ovn/defaults/main.yml
+++ b/roles/edpm_neutron_ovn/defaults/main.yml
@@ -24,7 +24,9 @@ edpm_neutron_ovn_common_volumes:
   - "{{ edpm_neutron_ovn_agent_lib_dir }}:/var/lib/neutron:shared,z"
   - "{{ edpm_neutron_ovn_agent_lib_dir }}/ovn_agent_haproxy_wrapper:/usr/local/bin/haproxy:ro"
   - "{{ edpm_neutron_ovn_agent_lib_dir }}/kill_scripts:/etc/neutron/kill_scripts:ro"
-  - /run/frr:/run/frr:shared,z
+
+edpm_neutron_ovn_frr_volumes:
+  - "/run/frr:{{ edpm_neutron_ovn_agent_frr_vty_socket }}:shared,z"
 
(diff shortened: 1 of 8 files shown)


roles/edpm_neutron_ovn/defaults/main.yml
```diff
@@ -24,7 +24,9 @@ edpm_neutron_ovn_common_volumes:
   - "{{ edpm_neutron_ovn_agent_lib_dir }}:/var/lib/neutron:shared,z"
   - "{{ edpm_neutron_ovn_agent_lib_dir }}/ovn_agent_haproxy_wrapper:/usr/local/bin/haproxy:ro"
   - "{{ edpm_neutron_ovn_agent_lib_dir }}/kill_scripts:/etc/neutron/kill_scripts:ro"
-  - /run/frr:/run/frr:shared,z
+
+edpm_neutron_ovn_frr_volumes:
```
eduolivares, line 28: should we add `edpm_neutron_ovn_frr_volumes` to `roles/edpm_neutron_ovn/meta/argument_specs.yml` file too?

eduolivares: Maybe not. At least I can't find `edpm_neutron_ovn_tls_volumes` on the argument_spec file.

mtomaska: Yep, I had the same question and decided to leave it out because I dont see a reason why a caller should modify it. A user will just set `edpm_neutron_ovn_agent_frr_vty_socket` and the code will take care of rest.

eduolivares (approved):

karelyatin (approved):

PR #1215, opened 2026-07-07
Make BGP graceful-shutdown configurable for eBGP support

m-blue-m:
```
Make BGP graceful-shutdown configurable for eBGP support
BGP graceful-shutdown (RFC 8326) was hardcoded in the FRR config
template. This prevents eBGP deployments from disabling it, where
graceful-shutdown adds the GRACEFUL_SHUTDOWN community to all
advertised routes causing peers to deprioritize them.

Adds edpm_frr_bgp_graceful_shutdown variable (default: true) to
allow eBGP nodes to opt out. Existing deployments are unaffected.

Related: OSPRH-28085
Assisted-By: Claude Code
Signed-off-by: Maor Blaustein <mblue@redhat.com>
```

### DIFF
--- a/roles/edpm_frr/defaults/main.yml
+++ b/roles/edpm_frr/defaults/main.yml
@@ -65,6 +65,7 @@ edpm_frr_watchfrr_options: ''
 edpm_frr_zebra: true
 edpm_frr_zebra_nht_resolve_via_default: true
 edpm_frr_zebra_graceful_restart_time: ''
+edpm_frr_bgp_graceful_shutdown: true
 edpm_frr_conf_custom_globals: ''
 edpm_frr_conf_custom_router_bgp: ''
 edpm_frr_conf_custom_ipv4: ''
--- a/roles/edpm_frr/meta/argument_specs.yml
+++ b/roles/edpm_frr/meta/argument_specs.yml
@@ -213,6 +213,14 @@ argument_specs:
         default: true
         description: ''
         type: bool
+      edpm_frr_bgp_graceful_shutdown:
+        default: true
+        description: >-
+          Enable BGP graceful-shutdown (RFC 8326), adding the
+          GRACEFUL_SHUTDOWN community to advertised routes. Set to
+          false for eBGP nodes where graceful-shutdown should not
+          apply during normal operation.
+        type: bool
       edpm_frr_zebra_graceful_restart_time:
         default: ''
         description: Zebra graceful_restart time (seconds)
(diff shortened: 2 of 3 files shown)


roles/edpm_frr/defaults/main.yml
```diff
@@ -65,6 +65,7 @@ edpm_frr_watchfrr_options: ''
 edpm_frr_zebra: true
 edpm_frr_zebra_nht_resolve_via_default: true
 edpm_frr_zebra_graceful_restart_time: ''
+edpm_frr_bgp_graceful_shutdown: true
```
mtomaska, line 68: I would be for defaulting it to `False`. We have also found it not useful for evpn and I keep deleting it from the frr.conf :). Hence, only enable it if you have specific use case for it, otherwise keep it off.

mtomaska (changes requested):

roles/edpm_frr/defaults/main.yml
m-blue-m: Thank you for the review :)

The current template has bgp graceful-shutdown hardcoded unconditionally for all EDPM nodes.
Defaulting to true preserves this existing behavior for all current BGP jobs (dt01, kappa, etc) without any functional change.
Defaulting to false would remove graceful-shutdown from all existing deployments which could be a disruptive change.
If there is consensus to disable it by default (including for EVPN), happy to change the default to false, but that should be a separate discussion/change since it affects all BGP jobs.
This current change meant for eBGP testing in CI (actually allows dt02 specific racks opting out of the configuration `on` default everywhere, the current behaviour).

mtomaska: Good point. In this case, leave it as `True`. We can slowly rollout overwriting it to `False` and if it does not cause any issues then we can switch it to `False`,
Frr [documentation](https://docs.frrouting.org/en/latest/bgp.html) states
`... Graceful BGP Session Shutdown to reduce the amount of lost traffic when taking BGP sessions down for maintenance. The use of the community needs to be supported from your peers side to actually have any effect...`
I am not even sure if we are taking advantage of that in our CI or not :)

mtomaska (approved):
/lgtm

eduolivares:
recheck
ci-fmw changes that fix tls issues were merged today

m-blue-m:
recheck cifmw-crc-podified-edpm-baremetal-bootc (RETRY_LIMIT)

m-blue-m:
recheck cifmw-crc-podified-edpm-baremetal (RETRY_LIMIT) already passed with same commit

m-blue-m:
recheck flaky gate failures

m-blue-m:
recheck flaky gate failures

eduolivares (approved):

karelyatin (approved):
/lgtm

PR #1234, opened 2026-07-27
Switch container build from EL9 to EL10

holser:
Update base images to UBI10, default image tag to latest-el10, and molecule delorean repo to centos10-master as prerequisite to RHOSO19

### DIFF
--- a/molecule/common/Containerfile.j2
+++ b/molecule/common/Containerfile.j2
@@ -5,7 +5,7 @@ FROM {{ item.registry.url }}/{{ item.image }}
 FROM {{ item.image }}
 {% endif %}
 
-RUN curl -o /etc/yum.repos.d/delorean.repo https://trunk.rdoproject.org/centos9-master/current/delorean.repo && \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,6 +1,6 @@
 TEST_ENV_REPOSITORY := https://github.com/openstack-k8s-operators/ci-framework.git
 IMAGE_TAG_BASE ?= quay.io/openstack-k8s-operators/openstack-ansibleee-runner
-IMAGE_TAG ?= latest
+IMAGE_TAG ?= latest-el10
```
slagle: What about keeping this as latest here for el10?
Then in the 18-stable branch we switch that to latest-el9? 
Does CI also need updating to use the right image depending on which branch of edpm-ansible is being tested, or does it already override with custom tags?
@rabi WDYT?

rabi: Yeah, we should leave it as `latest` though it does not matter. CI already uses branch specific tags ex. `18.0-fr6-lastest` and we push them accordingly from the github actions from branches.

holser: @slagle @rabi I changed it back to latest. Thank you very much!

slagle (approved):

rabi:
recheck

PR #1241, opened 2026-07-29
Fix: migrate ovn to quadlet minor update

oliashish:
Since we want to merge changes to Quadlet and CI works for the ovn change, but that is still running legacy container/systemd services. It was not caught in the CI because there is no config change for ovn_controller and previous .service unit under /etc/systemd/system/edpm_ovn_controller.service win in precedence over quadlet generated .service file under /run/systemd/generator/edpm_ovn_controller.service and restarted again without any issue.

These changes will make sure the new quadlet container are running and not the legacy container


slagle (approved):

oliashish:
recheck

slagle (approved):

PR #1242, opened 2026-07-29
Add infra.leapp collection to ansibleee requirements

holser:
infra.leapp collection is a collection for RHEL Upgrade. This patch adds infra.leapp collection instead of writing own role from scratch

Jira: [OSPRH-31558](https://redhat.atlassian.net/browse/OSPRH-31558)

[1] https://github.com/redhat-cop/infra.leapp


holser:
recheck

slagle (approved):

PR #1239, opened 2026-07-29
Bind telemetry exporters to ctlplane IP instead of host network

vyzigold:
Replace Network=host with explicit PublishPort directives bound to the control plane IP address in all telemetry exporter quadlets:

  - ceilometer_agent_compute: {{ ctlplane_ip }}:9101:9101
  - node_exporter:            {{ ctlplane_ip }}:9100:9100
  - podman_exporter:          {{ ctlplane_ip }}:9882:9882
  - openstack_network_exporter: {{ ctlplane_ip }}:9105:9105

Using the new PublishPort directives ensures the exporters serve metrics only on the control plane network.

The molecule test inventory is updated to supply a ctlplane_ip value so the quadlet templates render correctly during CI.

Generated-by: Claude-Code claude-opus-4-6

PR #1243, opened 2026-07-31
Bind ceilometer_ipmi to ctlplane IP instead of host network

vyzigold:
Replace Network=host with explicit PublishPort directive bound to the control plane IP for ceilometer ipmi agent.

Using the new PublishPort directive ensures the exporter serves metrics only on the control plane network.

The molecule test inventory is updated to supply a ctlplane_ip on localhost so that the quadlet template renders correctly and we have an IP that actually works with the PublishPort in CI.

### DIFF
--- a/roles/edpm_telemetry_power_monitoring/molecule/default/molecule.yml
+++ b/roles/edpm_telemetry_power_monitoring/molecule/default/molecule.yml
@@ -36,7 +36,7 @@ provisioner:
       all:
         hosts:
           compute-1:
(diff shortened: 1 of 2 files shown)


vyzigold:
This is a continuation of https://github.com/openstack-k8s-operators/edpm-ansible/pull/1239 now that https://github.com/openstack-k8s-operators/edpm-ansible/pull/1197 merged

slagle (approved):

vyzigold:
recheck

PR #1244, opened 2026-08-12
Fix edpm_bootstrap failed_when logic that silently ignores failures

slagle:
The install task's failed_when used AND-ed conditions that prevented
failure detection on CentOS Stream 9 and ignored return codes other
than 1. The rpm verify task checked stderr for a string that rpm -V
never outputs. Both bugs allowed EDPM nodes to silently end up in a
broken state during deployment.

Jira: [OSPRH-33601](https://redhat.atlassian.net/browse/OSPRH-33601)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/roles/edpm_bootstrap/tasks/packages.yml
+++ b/roles/edpm_bootstrap/tasks/packages.yml
@@ -46,9 +46,8 @@
     name: "{{ edpm_bootstrap_packages_bootstrap }}"
     state: present
   failed_when:
-    - (ansible_facts['distribution'] | lower) == 'redhat'
     - not ansible_check_mode | bool
-    - (edpm_bootstrap_packages_bootstrap_result.rc | int) == 1
+    - edpm_bootstrap_packages_bootstrap_result.rc != 0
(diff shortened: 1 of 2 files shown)


slagle:
recheck

slagle:
recheck

roles/edpm_bootstrap/tasks/packages.yml
```diff
@@ -59,6 +58,6 @@
   ansible.builtin.command: "rpm -V {{ edpm_bootstrap_packages_bootstrap | join(' ') }}"  # noqa: command-instead-of-module
```
rabi: Probably need to use `rpm -V --noconfig` as the files (e.g. NetworkManager.conf, nftables.conf) are modified before adoption and this check fails in the adoption job.

slagle: Indeed, I missed that on the first run. Thanks!

slagle:
recheck

rabi (approved):

rabi:
recheck

PR #1248, opened 2026-08-20
Remove kolla dependency from neutron ovn_agent

rabi:
Run neutron-ovn-agent directly via the Quadlet Exec= key and replace the kolla_start config staging with an init container plus direct bind mounts.

jira: [OSPRH-34417](https://redhat.atlassian.net/browse/OSPRH-34417)


roles/edpm_neutron_ovn/templates/quadlet/ovn_agent.container.j2
```diff
@@ -30,7 +30,7 @@ User=root
 PodmanArgs=--privileged --log-driver=journald
 PodmanArgs=--pid=host
 PodmanArgs=--cgroupns=host
-Environment=KOLLA_CONFIG_STRATEGY=COPY_ALWAYS
+Exec=/usr/bin/neutron-ovn-agent
```
slagle, line 33: we could combine the chown here, and not have to use an init container. 

`
Exec=/bin/bash -c "/usr/bin/chown -R neutron:neutron /var/lib/neutron && /usr/bin/neutron-ovn-agent"
`

WDYT?

rabi: Yeah, ovn_agent runs as root/privileged so we could inline it. I've kept the dedicated init container to match all other services (nova has used this pattern since pre-quadlet). The main ovn_agent container is Restart=always and inlining would re-run a recursive chown of /var/lib/neutron on every restart. Keeping it in a oneshot init also means if we later drop root from the ovn_agent container or need more prep steps, they live in one place.

Also, the reason for using init container for chown across services:   By chowning by name inside a container from that image, edpm-ansible doesn't duplicate that magic number or require a host-side neutron user with multiple sources of truth as done earlier in https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_cyborg/tasks/configure.yml#L113-L123

slagle (approved):

PR #1251, opened 2026-08-21
Remove kolla dependency from ceilometer_agent_compute

rabi:
Run ceilometer-polling directly via the Quadlet Exec= key instead of the kolla_start entrypoint.

jira: [OSPRH-34422](https://redhat.atlassian.net/browse/OSPRH-34422)


rabi:
recheck

rabi:
recheck

slagle (approved):

PR #1254, opened 2026-08-24
Remove kolla dependency from cyborg_agent

rabi:
Run the cyborg-agent directly via the Quadlet Exec= key and replace kolla_start config staging with direct bind mounts. *.conf files are discovered in the config dest and mounted into
/etc/cyborg/cyborg.conf.d/ at their final paths; an init container chowns /var/lib/cyborg for the cyborg user before the agent starts.


rabi:
recheck

roles/edpm_cyborg/tasks/configure.yml
```diff
@@ -132,3 +121,34 @@
...
+        | map(attribute='path')
+        | sort
+        | map('basename')
+        | map('regex_replace', '^(.*)$', '/etc/cyborg/cyborg.conf.d/\\1:ro'))
+      | map('join', ':')
+      | list
+    }}"
+  tags:
+    - configure
+    - cyborg
```
slagle, line 154: one thing about this pattern is that now configure and install can't be used separately in different ansible executions since the registered var won't be available to install.yml. That may have already been the case anyway, I don't know.

The quadlet rendering could be moved to configure.yml, and install.yml (which should probably be run.yml) should just start it IMO.

slagle (approved):

PR #1257, opened 2026-08-26
Remove kolla dependency from neutron sriov_agent

rabi:
Run neutron-sriov-nic-agent directly via the Quadlet Exec= key and replace the kolla_start config staging with an init container plus direct bind mounts.


rabi:
recheck

rabi:
recheck

slagle (approved):

PR #1258, opened 2026-08-26
Remove kolla dependency from frr

rabi:
Run the FRR daemons directly via the Quadlet Exec= key using frrinit.sh and replace kolla_start config staging with direct bind mounts of frr.conf and daemons at their final paths. An init container chowns /run/frr for the frr user before the main container starts.


rabi:
recheck

slagle (approved):

PR #1269, opened 2026-09-07
Fix os-net-config reapplying issue in adoption scenario

Jaganathancse:
This change is to avoid reapplying os-net-config issue in adoption scenario and also ensure only reapplied when using 'edpm_network_config_update' varibale is 'true'.


roles/edpm_network_config/tasks/os_net_config_tool.yml
```diff
@@ -32,13 +32,31 @@
       register: os_net_config_returncode_slurp
       when: os_net_config_returncode_stat.stat.exists
 
+    - name: Check tripleo os-net-config return code file exists
```
rabi: I think we should copy the return code files from tripleo-ansible to edpm-ansible in pre-adoption (may be in edpm-tripleo-cleanup role) if that's what we want, rather than messing up this role with tripleo specific logic.

rabi (changes requested):

roles/edpm_network_config/tasks/os_net_config_tool.yml
Jaganathancse: @rabi thanks for sharing suggestion, i would like to keep copying logic '/var/lib/tripleo-config/os-net-config.returncode' to '/var/lib/edpm-config/os-net-config.returncode' if triple return code  part of this role only . fine?
@karthiksundaravel ^

rabi: No, not in this role. Add it to edpm_tripleo_cleanup role or in some pre adoption tasks

Jaganathancse: @rabi when 'edpm_tripleo_cleanup' this role is executed in adoption nodeset CR? before configure-network?

rabi: Yes. 

https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/dataplane_adoption/tasks/main.yaml#L774

https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/dataplane_adoption/tasks/main.yaml#L861

Jaganathancse:
recheck

roles/edpm_tripleo_cleanup/tasks/main.yml
```diff
@@ -94,3 +94,30 @@
...
+    - name: Ensure /var/lib/edpm-config directory exists
+      ansible.builtin.file:
+        path: /var/lib/edpm-config
+        state: directory
+        owner: "{{ ansible_user | default(ansible_user_id) }}"
+        group: "{{ ansible_user | default(ansible_user_id) }}"
+        setype: container_file_t
+        selevel: s0
+        mode: "0755"
+    - name: Ensure /var/lib/edpm-config/os-net-config.returncode exist
```
karthiksundaravel, line 128: Should we add a check that the copy shall be done only when  /var/lib/edpm-config/os-net-config.returncode doesn't exist ?

Jaganathancse: > Should we add a check that the copy shall be done only when /var/lib/edpm-config/os-net-config.returncode doesn't exist ?

updated

Jaganathancse:
recheck

Jaganathancse:
recheck

roles/edpm_network_config/tasks/os_net_config_tool.yml
Jaganathancse: @rabi done the changes.

rabi (approved):
We probably should delete /var/lib/tripleo-config/os-net-config.returncode as well.

Jaganathancse:
> We probably should delete /var/lib/tripleo-config/os-net-config.returncode as well.
ok, @karthiksundaravel can we keep or delete this file? need for os-net-config tripleo reference?

Jaganathancse:
> We probably should delete /var/lib/tripleo-config/os-net-config.returncode as well.

thanks @rabi done the changes. please check it.

Jaganathancse:
recheck

Jaganathancse:
recheck

Jaganathancse:
recheck

rabi (approved):

Jaganathancse:
recheck

karelyatin:
<< This change is to avoid reapplying os-net-config issue in adoption scenario and also ensure only reapplied when using 'edpm_network_config_update' varibale is 'true'.
But we always wanted to apply os-net-config during adoption to honor DNS/Domain updates as now exist in OCP side. **edpm_network_config_update** was meant only for follow up reapplies.
Can you add more context/background/Jira for this change?

bogdando:
I understand that ideally there must be no change in networking setup when upgrading osp into rhoso. But that is something to discuss with deployment (core openstack-operators) folks - how to apply **only DNS** config and touch nothing else. Today, if you skip configure-network service or noop it during adoption like this patch suggests doing by transitioning the last-apply state from osp to rhoso - the DNS stack won't function and the cloud will become broken. That is the status quo

Jaganathancse:
> I understand that ideally there must be no change in networking setup when upgrading osp into rhoso. But that is something to discuss with deployment (core openstack-operators) folks - how to apply **only DNS** config and touch nothing else. Today, if you skip configure-network service or noop it during adoption like this patch suggests doing by transitioning the last-apply state from osp to rhoso - the DNS stack won't function and the cloud will become broken. That is the status quo

@bogdando need to use edpm_network_config_update true to update dns and network config if needed.. so cleanup and reconfiguration based on template will be applied properly. 
but now reapplied but done like fresh deployment, cleanup also was not proper and causing network failure.

Jaganathancse:
> > I understand that ideally there must be no change in networking setup when upgrading osp into rhoso. But that is something to discuss with deployment (core openstack-operators) folks - how to apply **only DNS** config and touch nothing else. Today, if you skip configure-network service or noop it during adoption like this patch suggests doing by transitioning the last-apply state from osp to rhoso - the DNS stack won't function and the cloud will become broken. That is the status quo
> 
> @bogdando need to use edpm_network_config_update true to update dns and network config if needed.. so cleanup and reconfiguration based on template will be applied properly. but now reapplied but done like fresh deployment, cleanup also was not proper and causing network failure.

still we can use 'edpm_network_config_update' true if need to update network and dns configuration. any blockers here?

PR #1272, opened 2026-09-09
Report AEE run termination in pod logs for non-successful runs

rabi:
The termination log carries only host failure counts, not the run verdict - on timeout it reads all zeros while status and rc=254 are only recoverable from pod exitCode plus runner internals. Print the AEE summary to pod stdout for non-successful runs; successful runs stay silent and the termination-log payload is unchanged.

Related jira: [OSPRH-35300](https://redhat.atlassian.net/browse/OSPRH-35300)


rabi:
recheck

slagle (approved):

PR #1274, opened 2026-09-15
(WIP) Feat: EDPM upgrade interface PoC for 18->19

oliashish:
This PR is a structural proposal of the upgrade plan for 18->19.
- The single entrypoint -> playbooks/upgrade.yml
- A marker to validate the all the upgrade tasks are successfully ran and only run once in lifetime
- The single entrypoint calls edpm_upgrade_services which eventually calls service level upgrade.yml and then run.yml in order
- A roles/<service>/tasks/upgrade.yml file for each service which handles all the upgrade tasks.

Jira: [OSPRH-27148](https://redhat.atlassian.net/browse/OSPRH-27148)


oliashish:
For accessing design doc for this change, please check the Jira comment

roles/edpm_neutron_ovn/tasks/update.yml
slagle, line 1: this would have all the tasks removed right? Did you want to show that as part of this PR?

roles/edpm_upgrade_services/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+    name: osp.edpm.edpm_neutron_ovn
+    tasks_from: run.yml
+    apply:
+      become: true
+  tags:
+    - edpm_neutron_ovn
+    - edpm_upgrade_services
+  when:
+    - '"neutron-ovn" in edpm_upgrade_services_running_services'
+    - ansible_local.edpm_upgrade.upgrade.to_version | default('') != '19'
```
slagle, line 49: Looking at this in practice, I wonder if it would make more sense to just include run.yml inside of upgrade.yml for each service? Each service should own the entirety of its upgrade logic, so if both of these are required, I'm thinking it might be a better interface to have it all in upgrade.yml. WDYT?

roles/edpm_neutron_ovn/tasks/update.yml
oliashish: Yes, I was waiting for the approval on the design.
I'll remove it

roles/edpm_upgrade_services/tasks/main.yml
oliashish: Yes, we can do that.
I thought about it too but for this one I followed the current style on what we have for update_services i.e calling update.yml and then run.yml together for each service from edpm_update_services/tasks/container.yml. 
But we can do the on you mentioned.

